PR #12, opened 2024-04-26
Make run samples

daw3rd:
This vastly simplifies the running of any of the samples by adding makefle targets include a new capability to run the transform.py with cli arguments instead of as a py main(). 

New `make help` includes (for example from code_quality):

```
...
minio-load           Load ../../../transforms/code/cod_quality/test-data/input into minio at local/test/code_quality/input 
minio-start          Start the minio server and alias to local
...
run-cli-sample       Run code_quality_transform.py (if it exists).
run-local-ray-sample Run code_quality_local_ray.py (if it exists).
run-local-sample     Run code_quality_local.py (if it exists).
run-s3-ray-sample    Run code_quality_s3_ray.py (if it exists).
...
```

Update readmes to reflect changes.

Minimal changes to makefiles. suspect this may be worth the minimal risk to merge and fix a lot of the outstanding issues on running the samples.

I've run all the above or edited them out when not supported (some don't support local non-ray and malware does not have s3).


daw3rd:
@blublinsky  Some more work is need on the minio targets

.make.defaults
```diff
@@ -219,3 +219,32 @@ __check_defined = \
...
+.defaults.minio.check::
+	@# Help: Check that minio and the mc CLI is installed. 
+	$(MAKE) CHECK_RUNNABLE=minio .defaults.check.installed
+	$(MAKE) CHECK_RUNNABLE=mc .defaults.check.installed
+
+.PHONY: .defaults.minio.start
+.defaults.minio.start::
+	$(MAKE) .defaults.minio.check
+	@# Help: Start the minio server and alias to local
+	minio server start &
```
blublinsky: I think you also need to make sure that port 9000 is available. Something similar to https://github.com/IBM/data-prep-lab/blob/dev/kind/hack/tools/kind_management.sh#L37

.make.defaults
```diff
@@ -219,3 +219,32 @@ __check_defined = \
...
+	mc admin info local
+	mc alias set local http://127.0.0.1:9000 minioadmin minioadmin
+
+.PHONY: .defaults.minio.load-test-data
```
blublinsky, line 289: You are missing creation of minio secrets

.make.defaults
```diff
@@ -219,3 +219,32 @@ __check_defined = \
...
+.defaults.minio.load-test-data::
+	$(MAKE) .defaults.minio.check
+	@# Help: Load $(MINIO_SRC) into minio at local/test/$(MINIO_DEST) 
+	mc mb  local/test
```
blublinsky: Do you need to define MINIO_SRC/MINIO_DEST in the individual make files

.make.defaults
daw3rd: I just always stop the server prior to starting it now.

.make.defaults
daw3rd: Fixed

.make.defaults
daw3rd: Yep, added to targets calling this one.

transforms/.make.transforms
```diff
@@ -204,3 +201,67 @@ conventions:
...
+	@# Help: Run the $(TRANSFORM_NAME)_local_ray.py file (if it exists).
+	$(MAKE) RUN_FILE=$(TRANSFORM_NAME)_local_ray.py .transforms.run-src-file
+
+.PHONY: .transforms.run-s3-ray-sample
+.transforms.run-s3-ray-sample:
+	@echo ""
+	@echo "Please be sure you've started (make minio-start) and loaded (make minio-load) minio"
+	@echo ""
+	$(MAKE) RUN_FILE=$(TRANSFORM_NAME)_s3_ray.py .transforms.run-src-file
+	@# Help: Run $(TRANSFORM_NAME)_s3_ray.py file (if it exists).  Assumes minio has beens started.
```
blublinsky, line 206: Although there is help here to start minio and load data, it starts executing regardless and execution can hang for a while retrying S3 calls. I would suggest to remove help and start minio and load data as part of the command

transforms/code/code_quality/README.md
```diff
@@ -76,7 +46,24 @@ the following command line arguments are available in addition to
...
 * "--hf_token" - input the Hugging Face auth token to download the tokenizer. This option is only required for the tokenizer's whose access is restricted in Hugging Face.
 
-## Executing S3 examples
+### Running the samples
+To run the samples, use the following `make` targets
+
+* `run-cli-sample` - runs src/code_quality_transform.py using command line args
+* `run-local-ray-sample` - runs src/code_quality_local_ray.py
+* `run-s3-ray-sample` - runs src/code_quality_s3_ray.py
+    * Requires prior invocation of `make minio-start minio-load` to load data into local minio for S3 access.
```
blublinsky, line 55: No one is going to read this. I suggest removing it and start minio as part of the command

transforms/code/proglang_select/README.md
```diff
@@ -63,8 +50,26 @@ the following command line arguments are available in addition to
...
-To execute S3 examples, please refer to this [document](../../../data-processing-lib/doc/using_s3_transformers.md)
-for setting up MinIO and mc prior to running the example
+### Running the samples
+To run the samples, use the following `make` targets
 
+* `run-cli-sample` - runs src/proglang_select_transform.py using command line args
+* `run-local-sample` - runs src/proglang_select_local.py
+* `run-local-ray-sample` - runs src/proglang_select_local_ray.py
+* `run-s3-ray-sample` - runs src/proglang_select_s3_ray.py
+    * Requires prior invocation of `make minio-start minio-load` to load data into local minio for S3 access.
```
blublinsky, line 61: Same as above in every readme

.make.defaults
```diff
@@ -219,3 +219,56 @@ __check_defined = \
...
+	rm -rf $(MINIO_DIR) 
+	mkdir  $(MINIO_DIR) 
+	echo Starting minio server			
+	minio server $(MINIO_DIR) &
+	sleep 1
+	mc alias set $(MINIO_ALIAS) http://127.0.0.1:9000 minioadmin minioadmin
+	mc admin user svcacct add --access-key "$(MINIO_ADMIN_USER)" --secret-key "$(MINIO_ADMIN_PWD)" $(MINIO_ALIAS) minioadmin
+	mc alias ls $(MINIO_ALIAS)
+	@echo "Test minio server started" 
+
```
blublinsky, line 288: We have Minio strat and Minio load, where load does start. I would suggest to remove load as an explicit target and add load functionality to start. Less typing

blublinsky (commented):
There are some suggestions on usability, but in general its great

blublinsky:
An afterthought. I think we need to add stop MinIo to clean command

transforms/.make.transforms
daw3rd: I originally had this, but then the minio stuff is almost invisible to the user. and, there is a lot of logging so its hard to see what's really going on.  Thus the change to make the user explicitly manage minio.  Usually i'm a proponent of hiding complexity,  but in this case, I thought it would best to make the user at least aware of minio.

transforms/code/code_quality/README.md
daw3rd: Agreed, but this and `make help` for this target suggest minio must be started.  So I don't think they can actually run this target w/o being aware of the fact that minio has to be started (assuming they read the one liner).

transforms/.make.transforms
blublinsky: You can add an echo. I am convinced it should be one. Occasional users should not know and care about MiniO. It should just work out of the box.

transforms/code/code_quality/README.md
blublinsky: Remember. They can't read for the life of them and do not want to

transforms/.make.transforms
daw3rd: If you run it you will see there is one. :)

```
$ make run-s3-ray-sample

Please be sure you've started (make minio-start) and loaded (make minio-load) minio

/Library/Developer/CommandLineTools/usr/bin/make RUN_FILE=noop_s3_ray.py .transforms.run-src-file
source venv/bin/activate;	\
	cd src;				\
	python noop_s3_ray.py 
```

blublinsky: And this is causing an issue. People do not read. The start S3 sample and if Minio is not started and there is no data, they complain that it does not work

daw3rd: can't do much better than that.  other than doing it all for them.  i'm willing to require this minimum amount of reading in favor of spitting way too much logging at them for all the start, load, run, stop steps.  Also, makes them aware of the steps.

blublinsky: That is what I am suggesting. Its easy and will remove all the confusions

.make.defaults
daw3rd: Done.

transforms/.make.transforms
daw3rd: Added check to be sure minio is running.

blublinsky (approved):
LGTM

PR #18, opened 2024-04-29
fix docs

shivdeep-singh-ibm:
Some markdown renderers need a new line at the start to properly render bullet points. This commit adds new line before bullet points in some markdown files which would be rendered by mkdocs.


daw3rd (approved):
lgtm

PR #19, opened 2024-04-29
Updated code quality test data to include relevant columns

Param-S:
Updated test data to include only the necessary columns to reduce the size of the test files.


daw3rd (approved):
Doesn't seem to have a made a big file size change, but let's do it!

PR #28, opened 2024-04-30
Removing issues and PR counters till we figure a better solution for it

Bytes-Explorer:
David, It is nice to have a counter on open issues and PRs. However, the current one does not seem to show the right count. Removing this temporarily till we find a better way to do this or fix it.


blublinsky (approved):
LGTM

daw3rd (approved):

PR #30, opened 2024-04-30
Updated license header for code_quality filter

Param-S:
as some parts of this implementation are based on
BigCode and CodeParrot preprocessing method and
customized as per data-prep-lab framework


transforms/code/code_quality/src/code_quality_transform.py
```diff
@@ -25,12 +33,6 @@
 CODE_QUALITY_PARAMS = "code_quality_params"
 os.environ["TOKENIZERS_PARALLELISM"] = "false"
```
daw3rd, line 34: should this be a cli/configurable

daw3rd (approved):

PR #48, opened 2024-05-01
remove GUF porting references

daw3rd:


shahrokhDaijavad (approved):
Change strted to started.

PR #53, opened 2024-05-01
add EXTRA_INDEX_URL support to makefiles and fix issue with DOCKER_IM…

daw3rd:
…AGE value


blublinsky:
I am sorry, I am missing something. Why do we need this exactly? its OSS, so we are not planning to use internal repos (I hope)

blublinsky (approved):
LGTM

PR #59, opened 2024-05-02
fix s3 sample data input path (again)

daw3rd:

### DIFF
--- a/transforms/.make.transforms
+++ b/transforms/.make.transforms
@@ -242,7 +242,7 @@ conventions:
 .transforms.minio-load::
 	@# Help: Install the test-data/input files into minio for $(TRANSFORM_NAME)
 	dir=$$(find $(REPOROOT)/transforms -type d -name $(TRANSFORM_NAME));	\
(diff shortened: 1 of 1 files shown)


transforms/.make.transforms
```diff
@@ -242,7 +242,7 @@ conventions:
 .transforms.minio-load::
 	@# Help: Install the test-data/input files into minio for $(TRANSFORM_NAME)
 	dir=$$(find $(REPOROOT)/transforms -type d -name $(TRANSFORM_NAME));	\
-	$(MAKE) MINIO_SRC=$$dir/test-data/input MINIO_DEST=$(TRANSFORM_NAME)/input/ .defaults.minio.load-test-data
+	$(MAKE) MINIO_SRC=$$dir/test-data/input MINIO_DEST=$(TRANSFORM_NAME)/ .defaults.minio.load-test-data
```
blublinsky, line 245: My suggestion is to do 
```
$(MAKE) MINIO_SRC=$$dir/test-data/input/ MINIO_DEST=$(TRANSFORM_NAME)/input .defaults.minio.load-test-data
```
They both work. The difference is that in your implementation moves the folder, while in what I am suggesting its moving the content of the folder

blublinsky (approved):
Your choice

PR #61, opened 2024-05-03
Updated Makefile comment and make minio-start target extendible.

daw3rd:
Also added support for extending the minio-start make target so transforms can add more than the input data to minio if needed.

None


blublinsky (approved):
LGTM

PR #69, opened 2024-05-05
fix template's Python and OS options

roytman:
The current bug template provides the following Python options:

image

The the OS list is not correct also.


shahrokhDaijavad (approved):
LGTM

Issue #73, opened 2024-05-06
[Bug] ingest2parquet issuing ERROR messages, but WARNING is preferred.


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Tools/ingest2parquet

### What happened + What you expected to happen

Testing of ingest2parquet show lots of ERROR messages, but does not fail the test.

### Reproduction script

Lots of error messages without faling the run of ingest2parquet_local.py?  Perhaps these can be changed to WARNINGS?
```
cd tools/ingest2parquet
make venv
make test-src
```

gets

```
...
Executing: python src/ingest2parquet_local.py
13:02:37 INFO - data factory data_ is using local data accessinput_folder - /home/dawood/git/fm-data-engineering/tools/ingest2parquet/test-data/input output_folder - /home/dawood/git/fm-data-engineering/tools/ingest2parquet/src/../test-data/output
13:02:37 INFO - data factory data_ max_files -1, n_sample -1
13:02:37 INFO - data factory data_ Not using data sets, checkpointing False, max files -1, random samples -1, files to use ['.zip']
Number of files is 2 
filepath /home/dawood/git/fm-data-engineering/tools/ingest2parquet/src/utils/lang_extensions.json
13:02:37 ERROR - Error -> 'utf-8' codec can't decode byte 0x80 in position 11: invalid start byte
 skipping environments-master/cfortunes/diebenkorn_notes.dat Error: No contents decoded
13:02:37 ERROR - Error -> 'utf-8' codec can't decode byte 0xc3 in position 7: invalid continuation byte
 skipping environments-master/cfortunes/obliquestrategies.dat Error: No contents decoded
13:02:37 ERROR - Error -> 'utf-8' codec can't decode byte 0xfc in position 10: invalid start byte
 skipping application-java/lib/application-java.jar Error: No contents decoded
13:02:37 ERROR - Error -> 'utf-8' codec can't decode byte 0xe5 in position 14: invalid continuation byte
 skipping application-java/lib/fabric-gateway-java-2.1.1.jar Error: No contents decoded
13:02:37 ERROR - Error -> 'utf-8' codec can't decode byte 0xf9 in position 10: invalid start byte
 skipping application-java/lib/fabric-sdk-java-2.1.1.jar Error: No contents decoded
[...]

Issue #47, opened 2024-05-01
ingest2parquet Readme file needs to be cleaned up


shahrokhDaijavad:
1) There are still references to IBM bluepile in the Readme (command line options)
AST string containing input/output paths. input_folder: Path to input folder of files to
                        be processed output_folder: Path to output folder of processed files Example: {
                        'input_folder': '/cos-optimal-llm-pile/bluepile-
                        processing/rel0_8/cc15_30_preproc_ededup', 'output_folder': '/cos-optimal-llm-
                        pile/bluepile-processing/rel0_8/cc15_30_preproc_ededup/processed' }

2) In the section: Run the script via command-line, shouldn't it be:
 python ingest2parquet_local.py , instead of python ingest2parque.py ?

shahrokhDaijavad:
This is more than a Readme file issue, because for S3 runs of this tool, we do it differently from other transforms, in which we use "make minio-start" (which starts the minio server, creates the input directory and puts a sample input file there), so I re-classify this a bug, because of the inconsistency of S3 runs of this tool with S3 runs of all other transforms.

Issue #46, opened 2024-05-01
ingest2parquet Makefile needs new run-*-sample targets similar to transforms


daw3rd:
At least 

```
run-local-sample
run-s3-sample
```

Like the transforms, the `run-s3-sample` should start and load the minio server and then ask the user to stop the minio server.
Starting/loading/stopping minio should be done with the following make targets, like the transforms

```
minio-start - start and load test-data/input
minio-stop - stops the server
```
.make.defaults has support for these.

daw3rd:
verified

PR #74, opened 2024-05-06
ingest2parquet updates

sapthasurendran:
- Removed Error messages from log for files skipped
- Added make targets to run local and s3 samples using minio 
- Associated readme changes

#46 
#47 
#73


tools/ingest2parquet/.make.tools
```diff
@@ -0,0 +1,23 @@
+include $(REPOROOT)/.make.defaults
+
+# Expects RUN_FILE RUN_ARGS
+.PHONY: .tools.run-src-file
```
daw3rd: This is a cut/paste from .make.transforms.  Can you

1.  Move this to .make.defaults
2. Rename to `.defaults.run-src-file`
3. And replace the definintion of  `.transforms.run-src-file:` in .make.transforms with `.transforms.run-src-file: .defaults.run-src-file`

daw3rd (changes requested):

transforms/.make.transforms
```diff
@@ -224,7 +224,7 @@ conventions:
 
 # Expects RUN_FILE
 .PHONY: .transforms.run-src-file
-.transforms.run-src-file:
+.transforms.run-src-file: .defaults.run-src-file
```
daw3rd, line 227: should the body of this target be removed now that it calls .defaults?

daw3rd: otherwise, i think this will run twice.

daw3rd (changes requested):

transforms/.make.transforms
sapthasurendran: yes , removed the body and tested. works fine.

daw3rd (approved):

transforms/.make.transforms
daw3rd: please mark associated issues fixed.

daw3rd:
Please mark associated issues as fixed.

PR #87, opened 2024-05-08
add mkdocs hook to update relative links

shivdeep-singh-ibm:

mkdocs doesn't render relative links to non markdown files of github repo. This patch adds a hook which updates the reltive links to github links while building mkdocs documentation.

[43](https://github.com/IBM/data-prep-lab/issues/43
)


mkdocs_hook.py
```diff
@@ -0,0 +1,194 @@
+import logging
```
daw3rd: Can we move this file into .github dir so it does not clutter the top of the repo.  Sample for mkdocs.yml if possible.

daw3rd (changes requested):

mkdocs_hook.py
shivdeep-singh-ibm: Let me test if it works that way.

daw3rd (approved):

PR #100, opened 2024-05-10
Core library refactorings and transform adjustments

daw3rd:
Clears up some class and package organization.
Makes TransformConfiguration hold most config for the transform itself.  Moved from launcher
Extend ray-based launcher test to support python launcher.


blublinsky:
class https://github.com/daw3rd/data-prep-kit/blob/dev/data-processing-lib/src/data_processing/launch/ray/transform_configuration.py is not used anywhere and should be removed. The actual implementation that is used is https://github.com/daw3rd/data-prep-kit/blob/dev/data-processing-lib/src/data_processing/launch/ray/transform_runtime.py. Maybe split that file into 2?

data-processing-lib/src/data_processing/launch/transform_launcher.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        self.transform_config = transform_config
+        self.name = self.transform_config.name
+        self.data_access_factory = data_access_factory
+
+    def launch(self):
+        raise ValueError("must be implemented by subclass");
+
+    def get_transform_name(self) -> str:
+        return self.name
```
blublinsky, line 25: Frankly, I do not see much point in this, except the common origin

blublinsky (approved):
LGTM

PR #102, opened 2024-05-10
Fix python definition

roytman:
Otherwise, we cannot re-define python command

Signed-off-by: Alexey Roytman [roytman@il.ibm.com](mailto:roytman@il.ibm.com)


blublinsky (approved):
LGTM

PR #105, opened 2024-05-10
Add RayTransformConfiguration to a capture runtime_class

daw3rd:
Centralizes all transform configuration into a single configuration class for the ray launcher.


blublinsky (approved):
LGTM

PR #106, opened 2024-05-10
rename launch package to runtime

daw3rd:
runtime is a more representative name as there are other things besides the launcher.


blublinsky (approved):
LGTM

PR #112, opened 2024-05-11
removed unnecessary classes

blublinsky:

Simplified overall object model


daw3rd (changes requested):
The TransformConfigurationProxy needs to be restored

daw3rd (approved):

PR #120, opened 2024-05-13
adding additional KubeRay support

blublinsky:
TO add support for image pull policy


blublinsky:
I was trying to add support for image pull policy, @revit13, please take a look. NOOP transform is modified to do Always. It also adds autoscaling, but we are not planning to do it any time soon (as far as I know). While testing it, I somehow reproduced M1 issue (job execution submission twice) and hopefully fixed it. @roytman if you can try tests on your M1 Mac before doing KFP, this will be helpful

roytman:
The PX does not resolve the issues on a real cluster.

roytman (approved):

PR #131, opened 2024-05-15
transform multi_launcher

blublinsky:

Sometimes we want to execute a set of transforms for different input/output configuration


yuanchi2807:
Very, very elegant solution and I applaud you for that @blublinsky . 
But this is in data-processing-lib, so until the next major release, we will not touch that. Worthy though.

blublinsky:
> Very, very elegant solution and I applaud you for that @blublinsky . But this is in data-processing-lib, so until the next major release, we will not touch that. Worthy though.

@yuanchi2807 we can have it in both places

daw3rd (approved):

PR #132, opened 2024-05-15
Library refactoring/renaming for clarity

daw3rd:
rename base_configuration param/field to transform_config in runtimeclasses


blublinsky (approved):
LGTM

PR #137, opened 2024-05-16
Build kfp component image using libs from sources.

revit13:
This PR build kfp component image using libs from sources instead of installing them from pypi repo.


roytman (approved):
LGTM

PR #140, opened 2024-05-16
Restructure the repository to distinguish/separate runtime libraries

daw3rd:
We need to prepare for the separation of runtimes into separate libraries.  This is a first step to push the current data-processing-lib directory in data-processing-lib/ray directory.


blublinsky:
It looks good in general and I understand the motivation. The only question is why directory is called Ray?  Its really Ray+PurePython, which is not reflected in the directory name, that might confuse the user

blublinsky (approved):

PR #141, opened 2024-05-17
Binary

blublinsky:

Adds support for binary files processing


blublinsky:
Noop transform has an example of binary files usage

daw3rd (approved):

PR #151, opened 2024-05-20
Moving kfp workflows transform_workflows to transform directory.

revit13:

Kfp workflows under `kfp/transform_workflows` directory and have been tailored for individual transforms thus are better suited to be placed within the `transform` directory.


roytman (commented):
You should update https://github.com/IBM/data-prep-kit/blob/dev/doc/repo.md

transforms/code/code_quality/ray/kfp/Makefile
```diff
@@ -0,0 +1,24 @@
+# Define the root of the local git clone for the common rules to be able
+# know where they are running from.
+REPOROOT=../../../../..
+include $(REPOROOT)/transforms/.make.transforms
+
+workflow-venv::
```
roytman: can we skip one or two Makefiles levels and jump directly to the relevant directory?

roytman:
if we have separation to kfpv1 and kfpv2 we need to update `kfp/requirements.env`

roytman:
and [data-prep-kit](https://github.com/IBM/data-prep-kit/tree/dev)/[data-processing-lib](https://github.com/IBM/data-prep-kit/tree/dev/data-processing-lib)/[doc](https://github.com/IBM/data-prep-kit/tree/dev/data-processing-lib/doc)

/simplest-transform-tutorial.md

transforms/code/code_quality/ray/kfp/Makefile
revit13: Yes we can. I'll do that Thanks

kfp/superworkflows/v1/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+.PHONY: workflow-reconcile-requirements
+workflow-reconcile-requirements:
+	@for file in $(PYTHON_WF); do \
+		$(MAKE) .transforms_workflows.reconcile-requirements PIPELINE_FILE=$$file; \
+	done
+
+.PHONY: clean
+clean:
+	find . -name "*.bak" -type f -delete
```
blublinsky: Why not just 	rm -rf *.back || true

revit13: Fixed. Thanks

daw3rd (commented):
While it may be that the current kfp is ray-specific, I wonder if the kfp directory with subdir ray doesn't make more sense?  Today, the ray directory is structured as a standard python project.  Adding kfp there sort of breaks that model (maybe).  What do you think about moving the noop/ray/kfp content to noop/kfp/ray?  If/when we wrap spark with kfp, we would also have noop/kfp/spark.

roytman:
> While it may be that the current kfp is ray-specific, I wonder if the kfp directory with subdir ray doesn't make more sense? Today, the ray directory is structured as a standard python project. Adding kfp there sort of breaks that model (maybe). What do you think about moving the noop/ray/kfp content to noop/kfp/ray? If/when we wrap spark with kfp, we would also have noop/kfp/spark.

Make sense, thank you.

revit13:
> if we have separation to kfpv1 and kfpv2 we need to update `kfp/requirements.env`

@roytman if thats OK I'll  address the versions update in another PR, dedicated to V2, as it also requires the change in the pipeline version deployed under kind directory. I can add that to https://github.com/IBM/data-prep-kit/tree/kfp-v2 Thanks

roytman:
> While it may be that the current kfp is ray-specific, I wonder if the kfp directory with subdir ray doesn't make more sense? Today, the ray directory is structured as a standard python project. Adding kfp there sort of breaks that model (maybe). What do you think about moving the noop/ray/kfp content to noop/kfp/ray? If/when we wrap spark with kfp, we would also have noop/kfp/spark.

Even the current model is not clear/correct. All source files are under the Ray directory, but some of the launchers are Ray-independent. 
I understood that @cmadam is against wrapping spark by KFP. 
Thoughts?

cmadam:
@roytman : I am not against wrapping spark with kfp, but, given the fact that we are aiming to have the spark release ready at the end of this week, we cannot integrate this feature in the immediate future; we will need some time to develop it. We also need to have a design discussion before starting this work.
I am currently working with @daw3rd to add spark entries in the data-processing-lib and the transforms in the `spark-runtime` branch, and we could use the same structure for kfp. For now, let us define placeholders for Spark, and we can develop this capability in the future.

roytman:
OK, I see

roytman:
@revit13 , please put the kfp pipelines in e.g. the `noop/kfp_ray` directory

revit13:
> @revit13 , please put the kfp pipelines in e.g. the `noop/kfp_ray` directory

@roytman the kfp pipelines are now under `kfp_ray`. Thanks

revit13:
FTI- To avoid long testing time by running both noop_wf and noop_multiple_wf workflows, I disabled noop_multiple_wf when running workflow-test in noop directory.

roytman (approved):
LGTM

PR #153, opened 2024-05-20
Naming, docs and fix for recent binary file processing changes

daw3rd:
Documentation, normalization and a fix to noop that did not use `byte_array` as a parameter.


blublinsky (commented):
This does not seem to add anything

PR #158, opened 2024-05-21
Handle empty input parameter.

Mohammad-nassar10:


blublinsky:
I am not sure what exactly this fixes. Empty params are processed correctly through the set of default vaues

kfp/kfp_support_lib/src/kfp_support/workflow_support/utils/kfp_utils.py
```diff
@@ -89,10 +89,15 @@ def runtime_name(ray_name: str = "", run_id: str = "") -> str:
...
         res = f"python {executor} "
         for key, value in d.items():
-            if isinstance(value, str):
-                res += f'--{key}="{value}" '
+            if isinstance(value, bool):
+                if value:
+                    res += f"--{key} "
+            elif isinstance(value, str):
+                if value != "":
+                    res += f'--{key}="{value}" '
```
daw3rd, line 96: what if the value has double-quotes in it?

Mohammad-nassar10:
The purpose of this is to "ignore" the empty input parameters and don't include them in the command line invocation of the transform so it will use the default values as defined by the library.

kfp/kfp_support_lib/src/kfp_support/workflow_support/utils/kfp_utils.py
Mohammad-nassar10: It shouldn't have double-quotes. For example, the dictionary keys and values should be quoted with single-quote.

blublinsky (approved):

.make.versions
```diff
@@ -23,5 +23,5 @@ CODE_QUALITY_VERSION=0.3.0
 DOC_QUALITY_VERSION=0.3.0
 INGEST_TO_PARQUET_VERSION=0.3.0
 
-KFP_DOCKER_VERSION=0.1.0
+KFP_DOCKER_VERSION=0.1.1-dev0
```
roytman: please update it according to the latest version 0.2.0

.make.versions
```diff
@@ -5,11 +5,11 @@
 
 # Data prep lab wheel version
 DPK_LIB_VERSION=0.1.0
-DPK_LIB_KFP_VERSION=0.1.0
+DPK_LIB_KFP_VERSION=0.1.1-dev0
```
roytman: Please update before merging

kfp/kfp_support_lib/src/kfp_support/workflow_support/utils/kfp_utils.py
roytman: What will be if the value is None?

roytman: I suggest to use 
```python
if len(value) != 0:
   res += f'--{key}="{value}" '
```

Mohammad-nassar10: It shouldn't be `None` value here, because if the user keep the input as blank field it will be an empty string. Also `len(None)` will throw an exception.

blublinsky: +1 Mohhammad

blublinsky (approved):

kfp/kfp_support_lib/src/kfp_support/workflow_support/utils/kfp_utils.py
blublinsky: After thinking about it some more I think it is even simpler. Considering that the code has proper defaults for everything, I thing we can do this:

```
        for key, value in d.items():
            if str(value) != "":
                if isinstance(value, str):
                    res += f'--{key}="{value}" '
                else:
                    res += f"--{key}={value} "
        return res
```
So we are just removing parameters with empty value

roytman (approved):
LGTM

Issue #85, opened 2024-05-08
[Feature] Add spark support for filter module


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

We want to support users who work with Spark, Ray and pure python. The goal is to enable the filter framework on Spark. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

daw3rd:
This is done.

PR #183, opened 2024-05-23
Spark runtime

cmadam:
Add Spark to the data prep kit

https://github.com/IBM/data-prep-kit/issues/85


data-processing-lib/ray/src/data_processing/transform/abstract_transform.py
```diff
@@ -0,0 +1,16 @@
...
+class AbstractTransform(Generic[DATA]):
+    def transform(self, data: DATA) -> tuple[list[DATA], dict[str, Any]]:
+        """
+        Converts input table into an output table.
+        If there is an error, an exception must be raised - exit()ing is not generally allowed when running in Ray.
+        :param table: input table
+        :return: a tuple of a list of 0 or more converted tables and a dictionary of statistics that will be
+        propagated to metadata
+        """
+        raise NotImplemented()
```
blublinsky, line 16: Why do we need this additional indirection?
Unlike Java Python does not distinguish methods based on the signature. So one more method of indirection creates more issues in the long run. What is the advantage?

data-processing-lib/ray/src/data_processing/transform/binary_transform.py
```diff
@@ -10,10 +10,15 @@
...
-from typing import Any
+from typing import Any, TypeVar
 
+from data_processing.transform.abstract_transform import AbstractTransform
 
-class AbstractBinaryTransform:
+
+DATA = TypeVar("DATA")
+
+
```
blublinsky, line 20: This class defines different method names. No need to define data

data-processing-lib/ray/src/data_processing/transform/table_transform.py
```diff
@@ -59,17 +59,19 @@ def transform_binary(self, byte_array: bytes, ext: str) -> tuple[list[tuple[byte
...
-        raise NotImplemented()
+    # def transform(self, table: pa.Table) -> tuple[list[pa.Table], dict[str, Any]]:
+    #     """
+    #     Converts input table into an output table.
+    #     If there is an error, an exception must be raised - exit()ing is not generally allowed when running in Ray.
+    #     :param table: input table
+    #     :return: a tuple of a list of 0 or more converted tables and a dictionary of statistics that will be
+    #     propagated to metadata
+    #     """
+    #     raise NotImplemented()
```
blublinsky, line 75: This makes it less readable

data-processing-lib/ray/test/data_processing_tests/launch/pure_python/multi_launcher_test.py
```diff
@@ -51,6 +51,10 @@ def _submit_for_execution(self) -> int:
         return 0
 
 
+import pytest
+
+
+@pytest.mark.skip(reason="failing for now aws credentials")
```
blublinsky: please fix. It has to be something on your box. Its passing everywhere else

data-processing-lib/ray/test/data_processing_tests/launch/ray/ray_multi_launcher_test.py
```diff
@@ -54,6 +54,10 @@ def _submit_for_execution(self) -> int:
         return 0
 
 
+import pytest
+
+
+@pytest.mark.skip(reason="failing for now aws credentials")
```
blublinsky: Same here

data-processing-lib/spark/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+RUN echo "spark:x:185:0::/opt/spark/work-dir:/bin/bash" >> /etc/passwd
+WORKDIR ${SPARK_HOME}/work-dir
+ENTRYPOINT [ "/opt/entrypoint.sh" ]
+
+# Specify the User that the actual main process will run as
+ARG spark_uid=185
+USER ${spark_uid}
+
+# add pyspark, py4j to PYTHONPATH
+ENV PYTHONPATH=${SPARK_HOME}/python/lib/pyspark.zip:${SPARK_HOME}/python/lib/py4j-0.10.9.7-src.zip:${PYTHONPATH}
```
blublinsky, line 89: How is this better compared to prebuild PySpark image - https://hub.docker.com/layers/bitnami/spark/3.5.1/images/sha256-88b317b10433b9d87bbc409460716ef006f67f29433e89d13f3e7a8ac170d701?context=explore? The build is here https://github.com/bitnami/containers/tree/main/bitnami/spark/3.5/debian-12. In general, I think the common rule should be to avoid creating our own images, unless we absolutely have to


Alternatively Spark provides https://github.com/apache/spark/blob/master/bin/docker-image-tool.sh, which can be used to build image with this dockerfile https://github.com/apache/spark/blob/master/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/bindings/python/Dockerfile

https://atirek-ak.medium.com/building-an-apache-spark-image-from-spark-binaries-fe1adeb9a964 has step by step instructions

data-processing-lib/spark/src/data_processing_spark/runtime/spark/__init__.py
```diff
@@ -0,0 +1,8 @@
+from data_processing_spark.runtime.spark.runtime_config import SparkTransformRuntimeConfiguration
```
blublinsky, line 1: Can we, please, remove empty init files above?

data-processing-lib/spark/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+    wget https://repo1.maven.org/maven2/org/apache/commons/commons-configuration2/2.10.1/commons-configuration2-2.10.1.jar && \
+    wget https://repo1.maven.org/maven2/com/google/re2j/re2j/1.7/re2j-1.7.jar
+
+RUN rm hadoop-shaded-guava-1.1.1.jar
+
+# use the latest python 3.11 slim bookworm image
+# ARG PYTHON_IMAGE=python:3.11.9-slim-bookworm
+# FROM $PYTHON_IMAGE
+FROM python:3.11.9-slim-bookworm
+
```
blublinsky, line 39: We need to support Python 3.10 as well. See here https://medium.com/programmers-journey/deadsimple-pyspark-docker-spark-cluster-on-your-laptop-9f12e915ecf4 about Python versions.

data-processing-lib/spark/README.md
```diff
@@ -0,0 +1,4 @@
+# Spark Transform Runtime 
+
+This is the extension of the core data-prep-kit library to support running transform
+in spark cluster.
```
blublinsky: Thats it? Probably need some more

data-processing-lib/spark/config/spark_profile_kube.yml
```diff
@@ -0,0 +1,14 @@
...
+spark.executor.cores: ${EXECUTOR_CORES}
+spark.sql.shuffle.partitions: ${NUM_TASKS}
+spark.task.cpus: ${TASK_CPUS}
+spark.sql.legacy.parquet.nanosAsLong: true
+spark.executor.decommission.forceKillTimeout: "10h"
+# spark.sql.files.ignoreCorruptFiles: true
+# configuration needed when running in kubernetes
+spark.kubernetes.authenticate.driver.serviceAccountName: ${SERVICE_ACCOUNT}
+spark.kubernetes.container.image: ${EXECUTOR_DOCKER_IMAGE}
+spark.kubernetes.namespace: ${EXECUTOR_NAMESPACE}
```
blublinsky, line 14: I am not convinced we need this. When running Spark on Kubernetes, see https://blog.devgenius.io/spark-operator-basics-2ad7ecd469e1 for operator and why approach

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_execution_config.py
```diff
@@ -0,0 +1,45 @@
...
+        self.local_config_filepath = None
+        self.kube_config_filepath = None
+
+    def add_input_params(self, parser: argparse.ArgumentParser) -> None:
+        super().add_input_params(parser)
+        parser.add_argument(
+            f"--{local_config_path_cli}",
+            type=str,
+            default=local_config_path_default,
+            help="Path to spark configuration for run",
```
blublinsky, line 30: I think this is bad. Spark config has to be defined outside an application itself. See https://github.com/kubeflow/spark-operator/blob/master/docs/user-guide.md for the example. And we need a "standard config for local run"

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_execution_config.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        parser.add_argument(
+            f"--{kube_config_path_cli}",
+            type=str,
+            default=kube_config_path_default,
+            help="Path to Kubernetes-based configuration.",
+        )
```
blublinsky, line 37: I think this is quite confusing for the user. I think the better option is to just define local vs remote run

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
```diff
@@ -0,0 +1,219 @@
...
+                spark_config = yaml.safe_load(os.path.expandvars(config_fp.read()))
+            app_name = spark_config.get("spark.app.name", "my-spark-app")
+            self.spark = SparkSession.builder.appName(app_name).config(map=spark_config).getOrCreate()
+        else:
+            # we are running in Kubernetes, use spark_profile.yaml and
+            # environment variables for configuration
+
+            server_port = os.environ["KUBERNETES_SERVICE_PORT"]
+            master_url = f"k8s://https://kubernetes.default:{server_port}"
+
```
blublinsky: See comment above. I do not think it is the right approach. We should bite the bullet and adopt external configuration https://github.com/kubeflow/spark-operator/blob/master/docs/user-guide.md. This will make code smaller and provide a better configurability moving forward

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_transform.py
```diff
@@ -0,0 +1,12 @@
...
+from data_processing.transform.abstract_transform import DATA, AbstractTransform
+from pyspark.sql import DataFrame
+
+
+class AbstractSparkTransform(AbstractTransform[DataFrame]):  # todo: Remove double quotes
+    def __init__(self, config: dict):
+        self.config = config
+
+    def transform(self, data: DataFrame) -> tuple[list[DataFrame], dict[str, Any]]:
+        raise NotImplemented()
```
blublinsky, line 12: This is fundamentally different from Python/Ray. Just bite a bullet and create a separate AbstractSparkTransform. Adding here AbstractTransform does not really add anything to the code. Its OO for the sake of OO nothing else. Also by using DataFrames here we are effectively limiting input to the Arrow tables. Just making sure that we understand this. There is no way back to the binary blobs

data-processing-lib/spark/src/data_processing_spark/test_support/launch/abstract_launcher_test.py
```diff
@@ -0,0 +1,8 @@
+from data_processing.test_support.launch.transform_test import (
```
blublinsky, line 5: Please remove empty init files

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
```diff
@@ -0,0 +1,219 @@
...
+)
+from data_processing_spark.runtime.spark.spark_transform import AbstractSparkTransform
+from pyspark.sql import DataFrame, SparkSession
+
+
+logger = get_logger(__name__)
+
+
+class SparkTransformLauncher(AbstractTransformLauncher):
+    """
```
blublinsky, line 33: Sorry this is not launcher. This is a combination of launcher and executor. Please split this into 2 classes:
1. Something that in Ray is called orchestrator
2. Actual launcher

blublinsky:
This is a good starting point, which I think needs quite a few changes. Another big question for me is that I do not see how this can be extended to support dedups or even bloclisting

data-processing-lib/ray/src/data_processing/transform/abstract_transform.py
daw3rd: It formalizes the notion of a "transform" which we can then use across runtimes.  I'd like to have some commonality across them and not look like we're reinventing.

data-processing-lib/ray/src/data_processing/transform/binary_transform.py
daw3rd: But we need to enable passing DATA to the super class as a generic type.  See line 21 below.

data-processing-lib/ray/src/data_processing/transform/table_transform.py
daw3rd: We're using the generic defintion of transform() from the super class so it is not needed here.

data-processing-lib/ray/test/data_processing_tests/launch/pure_python/multi_launcher_test.py
daw3rd: Thank you and fixed,  and yes this passes at the command line and probably would in pycharm if it would use my existing env vars :(.

data-processing-lib/ray/test/data_processing_tests/launch/ray/ray_multi_launcher_test.py
daw3rd: also fixed

data-processing-lib/spark/README.md
daw3rd: for sure, but I don't want to hold up the code changes for documentation

data-processing-lib/ray/src/data_processing/transform/abstract_transform.py
blublinsky: This is a very noble intention, but do we really need this? Can you name at least one realistic use case where this can be used? At the moment, it's an extra baggage that does not add anything

data-processing-lib/ray/src/data_processing/transform/binary_transform.py
blublinsky: That is the complexity I am talking about. Why?

data-processing-lib/ray/src/data_processing/transform/table_transform.py
blublinsky: Again, we sacrifice the readability of the code, and what is the gain? A simple cost/benefit analysis

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
daw3rd: orchestrator is an implementation detail and the "orchestration" is essentially done in side _run_transform() with  spark implementing the "orchestration" piece.  if useful, we can certainly add a separate orchestrator later.

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_transform.py
daw3rd: This comes in because of the AbstractTransformLauncher -> runtime config -> transform config -> AbstractTransform.  Spark code is using the AbstractTransformLauncher and therefore we have a common AbstractTransform across all runtimes/launchers.

data-processing-lib/ray/src/data_processing/transform/table_transform.py
daw3rd: Copying from below
`This comes in because of the AbstractTransformLauncher -> runtime config -> transform config -> AbstractTransform. Spark code is using the AbstractTransformLauncher and therefore we have a common AbstractTransform across all runtimes/launchers.`

data-processing-lib/ray/src/data_processing/transform/binary_transform.py
daw3rd: from comment below
`This comes in because of the AbstractTransformLauncher -> runtime config -> transform config -> AbstractTransform. Spark code is using the AbstractTransformLauncher and therefore we have a common AbstractTransform across all runtimes/launchers.`

data-processing-lib/ray/src/data_processing/transform/abstract_transform.py
daw3rd: from below
`This comes in because of the AbstractTransformLauncher -> runtime config -> transform config -> AbstractTransform. Spark code is using the AbstractTransformLauncher and therefore we have a common AbstractTransform across all runtimes/launchers.`

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
daw3rd: Your point here is a valid one and we should pursue it after we get this initial implementation in.  Is there any reason why we should not have this initial implementation, even if we decide to deprecate it in the future?  Would it somehow make all the class/launcher design obsolete? If yes, can you say why?

data-processing-lib/spark/src/data_processing_spark/test_support/launch/abstract_launcher_test.py
daw3rd: Done.

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
daw3rd: and actually, i think the notion of "orchestration" is a ray-specific implementation detail.  That is, our ray implementation has ray worker that is performing the orchestration - very ray specific IMHO.  I don't see why we need to  have an "orchestrator" for python for example?

data-processing-lib/spark/Dockerfile
cmadam: 1. We are already using `docker-image-tool.sh` tool to build the image.  Using this script directly is not a good practice, as it would require downloading a spark release on the local machine, and then running this script.  This approach cannot be automated, so we have encapsulated the download of the spark release and the invocation of this script in the docker image.

2. The bitnami spark image has several shortcomings - it is not built for running Spark in kubernetes, and it is not built for running Spark with S3.  As such, it requires several changes made to it, for example changing the `ENTRYPOINT`, without which it will not run in Kubernetes (on the driver node, or on the executor nodes). So we cannot use this image as is, and at this point I do not see any value in starting from this image, rather than from our base image.  Moreover, by looking at the size of this image (1.65GB), it is much larger than the size of our image (945MB); to use this image we will need to do additional investigation as to which packages are included in it, and where does the difference in size come from.

3. The article you cited contains an approach that has the drawback I listed under point 1 - it assumes that we are manually downloading a spark release on our machine and then we are manually running the `docker-image-tool.sh` script

data-processing-lib/ray/src/data_processing/transform/abstract_transform.py
blublinsky: And how is this helping us exactly?

data-processing-lib/ray/src/data_processing/transform/binary_transform.py
blublinsky: And the same question. How is this helping us?

data-processing-lib/ray/src/data_processing/transform/table_transform.py
blublinsky: And the same answer as above

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
blublinsky: There is nothing more permanent as temporary solution

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_transform.py
blublinsky: So again, we are making this more complex and less readable and for what?

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
blublinsky: It is not. The responsibility of the launcher is to launch, not define the execution. Execution definition (orchestration) is a separate concern from launching

data-processing-lib/ray/src/data_processing/transform/table_transform.py
daw3rd: We have a common/unified framework, not independent ones.

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_transform.py
daw3rd: complexity is in the ey of the beholder :)

data-processing-lib/spark/src/data_processing_spark/runtime/spark/spark_launcher.py
daw3rd: i respectfully disagree

daw3rd (approved):

PR #186, opened 2024-05-24
library versions update

blublinsky:
updated library/kfp image versions to 0.1.1 to capture the latest changes

### DIFF
--- a/.make.versions
+++ b/.make.versions
@@ -4,8 +4,8 @@
 ################################################################################
 
 # Data prep lab wheel version
-DPK_LIB_VERSION=0.1.0
-DPK_LIB_KFP_VERSION=0.1.0
+DPK_LIB_VERSION=0.1.1
+DPK_LIB_KFP_VERSION=0.1.1
 
 # Begin transform versions/tags
 BLOCKLIST_VERSION=0.3.0
@@ -23,5 +23,5 @@ CODE_QUALITY_VERSION=0.3.0
 DOC_QUALITY_VERSION=0.3.0
 INGEST_TO_PARQUET_VERSION=0.3.0
 
-KFP_DOCKER_VERSION=0.2.0
+KFP_DOCKER_VERSION=0.1.1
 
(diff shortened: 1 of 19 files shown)


.make.versions
```diff
@@ -23,5 +23,5 @@ CODE_QUALITY_VERSION=0.3.0
 DOC_QUALITY_VERSION=0.3.0
 INGEST_TO_PARQUET_VERSION=0.3.0
 
-KFP_DOCKER_VERSION=0.2.0
+KFP_DOCKER_VERSION=0.1.1
```
roytman, line 26: why did you returned the major version back?

kfp/kfp_support_lib/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "data_prep_toolkit_kfp"
-version = "0.1.0"
+version = "0.1.1"
```
roytman, line 3: the version is wrong

kfp/kfp_ray_components/createRayClusterComponent.yaml
```diff
@@ -11,7 +11,7 @@ inputs:
 
 implementation:
     container:
-        image: quay.io/dataprep1/data-prep-kit/kfp-data-processing:0.2.0
+        image: quay.io/dataprep1/data-prep-kit/kfp-data-processing:0.1.1
```
roytman, line 14: it should be 0.2.1

roytman (changes requested):

.make.versions
blublinsky: because we are really on 0.1.1. Trying to be consistent

roytman:
the pipelines should be updates as well

kfp/kfp_support_lib/pyproject.toml
blublinsky: no, its 0.1.1, which is correct

kfp/kfp_ray_components/createRayClusterComponent.yaml
blublinsky: See above

blublinsky:
> he pipelines should be updates as well

done

roytman (approved):

Issue #185, opened 2024-05-24
Python and ray modules need to be separated into different folders


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

Two requests: 
1) Python and ray modules need to be separated into different folders
2) Please consider if you want separate readme for folders. Readme should include instructions for usage for all runtimes. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Bytes-Explorer:
Needed for release 1

PR #213, opened 2024-05-29
Split data-processing-lib/ray into python and ray.

daw3rd:
Towards splitting the transforms into separate runtimes, we need to split the library first into python, ray and spark (the latter already exists).

This also fixes some DataAccessS3 bugs introduced with the spark-runtime merge.

https://github.com/IBM/data-prep-kit/issues/185


.make.versions
```diff
@@ -1,29 +1,31 @@
...
+NOOP_VERSION=0.9.0
+NOOP_SPARK_VERSION=0.2.0
+RESIZE_VERSION=0.4.0
+LANG_ID_VERSION=0.4.0
+TOKENIZATION_VERSION=0.4.0
+MALWARE_VERSION=0.5.0
+PROGLANG_SELECT_VERSION=0.4.0
+CODE_QUALITY_VERSION=0.4.0
+DOC_QUALITY_VERSION=0.4.0
+INGEST_TO_PARQUET_VERSION=0.4.0
```
blublinsky, line 28: I think it is about time to align image versions. Assuming that 0.2.0 is the major release, they should be 0.2.x

.make.versions
```diff
@@ -1,29 +1,31 @@
...
-# repository in Makefile format.
+# repository in Makefile format.  These are generally considered the version
+# numbers to be published on the next release/publishing of artifacts.
 ################################################################################
 
 # Data prep lab wheel version
-DPK_LIB_VERSION=0.1.1
-DPK_LIB_KFP_VERSION=0.1.1
+DPK_LIB_VERSION=0.2.0
+DPK_LIB_KFP_VERSION=0.2.0
```
blublinsky, line 9: Considering that ingest to parquet adds a few changes to the libraries, do you want to postpone PyPI creation? or do this as a minor update?

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
```diff
@@ -51,19 +51,17 @@ def __init__(
         self.s3_credentials = {} | s3_credentials
         access_key = self.get_access_key()
         if access_key is None:
+            logger.warning("S3 access key not provided. Defaulting to ''")
```
blublinsky: I do not see much point in this. If the access key is not provided, we should exit (throw an exception). Data Access factory do just this in validation

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
```diff
@@ -51,19 +51,17 @@ def __init__(
...
             access_key = ""
         secret_key = self.get_secret_key()
         if secret_key is None:
+            logger.warning("S3 secret key not provided. Defaulting to ''")
```
blublinsky: I do not see much point in this. If the secret key is not provided, we should exit (throw an exception). Data Access factory do just this in validation

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
```diff
@@ -73,17 +71,31 @@ def __init__(
...
-        return (self.s3_credentials.get("secret_key", None),)
+        return self.s3_credentials.get("secret_key", None)
 
     def get_endpoint(self):
-        return (self.s3_credentials.get("url", None),)
+        return self.s3_credentials.get("url", None)
 
     def get_region(self):
-        return (self.s3_credentials.get("region", None),)
+        return self.s3_credentials.get("region", None)
```
blublinsky, line 96: Sorry, these are completely artificials. It was fine before. I do not see the need for these

data-processing-lib/ray/pyproject.toml
```diff
@@ -1,23 +1,17 @@
...
 dependencies = [
+    "data-prep-toolkit",
     "ray[default]==2.9.3",
-    "pyarrow==15.0.2",
-    "boto3==1.34.69",
-    "argparse",
-    "mmh3",
-    # These two are to fix security issue
-    "fastapi>=0.109.1",
-    "pillow>=10.2.0",
```
blublinsky, line 20: Removed these 2 will create security issues in the image

data-processing-lib/ray/pyproject.toml
```diff
@@ -1,23 +1,17 @@
...
-description = "Data Preparation Toolkit Library"
+description = "Data Preparation Toolkit Library for Ray"
 license = {text = "Apache-2.0"}
 readme = {file = "README.md", content-type = "text/markdown"}
 authors = [
     { name = "David Wood", email = "dawood@us.ibm.com" },
     { name = "Boris Lublinsky", email = "blublinsky@ibm.com" },
 ]
 dependencies = [
+    "data-prep-toolkit",
```
blublinsky: I think we need a version here

.make.versions
daw3rd: that would be moving versoins backwards though. definitely not good

.make.versions
daw3rd: we can postpone publishing, probably until we're publishing transforms too

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
daw3rd: done

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
daw3rd: done

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
daw3rd: they are need for spark to set the credentials in hadoop.

data-processing-lib/ray/pyproject.toml
daw3rd: I removed them because I could not find them in the image or venv if i removed them here.  Maybe a specific transform is/was using them, but I can't see them any where now.

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
cmadam: If the S3 data access key is not provided, IMHO, the program should use the credentials given in the `~/.aws/credentials` file.  This is the default behavior of the `boto3` client.

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
cmadam: Is the data access factory currently checking the `~/.aws/credentials` file when running locally?

.make.versions
blublinsky: Currently they are all random numbers and we have no idea how to link them to the library version. So it is good

.make.versions
blublinsky: That can be a bit tooo long

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
blublinsky: Sorry, no. We tried it before and it led to a lot of confusion. Also, remember, our main goal is running on the cluster not locally

blublinsky (approved):

.make.versions
daw3rd: I think we agreed this will be taken up in a separate PR

data-processing-lib/ray/pyproject.toml
daw3rd: I moved these from the python toml back to the ray toml.

data-processing-lib/ray/pyproject.toml
daw3rd: done

PR #228, opened 2024-06-04
Transform conventions and documentation and typo in .make.defaults venv

daw3rd:
Add some more docs on transform project conventions.
Fix a bug in .make.defaults rule to build the venv when using pyproject.toml (it was ignored).


roytman (approved):
LGTM

PR #231, opened 2024-06-05
ingest to parquet rewrite

blublinsky:


data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
```diff
@@ -68,32 +62,12 @@ def __init__(
         self.m_files = m_files
         self.n_samples = n_samples
         self.files_to_use = files_to_use
-
-        logger.debug(f"S3 access key provided: {access_key}")
-        logger.debug(f"S3 secret key provided: no soup for you!")
```
daw3rd, line 73: why are you deleting all of this, it never hurts to have debugging logs.  and the get_*() methods are needed by spark.  would recommend only changing what is needed for this pr.

data-processing-lib/python/src/data_processing/test_support/launch/transform_test.py
```diff
@@ -21,7 +21,7 @@
 
 class AbstractTransformLauncherTest(AbstractTest):
     """
-    The launcher test class for all/most AbstractTransformLauncher implementations.
+    The Ray-based test class for all/most AbstractTransform implementations.
```
daw3rd: You seem to be dropping things from dev.  This comment was more correct and the new ray-specific one is not.

data-processing-lib/python/src/data_processing/test_support/launch/transform_test.py
```diff
@@ -30,58 +30,64 @@ class AbstractTransformLauncherTest(AbstractTest):
...
+            in_table_path: str,
+            expected_out_table_path: str,
+            ignore_columns: list[str],
     ):
         """
         Test the given transform and its runtime using the given CLI arguments, input directory of data files and expected output directory.
         Data is processed into a temporary output directory which is then compared with the directory of expected output.
         :param launcher: launcher configured to run the transform being tested
-        :param cli_params: a map of the simulated CLI arguments (w/o --).  This includes both the transform-specific CLI parameters and the launching args.
+        :param cli_params: a map of the simulated CLI arguments (w/o --).  This includes both the transform-specific CLI parameters and  the Ray launching args.
```
daw3rd: more comments lost

transforms/.make.transforms
```diff
@@ -293,7 +293,7 @@ conventions: .transforms.check_required_macros
 .transforms.minio-load::
 	@# Help: Install the test-data/input files into minio for $(TRANSFORM_NAME)
 	dir=$$(find $(REPOROOT)/transforms -type d -name $(TRANSFORM_NAME));	\
-	$(MAKE) MINIO_SRC=$$dir/test-data/input MINIO_DEST=$(TRANSFORM_NAME)/ .defaults.minio.load-test-data
+	$(MAKE) MINIO_SRC=$$dir/ray/test-data/input MINIO_DEST=$(TRANSFORM_NAME)/ .defaults.minio.load-test-data
```
daw3rd, line 296: `ray` should probably be changed to `$(TRANSFORM_RUNTIME)` to be runtime-independent.  The value of this macro is the name of the directory. one of python, ray or spark.

transforms/code/ingest_2_parquet/kfp_ray/v1/ingest_2_parquet_wf.py
```diff
@@ -0,0 +1,178 @@
...
+    ONE_WEEK_SEC,
+    ComponentUtils,
+)
+from data_processing.utils import GB
+
+
+# the name of the job script
+EXEC_SCRIPT_NAME: str = "ingest_2_parquet_transform.py"
+
+task_image = "quay.io/dataprep1/data-prep-kit/ingest_2_parquet:0.3.0"
```
daw3rd, line 27: think this version will eventually also need .dev6 suffix

transforms/code/ingest_2_parquet/ray/src/ingest_2_parquet_transform.py
```diff
@@ -0,0 +1,271 @@
...
+        lang_dict = _get_supported_languages(
+            lang_file=lang_file,
+            data_access=lang_data_access_factory.create_data_access(),
+        )
+        lang_refs = ray.put(lang_dict)
+        logger.debug(f"Placed language list into Ray object storage under reference{lang_refs}")
+        return {ingest_supported_languages: lang_refs} | self.params
+
+
+class IngestToParquetTransformConfiguration(TransformConfiguration):
```
daw3rd: Think we should be consistent on the naming of classes and files like the other transforms.  Pick `ingesttoparquet` or `ingest2parquet` or `ingest_2_parquet` or `ingest_to_parquet`, but lets be consistent.

transforms/code/ingest_2_parquet/ray/test-data/input/application-java.zip
daw3rd, line 1: too late, but do we really need a 27mb file?!

daw3rd (changes requested):

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
blublinsky: get_methods are artificial, the do exactly the same that get from dict does. All the logs are printed by the factory already. Why have the code that is unnecessary

data-processing-lib/python/src/data_processing/test_support/launch/transform_test.py
blublinsky: fixed

data-processing-lib/python/src/data_processing/test_support/launch/transform_test.py
blublinsky: fixed

transforms/.make.transforms
blublinsky: Just made it work. Moving forward, yes

transforms/code/ingest_2_parquet/kfp_ray/v1/ingest_2_parquet_wf.py
blublinsky: Again, for the first version its good enough

transforms/code/ingest_2_parquet/ray/src/ingest_2_parquet_transform.py
blublinsky: File names and class names do not have to be the same. File names are dependent on directory name in the make files

transforms/code/ingest_2_parquet/ray/test-data/input/application-java.zip
blublinsky: THats what she had
Just copied it

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
cmadam: IMO, the `get_*()` methods provide a nice encapsulation, hiding the internals from the module calling them, e.g. spark.  That being said, spark module would still work if it would use
```
data_access_s3.s3_credentials.get("access_key", None)
```
instead of:
```
data_access_s3.get_access_key()
```

blublinsky: @cmadam Why does Spark need them in the first place? Is it using methods provided by the class or using the class as a placeholder for info?

roytman (approved):
LGTM, but need an approvement from @daw3rd or @cmadam

Bytes-Explorer:
@shivdeep-singh-ibm Please verify that the functionality is exactly the same as the module that was written by Saptha.

shivdeep-singh-ibm (approved):

data-processing-lib/python/src/data_processing/data_access/data_access_s3.py
cmadam: @blublinsky:  The local and S3 data access classes do not have the capability of reading a list of files into a Spark DataFrame; this operation is done using Spark methods. So Spark is only partially using the data access interface, to get a list of files to process, or to get the S3 credentials, but reading and writing to/from Spark DataFrames is being done outside the data access, as data access does not have those capabilities.

blublinsky: @cmadam so which getters do you need? All of them?

blublinsky: @cmadam the getters are back.

cmadam (approved):

transforms/.make.transforms
daw3rd: unclear why you won't make this change like you did for the comments?  The time is now :)

transforms/code/ingest_2_parquet/kfp_ray/v1/ingest_2_parquet_wf.py
daw3rd: but now is the time,  do you think someone is going to fix this later?

transforms/code/ingest_2_parquet/ray/src/ingest_2_parquet_transform.py
daw3rd: consistency in conventions will help everyone in the end.

Also, currently this is for code, but i think the goal is to have this be more generic.  As such, it belongs in the universal tree not the code tree.  Yes, in anticipation of future extensions and not having to move it later.

PR #236, opened 2024-06-06
Fix tasks tags in kfp workflows.

revit13:

Currently the task tags in kfp workflow are not updated correctly. This PR  tries to fix that


.make.versions
```diff
@@ -13,29 +13,45 @@ RELEASE_VERSION_SUFFIX=.dev6
 # Data prep lab wheel version
 DPK_LIB_VERSION=0.2.0$(RELEASE_VERSION_SUFFIX)
 DPK_LIB_KFP_VERSION=0.2.0$(RELEASE_VERSION_SUFFIX)
-DPK_LIB_KFP_VERSION_v2=0.2.0$(RELEASE_VERSION_SUFFIX)
```
roytman: I think we need a separate image for v2

roytman (approved):
LGTM, but would like that @daw3rd or @blublinsky will take a look.

.make.versions
revit13: Fixed. Thanks

daw3rd:
LGTM

Issue #65, opened 2024-05-03
minio is not able to start


Hajar-Emami:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP workflows

### What happened + What you expected to happen

After installing the pre-required softwares manually, I got this by running make setup:
```
creating minio alias
mc: Configuration written to `/root/.mc/config.json`. Please update your access credentials.
mc: Successfully created `/root/.mc/share`.
mc: Initialized share uploads `/root/.mc/share/uploads.json` file.
mc: Initialized share downloads `/root/.mc/share/downloads.json` file.
Added `kfp` successfully.
creating test bucket
Bucket created successfully `kfp/test`.
copying data
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/code/code_quality/test-data/input/sample_1.parquet`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:50442->127.0.0.1:8080: i/o timeout
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/code/proglang_select/test-data/input/test1.parquet`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:52100->127.0.0.1:8080: i/o timeout
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/code/proglang_select/test-data/languages/allowed-code-languages.txt`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:53306->127.0.0.1:8080: i/o timeout
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/code/malware/test-data/input/sample.parquet`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:40052->127.0.0.1:8080: i/o timeout
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/universal/ededup/test-data/input/sample1.parquet`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:41522->127.0.0.1:8080: i/o timeout
mc: <ERROR> Failed to copy `/root/test/data-prep-lab/transforms/universal/noop/test-data/input/test1.parquet`. Get "http://localhost:8080/test/?location=": read tcp 127.0.0.1:39912->127.0.0.1:8080: i/o timeout
[...]

PR #240, opened 2024-06-07
Readme, pyproject metadata and makefile fixes in noop and filter.

daw3rd:
Update filter and noop readmes, Fix filter Makefiles regard running s…amples, add metadata to pyproject.toml for the libraries.

### DIFF
--- a/data-processing-lib/python/pyproject.toml
+++ b/data-processing-lib/python/pyproject.toml
@@ -2,6 +2,7 @@
 name = "data_prep_toolkit"
 version = "0.2.0.dev6"
 requires-python = ">=3.10"
+keywords = ["data", "data preprocessing", "data preparation", "llm", "generative", "ai", "fine-tuning", "llmapps" ]
 description = "Data Preparation Toolkit Library"
 license = {text = "Apache-2.0"}
 readme = {file = "README.md", content-type = "text/markdown"}
@@ -15,6 +16,11 @@ dependencies = [
     "argparse",
     "mmh3",
 ]
+
+[project_urls]
+Repository = "https://github.com/IBM/data-prep-kit"
+Issues = "https://github.com/IBM/data-prep-kit/issues"
+Documentation = "https://ibm.github.io/data-prep-kit/"
 
 [build-system]
 requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
--- a/transforms/universal/filter/ray/Makefile
+++ b/transforms/universal/filter/ray/Makefile
@@ -46,13 +46,9 @@ run-cli-sample:
 		--filter_columns_to_drop \"[ 'extra', 'cluster' ]\" "	\
                 .transforms.run-src-file
 
+run-local-sample: .transforms.run-local-ray-sample
 
-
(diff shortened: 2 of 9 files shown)


transforms/universal/filter/ray/Makefile
```diff
@@ -46,13 +46,9 @@ run-cli-sample:
...
 
-
-run-local-python-only-sample: .transforms.run-local-sample
-
-run-local-sample: .transforms.run-local-sample
-
-run-s3-sample: .transforms.run-s3-sample
+run-s3-sample: .transforms.run-s3-ray-sample
 
 minio-start:	.minio-start
```
blublinsky, line 54: Can we fix minio-start as well. Currently it does not copy additional files beyond input

daw3rd: can we make this a separate PR.  its pretty orthogonal to these changes

data-processing-lib/python/pyproject.toml
```diff
@@ -2,6 +2,7 @@
 name = "data_prep_toolkit"
 version = "0.2.0.dev6"
 requires-python = ">=3.10"
+keywords = ["data", "data preprocessing", "data preparation", "llm", "generative", "ai", "fine-tuning", "llmapps" ]
 description = "Data Preparation Toolkit Library"
 license = {text = "Apache-2.0"}
```
blublinsky, line 7: Should it specify Python in the description and maybe name??

transforms/universal/filter/ray/Makefile
blublinsky: We can, but, we talked about it several times and its still not there

daw3rd: Submit an issue and don't poke arbitrary fingers at random PRs over this.  Make an issue and fix it in a separate PR.  This PR is about something else!  We can't start holding up PRs over random unrelated (and unreported) issues.

data-processing-lib/python/pyproject.toml
daw3rd: In the future, please make a suggestion to your liking.  How about

`Data Preparation Toolkit for Python`

However, this seems a bit redundant since pypi sort of implies python.

blublinsky: We have 3 different libraries - one for Ray, one for Spark and one for ? nothing?

transforms/universal/filter/ray/Makefile
blublinsky: Well, `minio-start` is currently broken. In addition to what I noted here before its also always getting data from ray directory. So I am not convinced we should merge a PR relying on the broken things

cmadam: The issue https://github.com/IBM/data-prep-kit/issues/65 is covering this problem, but it has been marked as fixed.  Do we need to re-open that issue again?

blublinsky (approved):

data-processing-lib/python/pyproject.toml
cmadam: I think the name of the library, `Data Preparation Toolkit Library` is clear enough.

PR #243, opened 2024-06-08
Initial split of tokenization transform into ray and python

daw3rd:
Continuing work to split transfom runtimes.  This time it is universal/tokenization

Issue #185


.secrets.baseline
```diff
@@ -135,7 +135,17 @@
...
+      {
+        "hashed_secret": "ec36b3e3e61563a3002cbac1fb1407757b31b8ee",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 55,
+        "type": "Base64 High Entropy String",
+        "verified_result": null
+      }
+    ],
+    "transforms/universal/tokenization/ray/test-data/ds02/expected/metadata.json": [
```
blublinsky, line 148: Why is this here?

blublinsky:
Sorry, I do not like this whole thing. Its way to complex and cumbersome for users. It also creates tons of libraries

.secrets.baseline
daw3rd: this is not new, it needed to be added because  test-data/expected/metadata.json was copied from ray to python dirs.  For whatever reason, pre-commit seems to think the following is a secret

`"name": "/Users/xdang/00proj/04-FM/01_code/fm-data-engineering/transforms/universal/tokenization/output/ds02",`

blublinsky (approved):

PR #244, opened 2024-06-08
Deanw pr

deanwampler:
Use these changes as you see fit, or not...

The `Makefile` and `.make.defaults` change is optional; It adds a convenient `all` target.

The `*.toml` file changes add dependencies I found I needed, as documented in the "friction log" I shared June 8. 

N/A.


daw3rd:
This PR #247 addresses the `make test` failure which was due to the venv not being prebuilt.  
As to the `all` target, I'm not sure we're ready to declare what `all` really means.

daw3rd:
#247 has been merged.  can you close this one?  If you feel strongly about `all` maybe an issue for that?

PR #245, opened 2024-06-08
add retries counter to data processing

blublinsky:

Adds reading and adding data access retries to statistics

https://github.ibm.com/ai-models-data/data-prep-kit-inner/issues/16#issuecomment-81632985


data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -62,53 +62,56 @@ def get_random_file_set(n_samples: int, files: list[str]) -> list[str]:
...
-    def get_files_to_process(self) -> tuple[list[str], dict[str, float]]:
+    def get_files_to_process(self) -> tuple[list[str], dict[str, float], int]:
         """
         Get files to process
         :return: list of files and a dictionary of the files profile:
             "max_file_size_MB",
             "min_file_size_MB",
             "avg_file_size_MB",
             "total_file_size_MB"
+        and the number of retries
```
daw3rd: I don't know what this means?  Retries to do when reading the files?  The number of retries made in trying to get the list of files?  Something else?

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -62,53 +62,56 @@ def get_random_file_set(n_samples: int, files: list[str]) -> list[str]:
...
-    def get_files_to_process_internal(self) -> tuple[list[str], dict[str, float]]:
+    def get_files_to_process_internal(self) -> tuple[list[str], dict[str, float], int]:
         """
         Get files to process
         :return: list of files and a dictionary of the files profile:
             "max_file_size_MB",
             "min_file_size_MB",
             "avg_file_size_MB",
             "total_file_size_MB"
+        and number of retries
```
daw3rd: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -62,53 +62,56 @@ def get_random_file_set(n_samples: int, files: list[str]) -> list[str]:
...
         """
         pass
 
-    def get_file(self, path: str) -> bytes:
+    def get_file(self, path: str) -> tuple[bytes, int]:
         """
         Get file as a byte array
         :param path: file path
-        :return: bytes array of file content
+        :return: bytes array of file content and number of retries
```
daw3rd: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -121,14 +124,14 @@ def get_folder_files(self, path: str, extensions: list[str] = None, return_data:
...
-    def save_file(self, path: str, data: bytes) -> dict[str, Any]:
+    def save_file(self, path: str, data: bytes) -> tuple[dict[str, Any], int]:
         """
         Save byte array to the file
         :param path: file path
         :param data: byte array
         :return: a dictionary as
         defined https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/put_object.html
-        in the case of failure dict is None
+        in the case of failure dict is None and number of retries
```
daw3rd: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -139,18 +142,18 @@ def get_output_location(self, path: str) -> str:
...
-    def save_table(self, path: str, table: pa.Table) -> tuple[int, dict[str, Any]]:
+    def save_table(self, path: str, table: pa.Table) -> tuple[int, dict[str, Any], int]:
         """
         Save table to a given location
         :param path: location to save table
         :param table: table
         :return: size of table in memory and a dictionary as
         defined https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/put_object.html
-        in the case of failure dict is None
+        in the case of failure dict is None and number of retries
```
daw3rd: ...

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -185,17 +188,19 @@ def sample_input_data(self, n_samples: int = 10) -> dict[str, Any]:
...
         # Pick files to sample
         files = self.get_random_file_set(n_samples=n_samples, files=path_list)
         # Read table and compute number of docs and sizes
         number_of_docs = []
         table_sizes = []
         n_tables = 0
         for f in files:
-            table = self.get_table(path=f)
+            table, r = self.get_table(path=f)
+            retries += r
```
daw3rd, line 212: Ok, so this appears to be the number of retries across ALL files so we lose the details about which file/s is/are a problem.  Do we need that?  Maybe not yet, but at least the doc needs to be clear about what the number is.

daw3rd (changes requested):

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: It's an amount of retries for action execution. That's the only thing that we care about. If the number of retries starts to grow it means that S3 is under pressure

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: same

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: As mentioned above this is a way to detect extra pressure on S3. We want to see if the pressure is up by analyzing the amount of retries

data-processing-lib/python/src/data_processing/data_access/data_access.py
blublinsky: and fixed

roytman:
should we reference issues from the inner repository in the outer one?

blublinsky:
Do not see harm

data-processing-lib/python/src/data_processing/data_access/data_access.py
daw3rd: Changing 1 word is not what I was after.  More of an explanation like you gave in your comment.  From my understanding, I would suggest something like the following:

`The total number of retries in trying to read the list of files.  Retries are performed on read failures, which are typically due to timeouts`

Although ArrowS3 seems to have a max_retries, we do not seem to expose this anywhere.  Probably needs an issue and a separate PR.

data-processing-lib/python/src/data_processing/data_access/data_access.py
daw3rd: I don't really care why at this point.  I just want some clear documentation!

daw3rd:
> should we reference issues from the inner repository in the outer one?

I don't believe we should do that for lack of transparency

daw3rd (approved):
could be better

PR #247, opened 2024-06-09
Make all top level make targets pass w/o error

daw3rd:
Update all top level targets to pass recursive execution.
Especially change test-image in spark transforms to build the required base spark image in data-processing-lib/spark.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,11 +43,11 @@ test::
 	@# Help: Recursively $@ in all subdirs 
 	@$(MAKE) RULE=$@ .recurse
 
-publish::  
+publish::
 	@# Help: Recursively $@ in all subdirs 
 	@$(MAKE) RULE=$@ .recurse
 
-set-versions::  
+set-versions:  
 	@# Help: Recursively $@ in all subdirs 
 	@$(MAKE) RULE=$@ .recurse
 
--- a/transforms/code/code_quality/kfp_ray/Makefile
+++ b/transforms/code/code_quality/kfp_ray/Makefile
@@ -9,6 +9,10 @@ YAML_WF := $(patsubst %.py, %.yaml, ${PYTHON_WF})
 
 workflow-venv: .check_python_version ${WORKFLOW_VENV_ACTIVATE}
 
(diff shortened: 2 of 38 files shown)


Makefile
```diff
@@ -43,11 +43,11 @@ test::
 	@# Help: Recursively $@ in all subdirs 
 	@$(MAKE) RULE=$@ .recurse
 
-publish::  
+publish::
 	@# Help: Recursively $@ in all subdirs 
 	@$(MAKE) RULE=$@ .recurse
 
-set-versions::  
+set-versions:
```
roytman, line 50: @daw3rd , why do we use `::` in other rules, we don't need multiple rules for the same target, each executed independently?
and why do you remove it here? what is the difference?

transforms/code/code_quality/kfp_ray/Makefile
```diff
@@ -19,10 +23,14 @@ test-src::
 
 test-image::
 
+publish::
+
 image::
 
 load-image::
 
+set-versions: workflow-reconcile-requirements
```
revit13, line 32: Does it make sense to rename `workflow-reconcile-requirements` to `set-versions` to reduce the number of rules?

Makefile
daw3rd: The :: allows sub-makefiles to override to extend the function.  I'm not really sure we need this anymore.  However, for set-versions, I was just trying to make it consistent across all makefiles.  I guess when these are defined in the leaf makefiles, they can be whatever the leaf wants to use.  Problems seemed to arise when being inconsistent with : vs :: in makefiles that tried to override/extend one of the .make.* targets.  I'm beginning to think, the .make.* targets should all go back to : so they are thought of more as sub-routines that extendible methods.

transforms/code/code_quality/kfp_ray/Makefile
daw3rd: yes, but I didn't want to risk breaking anything that might have used the workflow target :).

roytman (approved):
LGTM

PR #261, opened 2024-06-11
small changes to get ready for pdf

blublinsky:

There are currently 2 main issues for implementing pdf conversion:
1. PDF conversion is based on 2 files - *.csv and *.pdf. So we do need to process them as a pair. Both files have the same name but two different extensions. The solution is to only read *.csv file and then convert the name to get a pdf extension. Something like this:
```
pdf_name = TransformUtils.get_file_extension(file_name)[0] + ".pdf" 
```
and then read a pdf file

2. As PDF is doing conversion from *.csv to *.parquet, the checkpointing is broken. Here checkpointing is fixed by:
    a. introducing a new parameter `files_to_checkpoint`. default is [.parquet]
    b. comparing just the full names without extensions for the purposes of checkpointing

### DIFF
--- a/data-processing-lib/python/src/data_processing/transform/binary_transform.py
+++ b/data-processing-lib/python/src/data_processing/transform/binary_transform.py
@@ -32,12 +32,12 @@ class AbstractBinaryTransform(AbstractTransform[DATA]):
         """
         self.config = config
 
-    def transform_binary(self, base_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
+    def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
         """
         Converts input file into o or more output files.
         If there is an error, an exception must be raised - exit()ing is not generally allowed.
         :param byte_array: contents of the input file to be transformed.
-        :param base_name: the base name of the file containing the given byte_array.
+        :param file_name: the name of the file containing the given byte_array.
         :return: a tuple of a list of 0 or more tuples and a dictionary of statistics that will be propagated
                 to metadata.  Each element of the return list, is a tuple of the transformed bytes and a string
                 holding the extension to be used when writing out the new bytes.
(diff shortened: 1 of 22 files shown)


dolfim-ibm (approved):
Thanks @blublinsky

data-processing-lib/python/src/data_processing/transform/binary_transform.py
```diff
@@ -32,12 +32,12 @@ def __init__(self, config: dict[str, Any]):
         """
         self.config = config
 
-    def transform_binary(self, base_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
+    def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
```
daw3rd, line 35: I would not make this a primary parameter and it should be optional.  Not every transform will get its data from a file and/or doesn't care about the source of the data.  Some might argue we should have used **kwargs here, but I'm ok if we put it at the end and default it to None.

daw3rd (changes requested):

data-processing-lib/python/src/data_processing/transform/binary_transform.py
blublinsky: Its the framework that invokes this, not the developer. Developer can decide whether to use it or not

daw3rd: and it should be optional.  can you please make the change.  Transforms will not ONLY be used by the table processor/orchestrator.  that's one of the key reasons to split out the python only library/implementations.

daw3rd: and of course, this provides backwards compatibility.  Never a bad bonus.

blublinsky: no, it will break the code. And which backward compatibility are you talking about? The parameter was there, it just changed from the file name to the full name, including directories.

daw3rd: my apologies.  i misunderstood the change as a **new** parameter, not a name/sementic change.

daw3rd (approved):

PR #288, opened 2024-06-17
Split code quality, malware and proglang select transforms into python and ray.

daw3rd:
We need python only versions of our transforms.

#185


daw3rd:
@takuyagt I have split the 3 code transforms into python and ray runtimes, but need your help with python malware docker image.  I left it with the original ray base image to get it to pass ci/cd,  but we should switch to the same python image that other python transforms use.   Please look at the https://github.com/IBM/data-prep-kit/tree/runtime-reorg branch and the malware/python/Dockerfile and see if you can get make test-image to pass after changing the first FROM to FROM [docker.io/python:3.10.14-slim-bullseye](http://docker.io/python:3.10.14-slim-bullseye) .  Thanks.

transforms/code/code_quality/README.md
```diff
@@ -0,0 +1,13 @@
...
+captures code specific metrics of input data.
+Per the set of 
+[transform project conventions](../../README.md#transform-project-conventions)
+the following runtimes are available:
+
+* [python](python/README.md) - provides the base python-based transformation 
+implementation.
+* [ray](ray/README.md) - enables the running of the base python transformation
+in a Ray runtime
+* [kfp](kfp_ray/README.md) - enables running the ray docker image for
```
Param-S: text references `noop`

transforms/code/malware/README.md
```diff
@@ -0,0 +1,14 @@
...
+the input parquet file identifying the virus signature.
+Per the set of 
+[transform project conventions](../../README.md#transform-project-conventions)
+the following runtimes are available:
+
+* [python](python/README.md) - provides the base python-based transformation 
+implementation.
+* [ray](ray/README.md) - enables the running of the base python transformation
+in a Ray runtime
+* [kfp](kfp_ray/README.md) - enables running the ray docker image for
```
Param-S: same here: `noop` word

transforms/code/proglang_select/README.md
```diff
@@ -0,0 +1,14 @@
...
+one of those specified in the transform configuration.
+Per the set of 
+[transform project conventions](../../README.md#transform-project-conventions)
+the following runtimes are available:
+
+* [python](python/README.md) - provides the base python-based transformation 
+implementation.
+* [ray](ray/README.md) - enables the running of the base python transformation
+in a Ray runtime
+* [kfp](kfp_ray/README.md) - enables running the ray docker image for
```
Param-S: same here: noop word

transforms/code/code_quality/README.md
daw3rd: done. thanks.

transforms/code/malware/README.md
daw3rd: done. thanks.

transforms/code/proglang_select/README.md
daw3rd: done. thanks.

transforms/code/code_quality/python/src/code_quality_transform.py
```diff
@@ -0,0 +1,334 @@
...
+            .append_column("avg_longest_lines", [avg_longest_lines_values])
+            .append_column("alphanum_frac", [alphanum_frac_values])
+            .append_column("char_token_ratio", [char_token_ratio_values])
+            .append_column("autogenerated", [is_autogenerated_values])
+            .append_column("config_or_test", [is_config_or_test_values])
+            .append_column("has_no_keywords", [has_no_keywords_values])
+            .append_column("has_few_assignments", [has_few_assignments_values])
+            .append_column("is_xml", [is_xml_values])
+            .append_column("is_html", [is_html_values])
+        )
```
blublinsky, line 279: Lines 247 till this one are duplicating code here TransformUtils.add_column. using this method can remove half of the code

transforms/code/code_quality/ray/Dockerfile
```diff
@@ -9,13 +9,19 @@ COPY --chown=ray:users data-processing-lib-python/ data-processing-lib-python/
 RUN cd data-processing-lib-python && pip install --no-cache-dir -e .
 COPY --chown=ray:users data-processing-lib-ray/ data-processing-lib-ray/ 
 RUN cd data-processing-lib-ray    && pip install --no-cache-dir -e .
+COPY --chown=ray:users python-transform/  python-transform/
```
blublinsky, line 12: pls, make sure its the right Ray version. should be 2.24.0

transforms/code/code_quality/ray/Dockerfile
```diff
@@ -9,13 +9,19 @@ COPY --chown=ray:users data-processing-lib-python/ data-processing-lib-python/
...
+RUN cd python-transform && pip install --no-cache-dir -e .
 
-COPY requirements.txt requirements.txt
-RUN pip install --no-cache-dir -r  requirements.txt
+#COPY requirements.txt requirements.txt
+#RUN pip install --no-cache-dir -r  requirements.txt
+
+COPY --chown=ray:users src/ src/
+COPY --chown=ray:users pyproject.toml pyproject.toml 
+RUN pip install --no-cache-dir -e .
```
blublinsky, line 20: Do you have to explicitely install here libraries, that this guys depends on?

transforms/code/malware/ray/Dockerfile
```diff
@@ -33,13 +33,16 @@ COPY --chown=ray:users data-processing-lib-python/ data-processing-lib-python/
 RUN cd data-processing-lib-python && pip install --no-cache-dir -e .
```
blublinsky, line 33: Pls make sure that it is derived from the proper Python version

transforms/code/proglang_select/python/src/proglang_select_transform.py
```diff
@@ -0,0 +1,172 @@
...
+        # else:
+        #     # This is recommended for production approach. In this case domain list is build by the
+        #     # runtime once, loaded to the object store and can be accessed by actors without additional reads
+        #     try:
+        #         logger.info(f"Loading languages to include from Ray storage under reference {languages_include_ref}")
+        #         self.languages_include = ray.get(languages_include_ref)
+        #     except Exception as e:
+        #         logger.info(f"Exception loading languages list from ray object storage {e}")
+        #         raise RuntimeError(f"exception loading from object storage for key {languages_include_ref}")
+
```
blublinsky, line 78: Well this is cheating. You are removing functionality important for Ray. I would suggest to move the correct init to Ray and Python version and remove it from here

transforms/code/proglang_select/ray/src/proglang_select_transform_ray.py
```diff
@@ -161,87 +72,10 @@ def get_transform_config(
             data_access=lang_data_access_factory.create_data_access(),
             logger=self.logger,
         )
-        lang_refs = ray.put(list(lang_list))
-        self.logger.info(f"Placed language list into Ray object storage under reference{lang_refs}")
-        return {lang_allowed_languages: lang_refs} | self.params
-
```
blublinsky, line 167: Why is this removed. In the case of Ray its the right thing to do. I know you did it because of Python version, but this is a good case why deriving from Python leads to issues

transforms/code/code_quality/python/src/code_quality_transform.py
daw3rd: Separate PR please - I'm not the one responsible for this transform.  this is about restructuring.

transforms/code/code_quality/ray/Dockerfile
daw3rd: I have checked around and all looks fine. Please point out the location of a real (not presumed) problem.

transforms/code/code_quality/ray/Dockerfile
daw3rd: They are now in the pyproject.toml, moved from requirements.txt

transforms/code/malware/ray/Dockerfile
daw3rd: again, please point to a specific (not presumed) problem.

transforms/code/proglang_select/python/src/proglang_select_transform.py
daw3rd: It seems reasonable to me to put the data read by the runtime config into the config dictionary instead of ray storage. the config dictionary is passed to workers and this is then more generic.  We still avoid having each worker read the data. in the end the data is passed through the network from orchestrator to worker. it might even be faster than putting in ray storage and pulling out of ray storage to just pass it once over the network as a dictionary.

daw3rd: Instead of putting it in ray object storage, i'm putting it in the config dictionary (in the ray runtime object apply_inputs() method.  This still avoid having each worker read the file, which I believe is the goal in the end.  In fact, putting it in ray object storage may be less efficient then just passing it in the already passed config dict.

transforms/code/proglang_select/ray/src/proglang_select_transform_ray.py
daw3rd: This is done with the previous comment in mind (see previous response).

transforms/code/proglang_select/python/src/proglang_select_transform.py
blublinsky: It will work fine for the small amounts of data, but can be quite expensive for the last data sizes. As for the network traffic its a wash, as calls from the orchestrator is over the network. Using Ray object storage is a well defined pattern that we are promoting. Please restore it

daw3rd: We should really allow self.languages_include to be set from the configuration as part of the dictionary config for pure python.  You say Ray is better for large objects,  yet you concur that the network traffic is a wash, so what's the problem.   Actually, I believe this is a better pattern for ray runtime to configure pure python.  Enforcing a new class (i.e. RayProgLangSelectTransform) to use Ray object storage and to match a pattern seems unnecessary and like jumping through unnecessary hoops.

blublinsky: I might be not clear. What I am saying is that we should allow for different implementation between pure Python and Ray. Moreover I am suggesting a clean way to do this by implementing different init between Python and Ray. We had this approach there and working for a long time. Its not jumping through hoops. Its implementing the relevant mechanism for relevant runtimes

blublinsky (approved):

PR #295, opened 2024-06-18
Move hack dirs to scripts dir

### LINKED ISSUE #293: [Feature] Restructure the hack directories
### Search before asking

### Component

Other

### Feature

We have the `transformers/hack` directory, which contains a single script, we need to move it to the `scripts` directory.
Check the possibility to do the same with the `kind/hack` 

revit13:
/Closes #293

This PR moves the hack subdirs under a new `scripts` directory

#293


roytman (approved):
LGTM

scripts/kind/tools/ingress.sh
```diff
@@ -2,19 +2,19 @@
 
 op=$1
 
-source ${ROOT_DIR}/hack/common.sh
+source ${REPOROOT}/scripts/kind/common.sh
```
roytman: can you define a local variable `{REPOROOT}/scripts/kind`, we have too many references to it.

scripts/kind/tools/ingress.sh
```diff
@@ -2,19 +2,19 @@
...
 
 deploy() {
-  kubectl apply -f ${ROOT_DIR}/hack/ray_api_server_ingress.yaml
+  kubectl apply -f ${REPOROOT}/scripts/kind/ray_api_server_ingress.yaml
```
roytman: can you add sleep 10 before L:8

transforms/.make.workflows
daw3rd, line 1: This is an inconsistent move.  A) it is not a shell script and B) if we do this, we should also move .make.defaults, .make.versions,.make.transforms etc.  But I'm not sure we want to do this.  What was your motivation here?

daw3rd (changes requested):

transforms/.make.workflows
revit13: @daw3rd transforms/.make.workflows file remains under transforms... are you refering to other file?

scripts/kind/tools/ingress.sh
revit13: Added. Thanks

scripts/kind/tools/ingress.sh
revit13: fixed. Thanks

daw3rd (approved):
lgtm

Issue #294, opened 2024-06-18
[Feature] Define Release process


roytman:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

The project needs a release process, first, it can be done manually, but after that, it should be based on GitHub actions.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

daw3rd:
This is almost complete.  We have merged in release-notes.md, RELEASE.md and scripts/release.sh.  Still a bit of work going on in a new PR #337 to get kfp publishing both v1 and v2 and to refine docs and scripts.

PR #297, opened 2024-06-18
Release process definition and implementation

roytman:

We need to define and after that automate the project release process.

https://github.com/IBM/data-prep-kit/issues/294

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,61 @@
+# Release Process
+
+The process of creating a release is described in this document.
+Note: below we list the manual steps to create a new release, later some of them will be automated by GitHub actions.
+
+Replace `X.Y.Z` with the version to be released.
+
+## 1. Update the .make.versions file
+
+**Question:** should we use the same versions for release, and for shared python modules and images, such as 
+`DPK_LIB_VERSION`, `DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2`?
+If YES, maybe we can combine all these variables into a single one.
+
+- In the [.make.versions](.make.versions) file update versions of the shared components `DPK_LIB_VERSION`, 
+`DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2` to the desired ones  
+- Update transformers versions
+- Set `RELEASE_VERSION_SUFFIX` to the empty string, or comment out the entry.
+
+## 2. Run local tests
+TODO
+## 3. Create a `releases/X.Y.Z` branch
+The `releases/X.Y.Z` branch should be created from a base branch. 
+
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+The process of creating a release is described in this document.
+Note: below we list the manual steps to create a new release, later some of them will be automated by GitHub actions.
+
+Replace `X.Y.Z` with the version to be released.
+
+## 1. Update the .make.versions file
+
+**Question:** should we use the same versions for release, and for shared python modules and images, such as 
+`DPK_LIB_VERSION`, `DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2`?
+If YES, maybe we can combine all these variables into a single one.
```
revit13, line 12: Does it makes sense to have `0.0.0` tag to all in dev branch and update it to the release tag upon creating a new release in the release branch (assuming their tag version is as the relase version tag)? there is no meaning to the tag version in dev as we build everything locally.

RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+
+- In the [.make.versions](.make.versions) file update versions of the shared components `DPK_LIB_VERSION`, 
+`DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2` to the desired ones  
+- Update transformers versions
+- Set `RELEASE_VERSION_SUFFIX` to the empty string, or comment out the entry.
+
+## 2. Update versions
+From the project root directory execute
+```shell
+make versions
```
revit13, line 22: I think its `make set-versions`

RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+
+## 1. Update the .make.versions file
+
+**Question:** should we use the same versions for release, and for shared python modules and images, such as 
+`DPK_LIB_VERSION`, `DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2`?
+If YES, maybe we can combine all these variables into a single one.
+
+- In the [.make.versions](.make.versions) file update versions of the shared components `DPK_LIB_VERSION`, 
+`DPK_LIB_KFP_VERSION`, `DPK_LIB_KFP_VERSION_v2`, `DPK_LIB_KFP_SHARED`, `KFP_DOCKER_VERSION`, `KFP_DOCKER_VERSION_v2` to the desired ones  
+- Update transformers versions
```
revit13, line 16: Its not clear how the transforms release process is done. should it be independent of the repo release? if so, we should not update the versions in this process.

RELEASE.md
revit13: BTW I think that the process of updating the versions in .make.versions should be done in the release branch so the .make.versions in dev branch should remain `0.0.0`...

roytman: > there is no meaning to the tag version in dev as we build everything locally.

it is not exactly true, for pipeline tests on a real cluster we need relevant images.

PR #301, opened 2024-06-18
move logger to ensure Ray logging is correct

blublinsky:
]Ensures correct Ray logging


transforms/code/proglang_select/ray/src/proglang_select_transform_ray.py
```diff
@@ -48,6 +46,8 @@
 
 
 def _get_supported_languages(lang_file: str, data_access: DataAccess) -> list[str]:
+    from data_processing.utils import get_logger
+    logger = get_logger(__name__)
```
daw3rd: maybe the logger should be passed in as an argument.  Otherwise, might this be different than what the transform has?

daw3rd (approved):

transforms/code/proglang_select/ray/src/proglang_select_transform_ray.py
blublinsky: good idea

PR #309, opened 2024-06-19
Some new thoughts on cutting a release, especially scripts/release.sh

daw3rd:
We need to automate the process of version management and release cutting.  This is just a start with some initial thoughts based on recent discussion.  Initial changes include

1. Change .make.versions to define a common repo-wide version for all libraries and images.
2. Add scripts/release.sh which tries to outline the steps needed to changes versions, create and tag a branch.
3. Makefile new target to `show-version` based on .make.versions content
4. Fixed a lot of set-versions problems
5. Set all versions back to 0.2.0.dev6

Relies on `make build publish` being enough to actually publish the release to pypi and quay.io

Also see PR #297 

#294


transforms/universal/profiler/kfp_ray/Makefile
```diff
@@ -29,8 +29,6 @@ image::
 
 load-image::
 
-set-versions: workflow-reconcile-requirements
```
revit13: Please rebase, this should be fixed now in dev branch. thanks

roytman:
@daw3rd please do not do the release automation before we agree on the process. Some of the functionality can be done by GitHub actions

transforms/universal/profiler/kfp_ray/Makefile
daw3rd: done

daw3rd:
> @daw3rd please do not do the release automation before we agree on the process. Some of the functionality can be done by GitHub actions

I didn't think I was, except that I added the make tests for the test, clean, publish, build, set-versions top-level rules and some Makefiles needed to implement these

transforms/universal/profiler/ray/Makefile
```diff
@@ -28,9 +28,8 @@ publish:: publish-dist publish-image
 
 publish-image:: .transforms.publish-image-ray
 
-# distribution versions is the same as image version.
 set-versions:
-	$(MAKE) TOML_VERSION=$(DOCKER_IMAGE_VERSION) .defaults.update-toml
+	$(MAKE) TRANSFORM_PYTHON_VERSION=$(DOCKER_IMAGE_VERSION) .transforms.set-versions
```
revit13: I do not think that's working , `TRANSFORM_PYTHON_VERSION` here is the value of `PROFILER_RAY_VERSION`  while it should be `PROFILER_PYTHON_VERSION`. I tried to fix that in https://github.com/IBM/data-prep-kit/pull/323

roytman:
We should deploy 2 kfp-data-processing images one for KFPv1 and another for KFPv2. Will try to update the script to proceed with that.

transforms/universal/profiler/ray/Makefile
daw3rd: You are correct.  To be more clear i set to `...=not-used` when there is a no python versions (profiler, e/fdedup, doc_id).  code_quality, malware and proglang_select did need the change.

HOWEVER, in the end since all version numbers are now the same, it was working ok :)

.make.versions
```diff
@@ -28,43 +41,47 @@ else
...
-PROGLANG_SELECT_RAY_VERSION=0.4.0$(RELEASE_VERSION_SUFFIX)
-PROGLANG_SELECT_PYTHON_VERSION=0.4.0$(RELEASE_VERSION_SUFFIX)
+PROGLANG_SELECT_PYTHON_VERSION=$(DPK_VERSION)
+PROGLANG_SELECT_RAY_VERSION=$(DPK_VERSION)
 
-CODE_QUALITY_PYTHON_VERSION=0.4.0$(RELEASE_VERSION_SUFFIX)
-CODE_QUALITY_RAY_VERSION=0.4.0$(RELEASE_VERSION_SUFFIX)
+CODE_QUALITY_RAY_VERSION=$(DPK_VERSION)
+CODE_QUALITY_PYTHON_VERSION=$(DPK_VERSION)
+CODE_QUALITY_RAY_VERSION=$(DPK_VERSION)
```
revit13: It appears twice.

revit13 (approved):

PR #313, opened 2024-06-20
remove unused Make targets

roytman:

As part of the simplification process, this PR remove some unused (empty) Make targets

### DIFF
--- a/.make.defaults
+++ b/.make.defaults
@@ -129,12 +129,18 @@ help::
 	    for i in $$SUB_MAKE_DIRS; do	\
 		if [ -e $$i/Makefile ]; then	\
 		    echo Using recursive $(RULE) rule in $$i;	\
-		    (cd $$i; $(MAKE) $(RULE));	\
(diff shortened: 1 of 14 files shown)


revit13:
@roytman LGTM btw is there a way to suppress printing? `@` did not work for me...

revit13 (approved):

daw3rd (commented):
If we're still using recursive, from the top, test, build set-versions, this is will now throw errors when doing these from the top.  DId you try thta?

roytman:
> If we're still using recursive, from the top, test, build set-versions, this is will now throw errors when doing these from the top. DId you try thta?

Yes I tried it, see the updated recursion target

daw3rd:
I think we're on another take with the PR #309 .  Can we close this?

roytman:
> I think we're on another take with the PR #309 . Can we close this?

did you mean https://github.com/IBM/data-prep-kit/pull/297 ? or both :-)

PR #316, opened 2024-06-21
renaming of ingest_2_parquet

blublinsky:

Update Ray version to the current one


blublinsky:
Readme need some work, but this is for the next PR. For now I just copied the one from tools

daw3rd (approved):

PR #322, opened 2024-06-24
add new params to lang_id to store the results of language identification

dtsuzuku-ibm:

To validate the non-existence of two columns and let user specify the column name.

### DIFF
--- a/transforms/language/lang_id/python/src/lang_id_transform.py
+++ b/transforms/language/lang_id/python/src/lang_id_transform.py
@@ -26,10 +26,14 @@ model_credential_key = "model_credential"
 model_kind_key = "model_kind"
 model_url_key = "model_url"
 content_column_name_key = "content_column_name"
+output_lang_column_name_key = "output_lang_column_name"
+output_score_column_name_key = "output_score_column_name"
 model_credential_cli_param = f"{cli_prefix}{model_credential_key}"
 model_kind_cli_param = f"{cli_prefix}{model_kind_key}"
 model_url_cli_param = f"{cli_prefix}{model_url_key}"
 content_column_name_cli_param = f"{cli_prefix}{content_column_name_key}"
+output_lang_column_name_cli_param = f"{cli_prefix}{output_lang_column_name_key}"
+output_score_column_name_cli_param = f"{cli_prefix}{output_score_column_name_key}"
 
 
 class LangIdentificationTransform(AbstractTableTransform):
@@ -53,7 +57,9 @@ class LangIdentificationTransform(AbstractTableTransform):
         self.nlp_langid = LangModelFactory.create_model(
             config.get(model_kind_key), config.get(model_url_key), config.get(model_credential_key)
         )
-        self.column_name = config.get(content_column_name_key)
+        self.content_column_name = config.get(content_column_name_key)
+        self.output_lang_column_name = config.get(output_lang_column_name_key)
+        self.output_score_column_name = config.get(output_score_column_name_key)
 
     def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Table], dict[str, Any]]:
         """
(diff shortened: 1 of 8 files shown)


dtsuzuku-ibm:
@daw3rd Could you review this PR?

transforms/language/lang_id/python/src/lang_id_transform.py
```diff
@@ -53,7 +57,9 @@ def __init__(self, config: dict[str, Any]):
         self.nlp_langid = LangModelFactory.create_model(
             config.get(model_kind_key), config.get(model_url_key), config.get(model_credential_key)
         )
-        self.column_name = config.get(content_column_name_key)
+        self.content_column_name = config.get(content_column_name_key)
```
daw3rd: You should not assume the config map will come from the CLI and apply_input_params() below.  As a result, you really should check for missing keys here and/or provide the defaults as appropriate.

transforms/language/lang_id/python/src/nlp.py
```diff
@@ -21,18 +21,14 @@
...
-                )
+    detected_language = pa.Table.from_pylist(
+        list(
+            map(
+                lambda r: {"lang": r[0], "score": r[1]},
+                map(lambda x: nlp.detect_lang(x), table[col_name].to_pylist()),
             )
         )
-    except Exception as e:
-        logger.warning("ERROR: %s, skipping the file", e)
```
daw3rd, line 34: You should throw the exception back out through transform().  The runtime will catch it and also issue a message, but we (i.e. the runtime) wants to be told about these failures.

transforms/language/lang_id/ray/src/lang_id_transform_ray.py
```diff
@@ -27,10 +27,14 @@
 model_kind_key = "model_kind"
 model_url_key = "model_url"
 content_column_name_key = "content_column_name"
+output_lang_column_name_key = "output_lang_column_name"
+output_score_column_name_key = "output_score_column_name"
```
daw3rd: Why are you redefining these here?  They should be the same as in the python transform right?  If so, use them instead.

daw3rd: Remember that this ray-runtime deployment of the python transform depends on the adjacent python source and is installed both in the venv and the docker image.

daw3rd (changes requested):

transforms/language/lang_id/python/src/nlp.py
dtsuzuku-ibm: Oops, thank you.

transforms/language/lang_id/ray/src/lang_id_transform_ray.py
dtsuzuku-ibm: I completely agree with your comment.
I just followed the pattern in noop. I thought that's the code conventions.

transforms/language/lang_id/python/src/lang_id_transform.py
dtsuzuku-ibm: Oh, ok. I'll set default value.

transforms/language/lang_id/python/src/nlp.py
dtsuzuku-ibm: Since I removed the code to catch exception, I suppose the updated code will throw any exception from get_lang_ds_pa() and consequently transform(). Am I missing something?

transforms/language/lang_id/ray/src/lang_id_transform_ray.py
daw3rd: Thanks for pointing this out.  My mistake in noop (and a couple of others).  Fixed in PR #336.

transforms/language/lang_id/python/src/lang_id_transform.py
daw3rd: Looks good.  One minor point, can you define the default values ("score", "content"0, which are used both in init() and in add_input_params(), using global variables and then reference those in the two methods?  A bit of extra indirection, but less prone to developer error at a later point in time. (I'm trying to promote this pattern with others as well).

dtsuzuku-ibm: Ok, I'll define global variable.

daw3rd (approved):
Thanks for the changes!

PR #323, opened 2024-06-24
Makefile rules updates

revit13:

This PR contains the following changes:
	- It renames `workflow-build` Makefile target to `setup` and `.workflows.set-versions` to `set-versions`
	- Fixes to rules like `.defaults.update-toml-lib-dep-versions`
	- It moves KFP and RAY versions to .make.versions and update `set-versions` target to update the files based on their versions
	- Fix clean target in kfp_ray directory
	- Fixes `.transforms.set-versions` and changes `set-versions` target in the transforms to use 
.transforms.set-versions


blublinsky (approved):

.make.versions
```diff
@@ -13,8 +13,19 @@ RELEASE_VERSION_SUFFIX=.dev6
...
-DPK_LIB_KFP_VERSION_v2=0.2.0$(RELEASE_VERSION_SUFFIX)
-DPK_LIB_KFP_SHARED=0.2.0$(RELEASE_VERSION_SUFFIX)
+
+KFP_DOCKER_VERSION=0.2.0$(RELEASE_VERSION_SUFFIX)
+KFP_DOCKER_VERSION_v2=0.2.0$(RELEASE_VERSION_SUFFIX)
+
+KFP_v2=2.7.0
+KFP_v1=1.8.22
+RAY=2.24.0
+
```
roytman, line 23: we don't define here the versions of kuberay operator and kuberay API server?

roytman (approved):
LGTM

Issue #238, opened 2024-06-06
Add a README file to the new Ingest2parquet transform 


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

A README file is needed in this directory: transforms/code/ingest_2_parquet/ray (as consistent with all other transforms)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

blublinsky:
@shahrokhDaijavad, I do not think it is for me. It probably should go to Sapta

shahrokhDaijavad:
Fair enough, @blublinsky. Let me assign @sapthasurendran, since it could be some simple editing of the Readme file that is already in the tools directory for this, i.e., tools/ingest2parquet/README.md

blublinsky:
@shahrokhDaijavad can you also ask her to remove the tools directory once she is happy with the current implementation

shahrokhDaijavad:
Yes, @blublinsky I agree that we should not have 2 modules doing the same thing. @Bytes-Explorer and @sapthasurendran, If after a review of the new ingest2parquet by Saptha, she is happy with it, the old code in the tools directory should be removed.

sapthasurendran:
@blublinsky could you please point me to the new ingest2parquet ?

blublinsky:
https://github.com/IBM/data-prep-kit/tree/dev/transforms/code/ingest_2_parquet

sapthasurendran:
@blublinsky Looking at the project structure and the readme  [here](https://github.com/IBM/data-prep-kit/blob/dev/transforms/universal/noop/README.md) , python and ray transforms are in different directories. In the latest [ingest_2_parquet](https://github.com/IBM/data-prep-kit/tree/dev/transforms/code/ingest_2_parquet)  these files are placed inside a single folder. For consistency can we have separate folders for different runtime?

blublinsky:
thank you @sapthasurendran, one of the questions for you is whether we should remove tools directory completely?
Also feel free to do any code refactoring

Bytes-Explorer:
We need to retain the tools directory for other things that are not transforms and other modules that will be added soon. I still propose to keep this under Tools, though I don't have a very strong opinion on it, so can go with what the group suggests.

blublinsky:
[...]

PR #329, opened 2024-06-24
Improvements to code2parquet transform

daw3rd:
We now only have a single python transform that can run independent of the python runtime.
Code2ParquetTransform class could not be run itself and required the python-runtime-specific sub-class to initialize some fields.  

1. Fix readmes.
2. Remove Code2PargetPythonTransform.
3. Change code2ParquetTransfrom initializer to do a better job of error checking and allow for the map of languages to be added. 
4. Define constants for keys and cli keys and some default constants
5. Don't print s3 keys
6. Continue to parse but deprecate domain and snapshot cli params, which were  never used.

#238


sapthasurendran (commented):
domain and snapshot keys are required for code datasets. Please add it back.

shivdeep-singh-ibm (approved):

transforms/code/code2parquet/python/README.md
```diff
@@ -66,136 +69,60 @@ Each file contained within the ZIP is transformed into a distinct row within the
...
+  --code2parquet_supported_langs_file CODE2PARQUET_SUPPORTED_LANGS_FILE
+                        Path to file containing the list of supported languages
+  --code2parquet_detect_programming_lang CODE2PARQUET_DETECT_PROGRAMMING_LANG
+                        generate programming lang
+ ```
+
+These correspond to the configuration keys described above.
+
+### Running the samples
+To run the samples, use the following `make` targets
```
sapthasurendran, line 104: Please add make venv command before running the below commands.

daw3rd: This is a good point, although when running `make run-cli-sample`, for example, it prints an error message indicating `make venv` needs to be run.

daw3rd:
> domain and snapshot keys are required for code datasets. Please add it back.

But these were never used in transform().  THey were parsed and I left that in, just to keep kfp happy.

PR #330, opened 2024-06-24
Corrected Readme to update file path, added more detail signoff steps

santoshborse:
- Correct the python file path in readme
- Adding more detail instruction to contributing doc


daw3rd (approved):

daw3rd:
Thank you!

PR #331, opened 2024-06-24
improve doc on transform design/expectations

daw3rd:
Clarity


shahrokhDaijavad (approved):
I like these clarifications.

PR #335, opened 2024-06-25
Build libs for kfp v1 and v2 in the release.

Mohammad-nassar10:
Add building and publishing the two kfp images of kfp v1 and v2 when creating a new release.

### DIFF
--- a/kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
+++ b/kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
@@ -40,11 +40,16 @@ endif
 build-dist :: set-versions .defaults.build-dist
 
 publish:: .check-env
+ifeq ($(KFPv2), 1)
+	# we want to prevent execution of the rule, when we run `make build` in upper directories and KFPv2==1
+	echo "Skipping publish as KFPv2 is defined"
+else
 	@# Help: Publish the wheel to testpypi
 	if [ -d "dist"]; then rm -r dist; fi
 	${PYTHON} -m pip install --upgrade build
 	${PYTHON} -m twine check dist/*
 	${PYTHON} -m twine upload --verbose --non-interactive dist/*
+endif
 
 .PHONY: venv
 venv:
--- a/kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
+++ b/kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
@@ -29,37 +29,22 @@ set-versions:
 	mv tt.toml pyproject.toml
 
 build:: set-versions venv
-ifeq ($(KFPv2), 1)
-	# we want to prevent execution of the rule, when we run `make build` in upper directories and KFPv2==1
-	echo "Skipping build as KFPv2 is defined"
-else
 	@# Help: Build the distribution for publishing to a pypi
 	$(MAKE) build-dist
-endif
 
 build-dist :: set-versions .defaults.build-dist
 
 publish::
-ifeq ($(KFPv2), 1)
-	# we want to prevent execution of the rule, when we run `make publish` in upper directories and KFPv2==1
-	echo "Skipping publish as KFPv2 is defined"
-else
 	$(MAKE) publish-dist
-endif
 
 publish-dist :: .check-env .defaults.publish-dist
 
 .PHONY: venv
 venv:
-ifeq ($(KFPv2), 1)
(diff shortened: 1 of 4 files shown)


kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
```diff
@@ -39,12 +39,15 @@ endif
 
 build-dist :: set-versions .defaults.build-dist
 
-publish:: .check-env
-	@# Help: Publish the wheel to testpypi
-	if [ -d "dist"]; then rm -r dist; fi
-	${PYTHON} -m pip install --upgrade build
-	${PYTHON} -m twine check dist/*
-	${PYTHON} -m twine upload --verbose --non-interactive dist/*
+publish::
```
roytman, line 37: Why do we need it? We can publish the Python modules independently on defined KFP version

scripts/release.sh
```diff
@@ -76,6 +76,12 @@ git push origin $tag
 # Requires quay credentials in the environment, DPL_DOCKER_REGISTRY_USER, DPK_DOCKER_REGISTRY_KEY
 if [ -z "$debug" ]; then
     make build publish
+    export KFPv2=1
+    make set-versions -C kfp
+    make build publish -C kfp/kfp_support_lib/kfp_v2_workflow_support/
+    make build publish -C kfp/kfp_ray_components/
```
roytman, line 81: only this part is KFP version dependent and should be republished

kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
Mohammad-nassar10: The `build` rule in v1 library has this if statement, so if the kfp version is v2 it will not build the v1 lib. Thus, we added the same if statement for `publish`.
We can remove this check from both `build` and `publish`.

revit13 (approved):

roytman (approved):
LGTM

kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
daw3rd: Why are we building/publishing only one version of kfp at a time.  If the repo contains both, the top level `make publish` should publish both, no?

daw3rd: I'm trying to shoot for a small set of comprehensive make targets that don't need configuration, so that for example, `make publish` publishes everything that is publishable from the repo.

daw3rd (changes requested):
Holding off merge until we closed on questions I've raised elswhere on this PR.

kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
roytman: the Python libraries are published once. 
However, the kfp image depends on which library should be used, so it should be built and published twice.

roytman:
> Holding off merge until we closed on questions I've raised elswhere on this PR.

see my answer.

daw3rd:
@Mohammad-nassar10 - @roytman and I spoke and agreed that 

1. the publishing of both kfp versions should be handled inside kfp/Makefile - transparently to the top level Makefile
2. We will remove/disable the actual `make build publish` in the release script and handle that either manually in the new branch or with a git action somehow.

As a result I think this PR can be closed.  Another one #338 is started to address those (and your fixes made here)

kfp/kfp_support_lib/kfp_v1_workflow_support/Makefile
blublinsky: We have 3 targets - shared library and 2 KFP libraries V1 and V2. As we are building each one from the source we can build all 3 simultaneously

scripts/release.sh
```diff
@@ -76,6 +76,11 @@ git push origin $tag
 # Requires quay credentials in the environment, DPL_DOCKER_REGISTRY_USER, DPK_DOCKER_REGISTRY_KEY
 if [ -z "$debug" ]; then
     make build publish
+    export KFPv2=1
```
blublinsky, line 79: Why do we care about quay credentials here? Its publishing to PyPi

daw3rd: yes, i removed this from the renamed release-branch.sh script which now does no publishing :)

PR #346, opened 2024-06-26
Fix Instruction to create NOOP transformer

santoshborse:
Follow tutorial to run the NOOP transform, fixed the read me with uptodate instructions


daw3rd (changes requested):
LGTM, except I think we spoke about similar changes also needed in the advanced transform tutorial. Can you add those too?

daw3rd (approved):

PR #347, opened 2024-06-26
code2parquet fixes on domain/snapshot and document_id

daw3rd:
Restore domain and snapshot column creation.
Set document_id column back to a uuid (was a hash). 
Update readme on configuration.

Restore behavior equivalent to tools/ingest2parquet.


shivdeep-singh-ibm (approved):
Looks good to me.

Issue #345, opened 2024-06-26
[Feature] Replace compilation of the KFP workflow definitions  from "setup" to "build"


roytman:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP workflows

### Feature

Currently, KFP workflow definition compiled as a result of the `setup` Make target, it should be replaced by `build`

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

revit13:
closed with https://github.com/IBM/data-prep-kit/pull/348

PR #348, opened 2024-06-27
Add workflow-build target

revit13:

This PR adds workflow-build target in transforms directory.
The `setup` target which is now used to build the workflows is left empty.

#345


roytman (approved):
LGTM

PR #354, opened 2024-06-27
Update release-branch script to not verify commits to avoid failures

daw3rd:
Reverting versions to 0.2.0.dev6 which was set to 0.2.1.dev0 on the last run of the release-branch.sh script
Fix release-branch-script.sh to not verify commits in case someone checked in one of the files with version changes, w/o pre-commit checks.


shahrokhDaijavad (approved):
Simple changes look good.

PR #359, opened 2024-06-27
Fixing Tutorial to run Python and Ray versions correctly

santoshborse:
Fixing Tutorial to run Python and Ray versions correctly


daw3rd (approved):
LGTM

Issue #355, opened 2024-06-27
[Bug] set-versions target in doc_id,e/fdedup,profiler is empty and not setting versions


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

CI/CD

### What happened + What you expected to happen

Running `make set-versions` does not apply .make.versions content to these transforms

- [ ] doc_id
- [ ] profiler
- [ ] ededup
- [ ] fdedup

### Reproduction script

make set-versions in doc_id/ray dir

### OS

MacOS (limited support)

### Python

Other

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

daw3rd:
Fixed in PR #360

PR #360, opened 2024-06-27
fix 'make set-versions' for doc_id, e/fdedup transforms

daw3rd:
`make set-versions` was not working in profiler, doc_id, e/fdedup ray transform directories.

Fixes issue #355


shahrokhDaijavad (approved):
LGTM

PR #364, opened 2024-06-29
Create new transform to ingest markdown (.md) files and convert to parquet format

bogdanscode:
Convert .md files to parquet files so that they can be processed by data prep pipeline
This is the preferred input for InstructLab

178

### DIFF
--- a/tools/ingest2parquet/src/ingest2parquet.py
+++ b/tools/ingest2parquet/src/ingest2parquet.py
@@ -74,52 +74,102 @@ def zip_to_table(data_access: DataAccess, file_path, detect_prog_lang: Any) -> p
     return table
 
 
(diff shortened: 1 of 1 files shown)


daw3rd (changes requested):
Thank you for the contribution.  This ingest2parquet tool is destined to be deprecated in favor of the new code2parquet transform which allows for more scalable conversions (i.e. think TB of data).  I might suggest using code2parquet as a template to create a new transform, name it markdown2parquet and put it in a new directory called transforms/language/markdown2parquet.  Let me know if you have questions

daw3rd:
Also, in the future can you sign your commits?

daw3rd:
Oh and the code2parquet transform is in transforms/code/code2parquet

PR #368, opened 2024-07-01
Fix set-versions target in kfp/kfp_ray_components.

revit13:
Fix set-versions target in kfp/kfp_ray_components to update the image tag version.


roytman (approved):
LGTM

Issue #386, opened 2024-07-07
[Bug] Wrong output folder when output_folder of DataAccess has a different absolute path


roytman:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core

### What happened + What you expected to happen

If a local DataAccess for a Transformer has the following  input and output folders:
```
input_folder = ./test
output_folder=/tmp/output
```
And the current execution directory is `/Users/home/foo`.

The output files will be generated in `/Users/home/foo/tmp/output` instead of `/tmp/output`
BTW, the `metadata.json` file will be in the right location.

### Reproduction script

For example, execute ./transforms/universal/filter/python/src/filter_local_python.py when [input_folder](https://github.com/IBM/data-prep-kit/blob/3623632c2eb2b8105df50013ae6a291b10454833/transforms/universal/filter/python/src/filter_local_python.py#L27) is `../test-data/input` and [output_folder](https://github.com/IBM/data-prep-kit/blob/3623632c2eb2b8105df50013ae6a291b10454833/transforms/universal/filter/python/src/filter_local_python.py#L28) is `/tmp/output`

### Anything else

Every time

### OS

Ubuntu

### Python

3.10.x

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

roytman:
done by https://github.com/IBM/data-prep-kit/pull/387

PR #387, opened 2024-07-07
fix relative input path

roytman:

Fixes situations when one of DataAccess local input or output folders is relative and another is absolute.  
 #386


daw3rd (approved):
lgtm

Issue #317, opened 2024-06-21
[Feature] Enable processing of local data by mounting local directories into the transform images


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

It will be useful if someone can take an image and run the image to apply the transform to data in the users' local file system, creating the transformed data in the local file system as well.  This can be done by mounting the input and output directories into the image and passing --data_local_config options to reference the mount points within the image.  However, do do this, we need to have locations in the image on which to mount the input and output directories.  To that end, all transforms images need a common pair of mount points.  This can be done in their Dockerfiles.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

daw3rd:
We should also have some documentation and/or samples showing how to do this.

blublinsky:
It will work without any changes.

daw3rd:
This is addressed with pr #394

PR #394, opened 2024-07-08
Document processing of local data using python transform image

daw3rd:
Makes it easy to process local data.

#317


roytman (commented):
Why do you add this mounted directories to pure python only, it is useful for Ray local as well.

roytman (approved):
LGTM

transforms/code/code2parquet/python/Dockerfile
```diff
@@ -10,6 +10,9 @@ RUN useradd -ms /bin/bash dpk
 USER dpk
 WORKDIR /home/dpk
 
+# Create directories to mount volumnes for processing data outside of the image.
+RUN mkdir input && mkdir output
```
blublinsky, line 14: I am not convinced we need this. According to https://www.warp.dev/terminus/docker-run-volume you can mount any local directory into a Docker container and then use it for execution. There is no need to do mkdir. Can you please verify this

transforms/code/code2parquet/python/Dockerfile
```diff
@@ -10,6 +10,9 @@ RUN useradd -ms /bin/bash dpk
...
+
```
blublinsky: I am not convinced we need this. According to https://www.warp.dev/terminus/docker-run-volume you can mount any local directory into a Docker container and then use it for execution. There is no need to do mkdir. Can you please verify this. Even more info is here https://docs.docker.com/storage/volumes/

blublinsky (changes requested):

transforms/code/code2parquet/python/Dockerfile
daw3rd: Ok yes, no need to create the mount points ahead of time.  Reverted the Dockerfiles and updated the transform readmes to point to the quickstart for running images.  Thanks.

blublinsky (approved):

PR #405, opened 2024-07-11
Enables configuration of makefile to use src or pypi for data-prep-kit library dependencies

daw3rd:
This is helpful if a transform implementation wants to NOT use the library source and instead 
1. use a published version of data-prep-toolkit from pypi.
2. not install lib source into docker images

Changes include
1. Adds USE_REPO_LIB_SRC=1 to default to original behavior.  Transform makefiles can override and set to 0 instead to get pypi dependencies and not build images with lib src.
3. Moved install of pytest, pytest-cov to .defaults.create-venv target.
4. Added some logging for target that installs from src dir.


shivdeep-singh-ibm (approved):
lgtm

PR #409, opened 2024-07-15
Suppot Multi-platform images.

revit13:

This PR add support for [Multi-platform images](https://docs.docker.com/build/building/multi-platform/)

It build the images for platforms linux/amd64 and linux/arm64


blublinsky:
its cute, but why do we really need it

PR #412, opened 2024-07-16
Remove creation of clusterrolebinding in kubeflow installation.

revit13:


roytman (commented):
@revit13 , why do you think that we don't need this setting?

blublinsky:
Second @roytman opinion

roytman (approved):
LGtM

PR #413, opened 2024-07-16
initial implementation of the simple python APIs for invoking transforms

blublinsky:

To simplify using transforms in user code, for example, notebooks or straight Python code


blublinsky:
This still requires a bit of work. Right now, it uninstalls the library but not its dependencies. was thinking about https://stackoverflow.com/questions/73600082/how-to-reference-a-requirements-txt-in-the-pyproject-toml-of-a-setuptools-projec to restore requirements.txt, so that we can get dependencies from there

blublinsky:
Also it will work for Ray the same way

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+    if not installer.uninstall(name=l_name):
+        logger.warning(f"failed uninstall transform {l_name}")
+    if res == 0:
+        return True
+    logger.warning(f"failed execution of transform {name}")
+
+
+execute_python_transform(
+    name="noop",
+    input_folder="/Users/borisl/IdeaProjects/data-prep-kit/transforms/universal/noop/python/test-data/input",
```
daw3rd: Oops!  your directory?

data-processing-lib/python/src/data_processing/utils/transform_configurator.py
```diff
@@ -0,0 +1,117 @@
...
+class TransformsConfiguration:
+    """
+    Configurations for existing transforms
+    """
+    def __init__(self):
+        """
+        Initialization - building transforms dictionary
+        """
+        self.transforms = {
+            "code2parquet": ("transforms/code/code2parquet/",       # Subdirectory in the project
```
daw3rd: The approach here is very delicate and only supports transforms that are part of the repo.  What if we restructure things to move them out of the repo in the future, then this becomes non-functional.  I suppose we can point to other repos instead of subdirs, but having an inventory like this does not seem correct.  Maybe an alternate way of configuring/populating this registry would be from a json file which is then more adaptable.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+            "output_folder": output_folder,
+        })}
+    p = location | params
+    # create configuration
+    klass = _import_class(t_class)
+    transform_configuration = klass()
+    # Set the command line args
+    sys.argv = ParamsUtils.dict_to_req(d=p)
+    # create launcher
+    launcher = PythonTransformLauncher(runtime_config=transform_configuration)
```
daw3rd: How does one extend/use/implement this for the other runtimes?

daw3rd (changes requested):

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+    components = name.split('.')
+    mod = __import__(components[0])
+    for comp in components[1:]:
+        mod = getattr(mod, comp)
+    return mod
+
+
+def execute_python_transform(name: str, input_folder: str, output_folder: str, params: dict[str, Any]) -> bool:
+    # get transform configuration
+    subdirectory, l_name, t_class = (TransformsConfiguration().
```
daw3rd: What names are allowed to be passed here?  Maybe a list() function is needed somewhere that provides the list of acceptable names.

daw3rd (changes requested):

daw3rd:
This solves the complexity of getting the right pip install commands to use a given transform.  In addition to `execute_transform` maybe we need an `install_transform` so that they can then import and use the transform as they would in any python file.  I think 2 issues are being addressed here. 1) pip install 2) execute.  I like 1, but not sure we have the right model for 2 here.  Maybe something like
```
install_transform("xyz")
install_transform("abc")
import xyz, abc
t1 = xyz()
t2 = abc()
pipelined = PipelinedTransform([t1, t2])
execute_transform(pipeline, input, output)
```

blublinsky:
The difference between the 2 is intermediate data on disk. I think it is important

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: this is just a simple example

data-processing-lib/python/src/data_processing/utils/transform_configurator.py
blublinsky: Sure, easy fix

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: You can use any launcher here. I wanted to do Python first, but will do Ray tomorrow. That is why there is Python in the method name

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: Sure thing, its easy

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+        return False
+
+    installer = PipInstaller()
+
+    # Install transform
+    if not installer.install(project=project, subdirectory=subdirectory, name=l_name):
+        logger.warning(f"failed to install transform {name}")
+        return False
+    if not installer.validate(name=l_name):
+        logger.warning(f"failed to install transform {l_name}")
```
roytman: you should separate the warning message from one on line 42, so I suggest printing "failed to validate transform "

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+        "data_local_config": ParamsUtils.convert_to_ast({
+            "input_folder": input_folder,
+            "output_folder": output_folder,
+        })}
+    p = location | params
+    # create configuration
+    klass = _import_class(t_class)
+    transform_configuration = klass()
+    # Set the command line args
+    sys.argv = ParamsUtils.dict_to_req(d=p)
```
roytman: we overwrite the arguments list; it will be required later, I suggest storing the original `sys.argv` and setting them back at the end of the function.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
roytman: so you decided to have a launcher per runtime, and do not use a single one for all (most of) runtimes?

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,75 @@
...
+    # get transform configuration
+    subdirectory, l_name, t_class = (TransformsConfiguration().
+                                     get_configuration(transform=name, runtime=TransformRuntime.PYTHON))
+    if subdirectory is None:
+        return False
+
+    installer = PipInstaller()
+
+    # Install transform
+    if not installer.install(project=project, subdirectory=subdirectory, name=l_name):
```
roytman: you automatically install and more important uninstall the transformer. May be it was installed, and required for something else, so we don't need to uninstall it. 
It is better to check if it is installed; if it is, do not install and **do not uninstall** the transformer.

roytman: as we discussed, it will not work. You install packages in a subprocess, so the results will be relevant to sub-processes only, but not for **this** process.

transforms/code/malware/python/src/malware_local_python.py
```diff
@@ -50,6 +51,6 @@
     # Set the simulated command line args
     sys.argv = ParamsUtils.dict_to_req(d=params | malware_params)
     # create launcher
-    launcher = PythonTransformLauncher(runtime_config=MalwareTransformConfiguration())
+    launcher = PythonTransformLauncher(runtime_config=MalwarePythonTransformConfiguration())
     # Launch the ray actor(s) to process the input
```
roytman: I know it is not part of the PR, but ray actor(s) in the pure python implementation....

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: Yes, they use different launchers and different configurations

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: right, its removed

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: done

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: done

blublinsky: I tested it, it works fine

transforms/code/malware/python/src/malware_local_python.py
blublinsky: It just fixing and obvious bug

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
roytman: you can use the same launcher, that will check if the configuration is inherited from Python, Ray or Spark configuration

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
roytman: Can you check it again?
Please validate that att the starting point you cannot load/import a class, 
install the relevant package and try to import/load the class again. 
It is possible that the package was installed, or was part of the PYTHONPATH even before its installation.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: fixed

data-processing-lib/python/src/data_processing/utils/transform_configurator.py
blublinsky: done

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: done

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
daw3rd: this should be implemented as a class and then the launch piece can be stubbed out.  Or pass in a method to do the launch, but a cut/paste with a one line change will not work.

daw3rd (changes requested):

blublinsky:
I have no idea what you mean with this comment:

```
this should be implemented as a class and then the launch piece can be stubbed out. Or pass in a method to do the launch, but a cut/paste with a one line change will not work.
```
You have to be more specific

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,127 @@
...
+
+project = "https://github.com/IBM/data-prep-kit.git"
+logger = get_logger(__name__)
+
+
+def _import_class(name):
+    """
+    Import class by name
+    :param name: name
+    :return:
```
roytman, line 35: can you add something about return value

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
```diff
@@ -0,0 +1,127 @@
...
+    :param t_class: transform configuration class
+    :param launcher: transform launcher
+    :param input_folder: input folder (local or S3)
+    :param output_folder: output folder (local or S3)
+    :param s3_config: S3 configuration - None local data
+    :param params: transform parameters
+    :return:
+    """
+    # Create data access params
+    if s3_config is None:
```
roytman, line 65: I think, it makes sense to add at least a warning when both S3 and local data are provided. Otherwise, somebody will have hard time to debug, why he provided local parameters, but the system uses S3.

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
```diff
@@ -0,0 +1,73 @@
...
+        :param project: git project
+        :param subdirectory: sub directory
+        :param name: name
+        :return: True if installation is successful, False in the case of failures
+        """
+        # build pip package string
+        package = f"git+{project}#subdirectory={subdirectory}&egg={name}"
+        try:
+            # Do pip install (in sub process)
+            subprocess.check_call([sys.executable, "-m", "pip", "install", package])
```
roytman, line 42: the resent recommendation from subprocess team is to use `subprocess.run`

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+import warnings
+warnings.filterwarnings('ignore')
+
+import subprocess
+import sys
+import pkg_resources
```
roytman, line 21: the module `pkg_resources` is deprecated and will be removed, they recommend to use `[importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources)` instead

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
```diff
@@ -0,0 +1,73 @@
...
+    @staticmethod
+    def uninstall(name: str):
+        """
+        Uninstall library
+        :param name: name
+        :return: True if the un installation succeeds, False otherwise
+        """
+        try:
+            # Invoke pip (in subprocess) to uninstall
+            subprocess.check_call([sys.executable, "-m", "pip", "uninstall", "-y", name])
```
roytman, line 71: again, subprocess.run

data-processing-lib/python/src/data_processing/utils/transform_configuration.json
```diff
@@ -0,0 +1,247 @@
+{
```
roytman, line 1: should we add automatic update, or verification of this file? E.g. if a new sub-directory is created in the transforms dir.

data-processing-lib/python/test/data_processing_tests/invoker/python_invoker_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+def test_configuration():
+    """
+    test configuration population
+    :return:
+    """
+    t_configuration = TransformsConfiguration()
+    transforms = t_configuration.get_available_transforms()
+    logger.info(f"available transforms {transforms}")
+    assert len(transforms) == 13
```
roytman, line 30: this magic number, can create issues when we add new transformers. Should we compare number of transformers subdirectories in the `transforms` tree.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_invoker.py
```diff
@@ -0,0 +1,78 @@
...
+            return False
+        p_installed = True
+
+    # Check if transformer already installed
+    r_installed = False
+    if not installer.validate(name=r_l_name):
+        # transformer is not installed, install it
+        if not installer.install(project=project, subdirectory=r_subdirectory, name=r_l_name):
+            logger.warning(f"failed to install transform {name}")
+            return False
```
roytman, line 75: we don't cleanup the python installation, but maybe it's OK

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: I have both tests and a notebook. It all works

transforms/code/malware/python/src/malware_local_python.py
blublinsky: not my issue

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: There is none. It just imports the class

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
blublinsky: Input_folder, output folder are defined in both cases. They have different formats, of course, but validation is a bit painful

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
blublinsky: Yes, but ... check_call verifies that the invocation was successful. it uses run underneath

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
blublinsky: I tried it, it does not seem to work

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
blublinsky: see above

data-processing-lib/python/src/data_processing/utils/transform_configuration.json
blublinsky: why?

data-processing-lib/python/test/data_processing_tests/invoker/python_invoker_test.py
blublinsky: Yes. Adding transform should change it

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_invoker.py
blublinsky: fixed

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
roytman: you can do something like `subprocess.check_call([sys.executable, "-m", "pip", "show", "-q", name])`
and check return code

data-processing-lib/python/src/data_processing/utils/pipinstaller.py
roytman: indeed, run has a special flag, to verify that the invocation was successful.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_invoker.py
roytman: the function has the return statement. 
`return mod`

roytman (approved):
LGTM

shivdeep-singh-ibm (approved):

daw3rd (changes requested):
This needs to wait on the outcome of out-of-band discussions on the need for this.

roytman:
@blublinsky , please remove your execution outputs from the Jupyter notebook.
line like: 
```
/Users/borisl/Projects/data-prep-kit/examples/notebooks/code/input_data
/Users/borisl/Projects/data-prep-kit/examples/notebooks/code/test-data/parquet_input
/Users/borisl/Projects/data-prep-kit/transforms/code/code2parquet/python/test-data/languages/lang_extensions.json
```
should not be the in the git.

daw3rd (approved):

PR #415, opened 2024-07-16
Added documentation on how to use data-access-factory

santoshborse:
Document DAF example


data-processing-lib/doc/overview.md
```diff
@@ -45,6 +45,7 @@ To learn more consider the following:
 * [Transform Exceptions](transform-exceptions.md)
 * [Transform Runtimes](transform-runtimes.md)
 * [Transform Examples](transform-tutorial-examples.md)
+* [Data Access Factory S3 Example](using-data-access-factory.md)
```
daw3rd: I think a more general discussion of the DataAccessFactory concept and what it facilitates would be a good starting point - a big picture sort of overview, including input selection, checkpointing, reading and writing, etc.
Then a discussion on an S3 and/or local example.

daw3rd (changes requested):

blublinsky:
thank you  @santoshborse. I would suggest starting from the importance of data access abstraction and the advantages it provides for the overall implementation. 
The next thing is to describe high level APIs that the abstraction provides and then talk about the factory

daw3rd (approved):

PR #416, opened 2024-07-17
Add pdf2parquet transform

dolfim-ibm:

This PR adds a transform for converting PDF files to Markdown


.make.defaults
```diff
@@ -214,6 +214,7 @@ __check_defined = \
 		touch pyproject.toml;	\
 	fi
 	$(DOCKER) build -t $(DOCKER_LOCAL_IMAGE)		\
+		--platform linux/amd64  \
```
dolfim-ibm: @daw3rd @blublinsky would this be ok or should it be put as some more fancy Makefile variable?

.make.defaults
```diff
@@ -37,6 +37,7 @@ ABS_REPOROOT=$(shell (cd $(REPOROOT); pwd))
 DOCKER_FILE?=Dockerfile
 #DOCKER_IMAGE_NAME?=xyzzy	# Must be defined by the includeing makefile
 DOCKER?=docker
+DOCKER_PLATFORM=?linux/amd64
```
daw3rd: Sorry, I misled you, ? before =.

transforms/universal/pdf2md/README.md
```diff
@@ -0,0 +1,14 @@
+# NOOP Transform 
+The NOOP transforms serves as a simple exemplar to demonstrate the development
```
daw3rd: Ooops :).  Please add appropriate doc here.

transforms/universal/pdf2md/README.md
```diff
@@ -0,0 +1,14 @@
...
+of a simple 1:1 transform.  Per the set of 
+[transform project conventions](../../README.md#transform-project-conventions)
+the following runtimes are available:
+
+* [python](python/README.md) - provides the base python-based transformation 
+implementation.
+* [ray](ray/README.md) - enables the running of the base python transformation
+in a Ray runtime
+* [spark](spark/README.md) - enables the running of a spark-based transformation
```
daw3rd: No spark yet.

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
daw3rd, line 1: noop?

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
```diff
@@ -0,0 +1,198 @@
...
+################################################################################
+import os
+
+import kfp.compiler as compiler
+import kfp.components as comp
+import kfp.dsl as dsl
+from workflow_support.compile_utils import ONE_HOUR_SEC, ONE_WEEK_SEC, ComponentUtils
+
+
+task_image = "quay.io/dataprep1/data-prep-kit/noop-ray:latest"
```
daw3rd: noop?

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
```diff
@@ -0,0 +1,198 @@
...
+
+# the name of the job script
+EXEC_SCRIPT_NAME: str = "noop_transform_ray.py"
```
daw3rd: noop?

transforms/universal/pdf2md/kfp_ray/noop_wf.py
```diff
@@ -0,0 +1,199 @@
...
+import kfp.compiler as compiler
+import kfp.components as comp
+import kfp.dsl as dsl
+from workflow_support.compile_utils import ONE_HOUR_SEC, ONE_WEEK_SEC, ComponentUtils
+
+
+task_image = "quay.io/dataprep1/data-prep-kit/noop-ray:latest"
+
+# the name of the job script
+EXEC_SCRIPT_NAME: str = "noop_transform_ray.py"
```
daw3rd: noop?

transforms/universal/pdf2md/kfp_ray/noop_wf.py
```diff
@@ -0,0 +1,199 @@
...
+################################################################################
+import os
+
+import kfp.compiler as compiler
+import kfp.components as comp
+import kfp.dsl as dsl
+from workflow_support.compile_utils import ONE_HOUR_SEC, ONE_WEEK_SEC, ComponentUtils
+
+
+task_image = "quay.io/dataprep1/data-prep-kit/noop-ray:latest"
```
daw3rd: noop?

transforms/universal/pdf2md/python/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM docker.io/python:3.11.9-slim-bookworm
```
daw3rd: Why the change from FROM docker.io/python:3.10.14-slim-bullseye

transforms/universal/pdf2md/python/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+COPY --chown=dpk:root src/ src/
+COPY --chown=dpk:root pyproject.toml pyproject.toml 
+RUN pip install --no-cache-dir -e .
+
+# Download models
+RUN python -c 'from deepsearch_glm.utils.load_pretrained_models import load_pretrained_nlp_models; load_pretrained_nlp_models(verbose=True);'
+RUN python -c 'from docling.document_converter import DocumentConverter; from pathlib import Path; DocumentConverter.download_models_hf(local_dir=Path("./artifacts/"));'
+
+# copy source data
+COPY ./src/pdf2md_transform.py .
```
daw3rd: This should not be needed as line 28 above will have installed it.

transforms/universal/pdf2md/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Output format
+
+The output format will contain all the columns of the metadata CSV file,
+with the addition of the following columns
+
+```jsonc
+{
+    // all the fields of the input CSV + ...
```
daw3rd: CSV -> parquet?

transforms/universal/pdf2md/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+    "pdf_convert_time": "float",  // time taken to convert the document in seconds
+}
+```
+
+
+## Parameters
+
+| Parameter  | Default  | Description  |
+|------------|----------|--------------|
+| `--pdf2md_modelsdir`                  | `./artifacts` | The location where the models are located or downloaded to |
```
daw3rd: We want to support configuraiton of the transform independent of the CLI.  Can you discuss/doc the configuration keys that the transform's initializer() supports.  In general, these keys should not include the CLI prefix (pdf2md_) since the prefix is used distinguish the args from other cli args, but not needed when configuring the transform directly.

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+from docling.datamodel.document import ConvertedDocument, DocumentConversionInput
+from docling.document_converter import DocumentConverter
+from docling.pipeline.standard_model_pipeline import PipelineOptions
+
+
+logger = get_logger(__name__)
+
+shortname = "pdf2md"
+cli_prefix = f"{shortname}_"
+pdf2md_modelsdir_key = f"{shortname}_modelsdir"
```
daw3rd: per earlier comment, the keys used in the initializer should not include the cli_prefix.

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+            }
+            return [
+                (TransformUtils.convert_arrow_to_binary(table=table), ".parquet")
+            ], metadata
+        except Exception as e:
+            print(f"Fatal error with file {file_name=}. No results produced.")
+            logger.error(f"Fatal error with file {file_name=}. No results produced.")
+            return [], {}
+
+# class PdfToMdRuntime(DefaultRayTransformRuntime):
```
daw3rd: Pdf2Md?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+#         # converter = DocumentConverter(artifacts_path=self.models_dir)
+#         # converter_refs = ray.put(converter)
+
+#         # logger.debug(f"Placed converter into Ray object storage under reference {converter_refs=}")
+#         # return {ingest_converter_refname: converter_refs} | self.params
+
+#         return self.params
+
+
+class PdfToMdTransformConfiguration(TransformConfiguration):
```
daw3rd: Pdf2Md?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        Add Transform-specific arguments to the given parser.
+        By convention a common prefix should be used for all mutator-specific CLI args
+        (e.g, noop_, pii_, etc.)
+        """
+        parser.add_argument(
+            f"--{pdf2md_modelsdir_key}",
+            type=str,
+            help="Path where to download models",
+            default="scratch",
```
daw3rd: All of these defaults need to also be implemented in the transform initializer.  Probably should define constants that hold the defaults and then use those instead of the literals in both places.

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+        )
+
+    def apply_input_params(self, args: Namespace) -> bool:
+        """
+        Validate and apply the arguments that have been parsed
+        :param args: user defined arguments.
+        :return: True, if validate pass or False otherwise
+        """
+        dargs = vars(args)
+        self.params = {
```
daw3rd: For the de-prefixing, see what noop does https://github.com/IBM/data-prep-kit/blob/736055c99c85e77e76cf5939c7cd124320a6c44e/transforms/universal/noop/python/src/noop_transform.py#L114

transforms/universal/pdf2md/python/src/pdf2md_transform_python.py
```diff
@@ -0,0 +1,42 @@
...
+    PythonTransformRuntimeConfiguration,
+)
+from data_processing.utils import get_logger
+from pdf2md_transform import PdfToMdTransformConfiguration
+
+
+logger = get_logger(__name__)
+
+
+class PdftoMdPythonTransformConfiguration(PythonTransformRuntimeConfiguration):
```
daw3rd: Pdf2Md?

transforms/universal/pdf2md/python/test/test_pdf2md.py
```diff
@@ -0,0 +1,40 @@
...
+import os
+import pyarrow as pa
+from data_processing.test_support import get_files_in_folder
+from data_processing.test_support.transform import AbstractBinaryTransformTest
+from data_processing.utils import TransformUtils
+from pdf2md_transform import PdfToMdTransform
+
+
+
+class TestPdfToMdTransform(AbstractBinaryTransformTest):
```
daw3rd: Pdf2Md?

transforms/universal/pdf2md/ray/README.md
```diff
@@ -0,0 +1,49 @@
+# NOOP Ray Transform
```
daw3rd: NOOP

transforms/universal/pdf2md/ray/README.md
```diff
@@ -0,0 +1,49 @@
+# NOOP Ray Transform 
+Please see the set of
+[transform project conventions](../../../README.md#transform-project-conventions)
+for details on general project conventions, transform configuration,
+testing and IDE set up.
+
+## Summary 
+This project wraps the [noop transform](../python) with a Ray runtime.
```
daw3rd: NOOP and elsewhere

daw3rd (changes requested):

.make.defaults
daw3rd: This is causing the image-based ci/cd to fail. Sorry for that.

transforms/universal/pdf2md/kfp_ray/README.md
```diff
@@ -0,0 +1,28 @@
+# NOOP Ray-base KubeFlow Pipeline Transformation 
+
+
+## Summary 
+This project allows execution of the [noop Ray transform](../ray) as a 
+[KubeFlow Pipeline](https://www.kubeflow.org/docs/components/pipelines/overview/)
```
blublinsky: This is noop

transforms/universal/pdf2md/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+```jsonc
+{
+    // all the fields of the input CSV + ...
+
+    "source_filename": "string",  // the basename of the source archive or file
+    "filename": "string",         // the basename of the PDF file
+    "contents": "string",         // the content of the PDF
+    "document_id": "string",      // the document id, corresponding to the hash of the content 
+    "ext": "string",              // the detected file extension
+    "hash": "string",             // the hash of the `contents` column
```
blublinsky: so doc_id == hash. Why have both?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
blublinsky: why?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+
+        data = []
+        success_doc_id = []
+        failed_doc_id = []
+        skipped_doc_id = []
+        number_of_rows = 0
+
+        try:
+            root_kind = filetype.guess(byte_array)
+
```
blublinsky: you explicitly have file name. Can you just use it ext?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+                    data.append(file_data)
+                    number_of_rows += 1
+
+                except Exception as e:
+                    failed_doc_id.append(archive_doc_filename)
+                    logger.warning(
+                        f"Exception {str(e)} processing file {archive_doc_filename}, skipping"
+                    )
+
```
blublinsky: So in this case we have a single row table?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+                                ext = kind.extension
+
+                                file_data = self._convert_pdf2md(doc_filename=archive_doc_filename, ext=ext, content_bytes=content_bytes)
+                                file_data["source_filename"] = TransformUtils.get_file_basename(file_name)
+
+                                data.append(file_data)
+                                number_of_rows += 1
+
+                            except Exception as e:
+                                failed_doc_id.append(archive_doc_filename)
```
blublinsky: You report just a counter, so please, use the counter, not array

transforms/universal/pdf2md/python/src/pdf2md_transform.py
blublinsky: please remove this class. Its not applicable to Python runtime

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+            do_ocr=self.do_ocr,
+        )
+        self._converter = DocumentConverter(artifacts_path=self.models_dir, pipeline_options=pipeline_options)
+
+        # self.doc_counter = Counter("worker_pdf_doc_count", "Number of PDF documents converted by the worker")
+        # self.page_counter = Counter("worker_pdf_pages_count", "Number of PDF pages converted by the worker")
+        # self.page_convert_gauge = Gauge("worker_pdf_page_avg_convert_time", "Average time for converting a single PDF page on each worker")
+        # self.doc_convert_gauge = Gauge("worker_pdf_convert_time", "Time spent converting a single document")
+
+    def _update_metrics(self, num_pages: int, elapse_time: float):
```
blublinsky: why is this here?

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,346 @@
...
+        # Convert PDF to Markdown
+        start_time = time.time()
+        buf = io.BytesIO(content_bytes)
+        input_docs = DocumentStream(filename=doc_filename, stream=buf)
+        input = DocumentConversionInput.from_streams([input_docs])
+
+        converted_docs = self._converter.convert(input)
+        doc: ConvertedDocument = next(converted_docs, None)
+        if doc is None or doc.output is None:
+            raise RuntimeError("Failed in converting.")
```
blublinsky: Can we get out of memory here and what is going to happen in this case?

transforms/universal/pdf2md/python/README.md
blublinsky: These are input parameters, so they are fine

transforms/universal/pdf2md/python/Dockerfile
dolfim-ibm: we have python 3.11 as minimum version required

transforms/universal/pdf2md/python/Dockerfile
dolfim-ibm: oh correct. then we might have to update the noop example as well.

transforms/universal/pdf2md/python/README.md
dolfim-ibm: this was a leftover. removed

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: file extensions are notoriously known to be bad indicator of the file type. we have seen many ".pdf" that instead and html error pages or ".php" files that are instead PDF, etc

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: when this exception triggers, I think the output will be an empty table

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: We can easily rename all. At the moment we used the same scheme as in https://github.com/IBM/data-prep-kit/blob/dev/transforms/code/code2parquet/python/src/code2parquet_transform.py.
@daw3rd @blublinsky just let us know which naming convention is advised.

transforms/universal/pdf2md/python/README.md
daw3rd: @blublinsky no, we separately have config parameters to the transform and CLI parameters.  this is well-established pattern in other transforms.  My comment stands.

transforms/universal/pdf2md/python/src/pdf2md_transform.py
daw3rd: I see no reason to use 2 in one place and `To` in another.  code2parquet should be fixed, but that his a separate issue.

transforms/universal/pdf2md/python/src/pdf2md_transform.py
```diff
@@ -0,0 +1,347 @@
...
+
+            table = pa.Table.from_pylist(data)
+            metadata = {
+                "nrows": len(table),
+                "nsuccess": len(success_doc_id),
+                "nfail": len(failed_doc_id),
+                "nskip": len(skipped_doc_id),
+            }
+            return [
+                (TransformUtils.convert_arrow_to_binary(table=table), ".parquet")
```
daw3rd: This transform produces .parquet, not .md.  As such it should be renamed to pdf2parquet, similar to code2parquet.

daw3rd (changes requested):
Separately, we are you using markdown and not raw text in the content column?

transforms/universal/pdf2md/python/README.md
dolfim-ibm: So should we remove `--pdf2md_` from the README? the usage and meaning is still the same.
from the latest commit, the transform doesn't require the prefix anymore for its config

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: this was a preparation for the ray version, now it is removed

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: has been changed

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: The transform initializer has the same values. indeed, we can improve by using some literals

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: thanks for the pointers, we didn't know about it. should be there

transforms/universal/pdf2md/python/README.md
dolfim-ibm: now we used `uuid4` as in the other examples

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: this is overloaded in the ray implementation

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: if the OOM is triggered via an exception, it will simply skip one document in the zip.
if the OOM is killing the process, the whole transform is killed, and we assume the ray orchestrator will take care, like in all the debugging we did together

transforms/universal/pdf2md/python/README.md
daw3rd: No, we should add the non-CLI config params that do not include pdf2md_  prefix to the README, then the CLI --pdf2md* params can just refer to that documentation.  The point here is that there is configuration of the transform that is independent of the CLI.  the prefix is used in the CLI to help better distinguish these within the very large set of params that are presented in the cloud/kfp gui to run at scale.

transforms/universal/pdf2md/python/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+COPY --chown=dpk:root pyproject.toml pyproject.toml 
+RUN pip install --no-cache-dir -e .
+
+# Download models
+RUN python -c 'from deepsearch_glm.utils.load_pretrained_models import load_pretrained_nlp_models; load_pretrained_nlp_models(verbose=True);'
+RUN python -c 'from docling.document_converter import DocumentConverter; s=DocumentConverter.download_models_hf(); print(f"Models cached in {s}")'
+
+# copy source data
+COPY --chown=dpk:root src/ .
```
daw3rd: The pdf2md_transform_python.py file needs to be in the home directory as that is, by convention, where the main()s live in the images.

transforms/universal/pdf2md/python/Makefile
```diff
@@ -0,0 +1,67 @@
+# Define the root of the local git clone for the common rules to be able 
+# know where they are running from.
+REPOROOT=../../../..
+# Include a library of common .transform.* targets which most
+# transforms should be able to reuse.  However, feel free
+# to override/redefine the rules below. 
+
+# $(REPOROOT)/.make.versions file contains the versions
+
+TRANSFORM_RUNTIME=python
```
daw3rd: TRANSFORM_RUNTIME should not be need now that you are in a `python` directory.

transforms/universal/pdf2md/python/.dockerignore
```diff
@@ -0,0 +1,6 @@
+venv/
```
daw3rd: Do you need to ignore HF_CACHE folder you reference in the help text for the transform?

transforms/universal/pdf2md/python/.gitignore
```diff
@@ -0,0 +1,39 @@
+test-data/output
```
daw3rd: Do you need to ignore HF_CACHE folder you reference in the help text for the transform?

daw3rd (changes requested):

transforms/universal/pdf2md/python/Dockerfile
dolfim-ibm: this is what this line does, or?

transforms/universal/pdf2md/python/.dockerignore
dolfim-ibm: the huggingface cache is usually in the home folder, not inside the repo, so I don't think it is needed

transforms/universal/pdf2md/python/.gitignore
dolfim-ibm: the huggingface cache is usually in the home folder, not inside the repo, so I don't think it is needed

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: literals are now in place

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
dolfim-ibm: done

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
dolfim-ibm: done

transforms/universal/pdf2md/kfp_ray/noop_multiple_wf.py
dolfim-ibm: done

transforms/universal/pdf2md/kfp_ray/noop_wf.py
dolfim-ibm: done

transforms/universal/pdf2md/kfp_ray/noop_wf.py
dolfim-ibm: done

transforms/universal/pdf2md/python/Dockerfile
dolfim-ibm: done

transforms/universal/pdf2md/python/README.md
dolfim-ibm: this is done

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: done

transforms/universal/pdf2md/python/src/pdf2md_transform.py
dolfim-ibm: done

transforms/universal/pdf2md/python/test/test_pdf2md.py
dolfim-ibm: done

transforms/universal/pdf2md/ray/README.md
dolfim-ibm: done

transforms/universal/pdf2md/ray/README.md
dolfim-ibm: done

transforms/universal/pdf2md/kfp_ray/README.md
dolfim-ibm: done

transforms/universal/pdf2md/python/Makefile
dolfim-ibm: done

transforms/universal/pdf2parquet/python/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+
+## Parameters
+
+The transform can be initialized with the following parameters.
+
+| Parameter  | Default  | Description  |
+|------------|----------|--------------|
+| `artfacts_path`              | <unset> | Path where to models artifacts are located, if unset they will be downloaded in the HF_CACHE folder. |
```
daw3rd: I see no documentation on what this actually is.  Also, not sure, but i think you mean `artifacts_path` here.

transforms/universal/pdf2parquet/python/src/pdf2parquet_transform.py
```diff
@@ -0,0 +1,269 @@
...
+                "nrows": len(table),
+                "nsuccess": len(success_doc_id),
+                "nfail": len(failed_doc_id),
+                "nskip": len(skipped_doc_id),
+            }
+            return [
+                (TransformUtils.convert_arrow_to_binary(table=table), ".parquet")
+            ], metadata
+        except Exception as e:
+            logger.error(f"Fatal error with file {file_name=}. No results produced.")
```
daw3rd: You should throw an exception instead of eating it.  The runtime will log it and may keep a count.  The result will be the same, that is, that there is no file written for this call to transform().

transforms/universal/pdf2parquet/python/src/pdf2parquet_transform.py
```diff
@@ -0,0 +1,269 @@
...
+            self.artifacts_path = Path(self.artifacts_path)
+        self.do_table_structure = config.get(pdf2parquet_do_table_structure_key, pdf2parquet_do_table_structure_default)
+        self.do_ocr = config.get(pdf2parquet_do_ocr_key, pdf2parquet_do_ocr_default)
+
+        logger.info("Initializing models")
+        pipeline_options = PipelineOptions(
+            do_table_structure=self.do_table_structure,
+            do_ocr=self.do_ocr,
+        )
+        self._converter = DocumentConverter(artifacts_path=self.artifacts_path, pipeline_options=pipeline_options)
```
daw3rd: Is the artifact_path a path in the local file system?  If so, you should also support S3 using a DataAccessFactory.

transforms/universal/pdf2parquet/python/test/test_pdf2parquet.py
```diff
@@ -0,0 +1,55 @@
...
+    def get_test_transform_fixtures(self) -> list[tuple]:
+        dal = DataAccessLocal()
+        basedir = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), "../test-data")
+        )
+        input_dir = os.path.join(basedir, "input")
+        input_files = get_files_in_folder(input_dir, ".pdf")
+        input_files = [(name, binary) for name, binary in input_files.items()]
+        expected_metadata_list = [{"nrows": 1, "nsuccess": 1, "nfail": 0, "nskip": 0}, {}]
+        config = {}
```
daw3rd: can you add a test that uses the do_ocr and do_table_conversions?

transforms/universal/pdf2parquet/python/src/pdf2parquet_transform.py
```diff
@@ -0,0 +1,269 @@
...
+        input_docs = DocumentStream(filename=doc_filename, stream=buf)
+        input = DocumentConversionInput.from_streams([input_docs])
+
+        converted_docs = self._converter.convert(input)
+        doc: ConvertedDocument = next(converted_docs, None)
+        if doc is None or doc.output is None:
+            raise RuntimeError("Failed in converting.")
+        elapse_time = time.time() - start_time
+
+        content_string = doc.render_as_markdown()
```
daw3rd: the `content` column needs to contain raw text.

transforms/universal/pdf2parquet/python/test/test_pdf2parquet_python.py
```diff
@@ -0,0 +1,48 @@
...
+        fixtures = []
+        launcher = PythonTransformLauncher(Pdf2ParquetPythonTransformConfiguration())
+        fixtures.append(
+            (
+                launcher,
+                config,
+                basedir + "/input",
+                basedir + "/expected",
+                # this is added as a fixture to remove these columns from comparison
+                ["date_acquired", "document_id", "pdf_convert_time", "contents"],
```
daw3rd: If you ignore `contents` you're not really doing a test of the generated vs expected text.

transforms/universal/pdf2parquet/ray/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+COPY --chown=ray:users pyproject.toml pyproject.toml 
+RUN pip install --no-cache-dir -e .
+
+# Download models
+RUN python -c 'from deepsearch_glm.utils.load_pretrained_models import load_pretrained_nlp_models; load_pretrained_nlp_models(verbose=True);'
+# RUN python -c 'from docling.document_converter import DocumentConverter; from pathlib import Path; DocumentConverter.download_models_hf(local_dir=Path("./artifacts/"));'
+RUN python -c 'from docling.document_converter import DocumentConverter; s=DocumentConverter.download_models_hf(); print(f"Models cached in {s}")'
+
+# copy source data
+COPY --chown=ray:users src/ .
```
daw3rd: You need to copy in pdf2parquet_transform_ray.py to be in the home dir.

transforms/universal/pdf2parquet/python/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+COPY --chown=dpk:root pyproject.toml pyproject.toml 
+RUN pip install --no-cache-dir -e .
+
+# Download models
+RUN python -c 'from deepsearch_glm.utils.load_pretrained_models import load_pretrained_nlp_models; load_pretrained_nlp_models(verbose=True);'
+RUN python -c 'from docling.document_converter import DocumentConverter; s=DocumentConverter.download_models_hf(); print(f"Models cached in {s}")'
+
+# copy source data
+COPY --chown=dpk:root src/ .
+COPY ./src/pdf2parquet_local.py local/
```
daw3rd: You need to copy in pdf2parquet_transform_python.py to be in the home dir.

transforms/universal/pdf2parquet/ray/src/pdf2parquet_transform_ray.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        super().__init__(config)
+
+        self.doc_counter = Counter("worker_pdf_doc_count", "Number of PDF documents converted by the worker")
+        self.page_counter = Counter("worker_pdf_pages_count", "Number of PDF pages converted by the worker")
+        self.page_convert_gauge = Gauge("worker_pdf_page_avg_convert_time", "Average time for converting a single PDF page on each worker")
+        self.doc_convert_gauge = Gauge("worker_pdf_convert_time", "Time spent converting a single document")
+    
+    def _update_metrics(self, num_pages: int, elapse_time: float):
+        self.page_convert_gauge.set(elapse_time / num_pages)
```
daw3rd: @blublinsky can confirm, but if you publish these in the metadata {} returned from the python transform(), the ray runtime will aggregate/publish them for you.  As such, you would not need to have a Ray-specific sub-class of the transform.

daw3rd (changes requested):

transforms/universal/pdf2parquet/python/test/test_pdf2parquet_python.py
dolfim-ibm: it is checking the hash of the content.

transforms/universal/pdf2parquet/ray/Dockerfile
dolfim-ibm: this is what this command is doing, or am I missing something?

transforms/universal/pdf2parquet/python/Dockerfile
dolfim-ibm: This is coming from the other examples e.g. https://github.com/IBM/data-prep-kit/blob/dev/transforms/universal/noop/python/Dockerfile (lines 20 and 28-30). All the files in `/src` are in the home directory (otherwise the tests won't even run).
If this extra `local/` is not needed we simply remove it.

transforms/universal/pdf2parquet/python/README.md
dolfim-ibm: good catch. now fixed

transforms/universal/pdf2parquet/python/src/pdf2parquet_transform.py
dolfim-ibm: fixed

transforms/universal/pdf2parquet/python/test/test_pdf2parquet_python.py
```diff
@@ -0,0 +1,77 @@
...
+        fixtures = []
+        launcher = PythonTransformLauncher(Pdf2ParquetPythonTransformConfiguration())
+        fixtures.append(
+            (
+                launcher,
+                config,
+                basedir + "/input",
+                basedir + "/expected",
+                # this is added as a fixture to remove these columns from comparison
+                ["date_acquired", "document_id", "pdf_convert_time", "contents", "num_doc_elements"],
```
daw3rd: why are you removing contents and num_doc_elements from comparison.  Would think this is the whole point of the test.

transforms/universal/pdf2parquet/python/test/test_pdf2parquet_python.py
daw3rd: but why not the content

daw3rd (changes requested):

transforms/universal/pdf2parquet/python/Dockerfile
daw3rd: Yes, but then it does
`COPY ./src/noop_transform_python.py .`
This should be failing `make test-image` without this.  hmm

transforms/universal/pdf2parquet/ray/Dockerfile
daw3rd: we need `$(HOME)/pdf2parquet_transform_ray.py` , not `$(HOME)/src/pdf2parquet_transform_ray.py `

daw3rd (changes requested):

transforms/universal/pdf2parquet/ray/Dockerfile
daw3rd: see https://github.com/dolfim-ibm/data-prep-kit/blob/644a8766d896e2dec802b6bb43a7ab5eab4c5eab/transforms/universal/pdf2parquet/kfp_ray/pdf2parquet_wf.py#L23, for example.

transforms/universal/pdf2md/python/README.md
blublinsky: so, at least change the comment to define which is which

transforms/universal/pdf2md/python/src/pdf2md_transform.py
blublinsky: I get it, but it does not belong here

transforms/universal/pdf2md/python/src/pdf2md_transform.py
blublinsky: I think it will kill the process

transforms/universal/pdf2md/python/README.md
dolfim-ibm: yes, this was done in commit https://github.com/IBM/data-prep-kit/commit/1a8f0dbc13256c3fa76ec465fb16270aa0930c8a (5 days ago)

transforms/universal/pdf2parquet/ray/src/pdf2parquet_transform_ray.py
daw3rd: Per discussion with @blublinsky I'm ok with this.

daw3rd (approved):
LGTM.  Thanks!

Issue #374, opened 2024-07-03
[Feature] Enable composable/pipelined transforms in the python API


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core

### Feature

It would be useful to be able define a sequence of transforms that are run one after the other, within a python process.  This should be implemented as an AbstractBinaryTransform, probably that takes a list of transform instances.  The configuration for this class could either be a single dictionary or a list of dictionaries corresponding 1:1 with the list of transforms.  This "pipeline transform" should also be runnable in a runtime like any other transform.  This latter may mean a single config dict is used to initialize the "pipeline transform".  

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

blublinsky:
I am not convinced that this is the right thing to do. It will only work if all transforms in the sequence have comparable dependencies, which is not a save assumption. It might easily lead to library conflicts, which are hard to debug. We intentionally are running different transforms in different Python processes to avoid this issue

daw3rd:
Start of formal design discussion, given recent interest in this requirement.  The basic requirement is as follows then:

1. Ability to have a series of transforms called in sequence on an in-memory object, independent of run-time.

The base AbstractBinaryTransform (and AbstractTableTRansform) already work on in-memory objects, so I believe we can use these frameworks to implement a transform that calls a list of transforms, passing outputs to inputs as in-memory objects.  This same transform could then also be run as any other transform in any of our run-times to operate on on-disc data to create new output files.

Some complications/considerations

1. Creation of the transform in its most simple programmatic form.  For example, for code2parquet->noop :
```python
c2p = Code2ParquetTransform({ "domain" : "foo"})
noop = NoopTransform({"sleep": 1 })
p = Pipeline([c2p, noop])
```
However, to support CLI configuration for the run-times, a dictionary should be supported.  
Maybe  a nested/structured JSON/AST needs to be used to specify the configuration of the transforms.  
```json
{ 
    "code2parquet" :     { "domain": "foo" }, 
    "noop" : { "sleep": 1 }
}
```
Then the pipeline cli would be --pipeline_transforms "{ "code2parquet":  ...}".  How would we map from transform string references (i.e. "code2parquet) to a python class.  For example, in the above, using "code2parquet" would require a registry of mappings of strings to transform classes.  Are there alternatives?  Maybe the full python class name?
[...]

PR #424, opened 2024-07-18
initial pass at a pipelining transform

daw3rd:
We would like to define a sequence of transforms using python and run the sequence as any other transform would be run.

#374


data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+                returned by the transform/flush_binary() methods.
+        """
+        super().__init__(config)
+        self.input_extension = None
+        self.transforms = config.get(transform_key, None)
+        if self.transforms is None:
+            raise ValueError(f"Missing configuration key {transform_key} specifying the list of transforms to run")
+        for transform in self.transforms:
+            if not isinstance(transform, AbstractBinaryTransform):
+                raise ValueError(f"{transform} is not an instance of AbstractBinaryTransform")
```
blublinsky: Every transform here can have its own config parameters. Where are transforms initialized?

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+            to_process = pending_to_process
+            pending_to_process = []
+            for tp in to_process:  # Over all outputs from the last transform  (or the initial input)
+                fname = tp[0]
+                byte_array = tp[1]
+                transformation_tuples, metadata = transform.transform_binary(fname, byte_array)
+                # Capture the list of outputs from this transform as inputs to the next (or as the return values).
+                for transformation in transformation_tuples:
+                    transformed, extension = transformation
+                    fname = transform_name + "-output" + extension
```
blublinsky: This will break transform logic, if the need the input file name, for example codetoparquet or pdf conversion

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    def flush_binary(self) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
+        """
+        Call flush on all transforms in the pipeline and aggregated the results.
+        Returns:
+
+        """
+        r_bytes = []
+        r_metadata = {}
+        for transform in self.transforms:
+            transformation_tuples, metadata = transform.flush_binary()
```
blublinsky: This is completely wrong. flush from the first transform has to be processed by the rest of them, then fluch from the second transform has to be processed by remaining one and so on

data-processing-lib/python/test/data_processing_tests/transform/test_binary_pipeline.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def get_test_transform_fixtures(self) -> list[Tuple]:
+        # Defines correct evaluation of pipeline for the expected number of tables produced.
+        # It does NOT test the transformation of the transform contained in the pipeline other
+        # than to make sure the byte arrays are not changed due to using a NoopTransform in the pipeline.
+        # .parquet is used as the extension because the transforms being used are AbstractTableTransforms
+        # which use/expect parquet files.
+        fixtures = []
+        noop0 = NOOPTransform({"sleep": 0})
+        noop1 = NOOPTransform({"sleep": 0})
```
blublinsky, line 68: This is complete cheating

blublinsky (commented):
This is really trvialization

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
daw3rd: Good question.  not sure we need that in the first pass, but open to suggestions.  Initially this may be for the non-launched/embedded/notebooks.

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
daw3rd: don't they all only operate on the extension?

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
daw3rd: ah good.  yes, A->B->C,  A.flush() has to be fed to B.transform().  ugly, but true.

data-processing-lib/python/test/data_processing_tests/transform/test_binary_pipeline.py
daw3rd: its a start.  that said, we're not testing the application of the underlying transforms, as much as the structure of the output.  but yes, would nice to have a better test, but would require having transforms othre than NOOP in the test_support packages.

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
blublinsky: not all of them

blublinsky:
These issues just magnify why this approach is not starter (in my opinion)

data-processing-lib/python/src/data_processing/transform/binary_pipeline.py
daw3rd: i could use use original name, but then using the same name for all seems incorrect.  Suggestion?

data-processing-lib/python/test/data_processing_tests/transform/test_binary_pipeline.py
daw3rd: Ah you are referring to the configuration part.  We have always been saying that transforms can be configured outside of the CLI/runtime mechanics.  I'm doing that here.  However, it is true, that to run a pipeline transform in a runtime may require more work - this is more for the python only non-runtime users, at least initially.

daw3rd:
```python
from typing import Any
import pyarrow as pa
from pyarrow import Table
from data_processing.transform import AbstractTableTransform
from data_processing.transform.binary_pipeline import PipelinedBinaryTransform
from data_processing.utils import TransformUtils

class HelloTransform(AbstractTableTransform):
    """" Adds a column of greetings """
    def __init__(self, config:dict):
        self.who = config.get("who", "World")
        self.column_name = config.get("column_name", "greeting")

    def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Table], dict[str, Any]]:
        # Create a new column with each row holding the who value
        new_column = ["Hello " + self.who + "!"] * table.num_rows
        # Append the column to create a new table with the configured name.
        table = TransformUtils.add_column(table=table, name=self.column_name, content=new_column)
        return [table], {}

if __name__ == "__main__":

    ids = pa.array([0 ])
    contents = pa.array(["Some content"])
    names = ["doc_id", "content"]
    table = pa.Table.from_arrays([ids, contents], names=names)

    hello0 = HelloTransform({})
    hello1 = HelloTransform({"column_name": "greating2", "who": "David"})
    hello2 = HelloTransform({"column_name": "greating3", "who": "Boris"})
    pipeline = PipelinedTransform({"transforms": [ hello0, hello1, hello2]})
    tables, metadata = pipeline.transform(table)
    table = tables[0]
    print(f"table={table}")
```
produces
```
table=pyarrow.Table
doc_id: int64
content: string
greeting: string
greating2: string
greating3: string
----
doc_id: [[0]]
content: [["Some content"]]
greeting: [["Hello World!"]]
greating2: [["Hello David!"]]
greating3: [["Hello Boris!"]]
```

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -100,8 +100,9 @@ def get_table(self, path: str) -> tuple[pa.table, int]:
         """
         Get pyArrow table for a given path
         :param path - file path
-        :return: pyArrow table or None, if the table read failed and number of operation retries.
-                 Retries are performed on operation failures and are typically due to the resource overload.
+        :return: Tuple containing
+            pyarrow.Table: PyArrow table if read successfully, None otherwise.
+            the number of retries. Retries are performed on operation failures and are typically due to the resource overload.
         """
         pass
```
blublinsky, line 108: we already have it https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/python/src/data_processing/data_access/data_access.py#L99

data-processing-lib/python/src/data_processing/data_access/data_access_local.py
```diff
@@ -260,8 +260,9 @@ def get_table(self, path: str) -> tuple[pa.table, int]:
         Args:
             path (str): Path to the file containing the table.
 
-        Returns:
+        Returns: Tuple containing
             pyarrow.Table: PyArrow table if read successfully, None otherwise.
+            the number of retries.
         """
 
         try:
```
blublinsky, line 268: We already have it https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/python/src/data_processing/data_access/data_access_local.py#L256

blublinsky (changes requested):

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+            from the name of the generating transform class name, but should not be relied on.
+        * If a transform produces multiple outputs (must be with the same extension) each output is applied through
+            the subsequent transforms in the pipeline.
+    Restrictions include:
+        * metadata produced is merged across all transforms, for any given call to transform/flush_binary() methods.
+    """
+
+    def __init__(self, config: dict[str, Any]):
+        """
+        Create the pipeline using a list of initialize transforms
```
blublinsky, line 32: initialized?

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+            config:  dictionary holding the following keys
+                transforms : a list of AbstractBinaryTransform instances.  All transforms must expect and produce
+                the same data type (within the binary array) represented by the file extensions passed into and
+                returned by the transform/flush_binary() methods.
+        """
+        super().__init__(config)
+        self.input_extension = None
+        self.transforms = config.get(transform_key, None)
+        if self.transforms is None:
+            raise ValueError(f"Missing configuration key {transform_key} specifying the list of transforms to run")
```
blublinsky, line 43: Also make sure the list is not empty and actually contains transforms, not some garbage

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+        Args:
+            file_name:
+            byte_array:
+        Returns:
+
+        """
+        r_bytes, r_metadata = self._apply_transforms_to_datum(self.transforms, (file_name, byte_array))
+        return r_bytes, r_metadata
+
+    def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Table], dict[str, Any]]:
```
blublinsky: why is this table, it should be binary. it should check the extension and then decide whether this is a table

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+            r_bytes.append(tp)
+
+        return r_bytes, r_metadata
+
+    def _call_transform(
+        self,
+        transform: Union[AbstractTableTransform, AbstractBinaryTransform],
+        datum: Union[pa.Table, bytearray],
+        file_name: str,
+    ):
```
blublinsky: This seems to be wrong. Transform is always binary. A table is just one of the cases of binary. no need for Unions. Take a look at https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/python/src/data_processing/transform/table_transform.py. it takes binary and returns binary. Table is just an intermediate format

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+        if is_table_transform:
+            tables, metadata = transform.transform(datum, file_name)
+            transformation_tuples = []
+            for table in tables:
+                transformation_tuples.append((table, ".parquet"))
+        else:
+            transformation_tuples, metadata = transform.transform_binary(file_name, datum)
+        return transformation_tuples, metadata
+
+    def _call_flush(self, transform: Union[AbstractTableTransform, AbstractBinaryTransform]):
```
blublinsky, line 183: Here this union stuff again

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
```diff
@@ -0,0 +1,202 @@
...
+        is_table_transform = isinstance(transform, AbstractTableTransform)
+        if is_table_transform:
+            tables, metadata = transform.flush()
+            transformation_tuples = []
+            for table in tables:
+                transformation_tuples.append((table, ".parquet"))
+        else:
+            transformation_tuples, metadata = transform.flush_binary()
```
blublinsky, line 191: This should be the only method called. Why is it more complex that it needs to be

data-processing-lib/python/test/data_processing_tests/transform/test_binary_pipeline.py
blublinsky: I repeat my sentiment. Its circumventing the "normal" execution

blublinsky:
This is breaking cli

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
daw3rd: I'm trying to avoid having someone that wants to transform a table, first having to convert to bytes to call transform_binary().  i think as you said, we may need to restrict this to transform_binary(), or allow transform() but check to be sure it is a Table.  Or, have transform_binary() do the type checking.  I think it would be useful though to avoid unnecessary conversion back and forth of Table to bytes and bytes to Table, for each transform.

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
daw3rd: Understood, I was/am trying to find a way to avoid requiring the user to call transform_binary() if they are starting with a Table.  And if the list of transforms is all TableTransforms, we should try and avoid two byte/Table conversions on each transform.

data-processing-lib/python/src/data_processing/transform/pipelined_transform.py
daw3rd: Again, trying to avoid byte/Table conversion, but not sure its ready yet.

daw3rd:
> This is breaking cli

I have not completed that part yet.  This is still a Draft afterall :)

PR #427, opened 2024-07-21
Fix get config parameter function.

### LINKED ISSUE #425: [Bug] get_config_parameter returns without checking if the config value exists
### Search before asking

### Component

Library/core

The method `get_config_parameter` in `data-processing-lib-python/src/data_processing/utils/params_utils.py` returns the first config key encountered. In the usage scenario where more than one config key is mentioned but only one key has a valid value, the method returns the first config key with possibly empty string. 

The method should check if the key has a non-zero length value before treating this key as a valid configuration. 

### Reproduction script

The issue was encountered when both `data_s3_config `and `data_lh_config` keys are present while `data_lh_config` was an empty string. 

### Anything else

Possible fix: 

```
def get_config_parameter(params: dict[str, Any]) -> str:
    """
    Get parameters for multi launch
    :param params: original parameters
    :return: tuple of name of repeated parameter, list of repeated parameters, and dict of non changing ones
    """
    # find config parameter
    config = None
    for key in params.keys():
      if key.startswith("data") and key.endswith("config"):
        if len(params[key]) > 1:
          config = key
          break
    return config
```

@dtsuzuku-ibm san discovered this bug. 

### OS

Ubuntu

### Python

3.10.x

Mohammad-nassar10:
Fixing a bug in getting the config parameter. Fixed according to @yuanchi2807 suggestion.

Fixes #425

### DIFF
--- a/data-processing-lib/python/src/data_processing/utils/params_utils.py
+++ b/data-processing-lib/python/src/data_processing/utils/params_utils.py
@@ -94,16 +94,17 @@ class ParamsUtils:
     @staticmethod
     def get_config_parameter(params: dict[str, Any]) -> str:
         """
-        Get parameters for multi launch
+        Get the key name of the config parameter
         :param params: original parameters
-        :return: tuple of name of repeated parameter, list of repeated parameters, and dict of non changing ones
+        :return: the name of the key for the config parameter
         """
         # find config parameter
         config = None
         for key in params.keys():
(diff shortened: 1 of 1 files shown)


yuanchi2807 (commented):
WIth the change, there will not be side effects elsewhere I hope.

blublinsky (approved):

blublinsky:
@yuanchi2807 none whatsoever

PR #430, opened 2024-07-22
Fix resize to not switch between schemas and add permissive_schema_merge option

### LINKED ISSUE #429: [Bug] Resize behaves badly when there are lots of schema changes
### Search before asking

### Component

Transforms/Other

Consider the following, using  SCHEMA(size) notation. 

If resize encounters, A(1), A(1), B(1), A(1), A(1), B(1) and you  are merging at a size of 3, the only thing that gets spit out is the last B(1), as it is today.  I'm arguing for a fix so that it focuses/keeps schema A and then you'll at least get A(3), A(1).  

### Reproduction script

None

### Anything else

_No response_

### OS

Other

### Python

3.10.x, 3.11.x

- [X] Yes I am willing to submit a PR!

daw3rd:
Fixes a problem in arising when there are lots of schema mismatches encountered.
Also adds `permissive_schema_merge` option and cli parameter

Fixes #429

### DIFF
--- a/transforms/universal/resize/python/src/resize_transform.py
+++ b/transforms/universal/resize/python/src/resize_transform.py
@@ -21,14 +21,17 @@ from data_processing.utils import LOCAL_TO_DISK, MB, CLIArgumentProvider, get_lo
 max_rows_per_table_key = "max_rows_per_table"
 max_mbytes_per_table_key = "max_mbytes_per_table"
 size_type_key = "size_type"
+permissive_schema_merge_key = "merge_schemas"
 shortname = "resize"
 cli_prefix = f"{shortname}_"
 max_rows_per_table_cli_param = f"{cli_prefix}{max_rows_per_table_key}"
 max_mbytes_per_table_cli_param = f"{cli_prefix}{max_mbytes_per_table_key}"
 size_type_cli_param = f"{cli_prefix}{size_type_key}"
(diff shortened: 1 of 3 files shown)


transforms/universal/resize/python/src/resize_transform.py
```diff
@@ -71,12 +85,10 @@ def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Tab
                 self.logger.debug(
                     f"concatenating buffer with {self.buffer.num_rows} rows to table with {table.num_rows} rows"
                 )
-                table = pa.concat_tables([self.buffer, table], unicode_promote_options="permissive")
+                table = pa.concat_tables([self.buffer, table], unicode_promote_options=self.unicode_promote_options)
                 self.logger.debug(f"concatenated table has {table.num_rows} rows")
                 self.buffer = None
             except Exception as e:  # Can happen if schemas are different
-                # Throw away the buffer and try and keep the current table by placing it in the buffer.
```
blublinsky, line 78: Sorry, as I explained before the code is right

blublinsky (changes requested):

daw3rd:
We have agreed that the real solution is to stop all processing of files in this case, which requires an enhancement to the runtime framework.

PR #432, opened 2024-07-22
remove --force-reinstall on pip installs

daw3rd:
This was causing some build failures due to uninstalling the source level dependencies.  Originally found by @blublinsky


blublinsky:
Are you just trying to piecemeal https://github.com/IBM/data-prep-kit/pull/413? Not a good idea

daw3rd:
This change went into the merged PR #452

PR #438, opened 2024-07-23
Fix repo lvl skip unreadable files

shivdeep-singh-ibm:

### DIFF
--- a/transforms/code/repo_level_ordering/ray/src/dpk_repo_level_order/internal/repo_grouper.py
+++ b/transforms/code/repo_level_ordering/ray/src/dpk_repo_level_order/internal/repo_grouper.py
@@ -42,6 +42,8 @@ class GroupByRepo:
 
     def process(self, repo: str, files: List[str]):
         cumulated_table = self._read_parquet_bulk(files)
+        if cumulated_table is None:
+            return
         repo_table = self._filter_table_by_column(cumulated_table, self.repo_column_name, repo)
 
         def sanitize_path(repo_name):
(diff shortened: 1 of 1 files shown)


transforms/code/repo_level_ordering/ray/src/dpk_repo_level_order/internal/repo_grouper.py
```diff
@@ -62,14 +64,26 @@ def _write_parquet(self, table, repo_name):
...
+            table, _ = self.data_access.get_table(path)
+            if table is not None:
+                return table.to_pandas()
+            else:
+                return None
+
+        tables_ = list(
             map(
-                lambda x: self.data_access.get_table(os.path.normpath(x))[0].to_pandas(),
+                lambda x: get_dataframes(os.path.normpath(x)),
```
Param-S, line 76: if lamda function failed to read one or more files, then function should record the failure. Currently, it ignores the failure. IMHO, it is critical error, it should flag the error & terminate.

Param-S (changes requested):

transforms/code/repo_level_ordering/ray/src/dpk_repo_level_order/internal/repo_grouper.py
daw3rd: The recommendation is to have the transform() throw the exception to the runtime.  See https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/doc/transform-exceptions.md

shivdeep-singh-ibm:
not required.

PR #446, opened 2024-07-25
fix path in notebook

shivdeep-singh-ibm:

since notebook has moved into folder, we need to update paths of `lang_extensions.json` wrt new notebook path


shahrokhDaijavad (approved):
LGTM. I did test this by going through all cells of the Notebook.

PR #456, opened 2024-07-25
Fixes some bad assert statements and adds fuzzy floating point comparison 2

daw3rd:

1. Need to fix asserts that were not being tested properly in test_support abstract_test.py
2. Adds support for floating point values to be within 1% of each other in parquet in anticipation of model scores being different values on different architectures.
3. Required updating some of the binary transform test files since we were not correctly comparing sizes until now.


data-processing-lib/python/src/data_processing/test_support/abstract_test.py
```diff
@@ -96,7 +99,46 @@ def validate_expected_tables(table_list: list[pa.Table], expected_table_list: li
...
+        of each other as defined by global _allowed_float_percent_diff.
+        We assume the schema has already been compared and is equivalent.
+        Args:
+            table_index: index of tables that is the source of the rows.
+            row_index:
+            test_row:
+            expected_row:
+        """
+        assert test_row.num_rows == 1, "Invalid usage.  Expected test table with 1 row"
+        assert expected_row.num_rows == 1, "Invalid usage.  Expected expected table with 1 row"
```
blublinsky, line 118: What exactly are we asserting here? That the table has more than 1 row? Why is this required?

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
```diff
@@ -96,7 +99,46 @@ def validate_expected_tables(table_list: list[pa.Table], expected_table_list: li
...
+        if test_row != expected_row:
+            # Else look for floating point values that might differ within the allowance
+            msg = f"Row {row_index} of table {table_index} are not equal\n\tTransformed: {test_row}\n\tExpected   : {expected_row}"
+            test_columns = test_row
+            expected_columns = test_row
```
blublinsky: Why these extra variables? did you mean:
test_columns = test_row.schema.columns
expected_columns = expected_row.schema.columns

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
daw3rd: only to be sure that the caller (us) is using this method correctly.

daw3rd: and the code that follows expects a single row

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
daw3rd: actually, there is a bug here.  expected_columns = test_row.  Will remove and fix bug.

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
```diff
@@ -96,7 +99,45 @@ def validate_expected_tables(table_list: list[pa.Table], expected_table_list: li
...
+        We assume the schema has already been compared and is equivalent.
+        Args:
+            table_index: index of tables that is the source of the rows.
+            row_index:
+            test_row:
+            expected_row:
+        """
+        assert test_row.num_rows == 1, "Invalid usage.  Expected test table with 1 row"
+        assert expected_row.num_rows == 1, "Invalid usage.  Expected expected table with 1 row"
+        if test_row != expected_row:
```
blublinsky, line 119: This compares 2 tables, not the amount of columns in 2 tables. see https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.names to get column names

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
```diff
@@ -96,7 +99,45 @@ def validate_expected_tables(table_list: list[pa.Table], expected_table_list: li
...
+            # Else look for floating point values that might differ within the allowance
+            msg = f"Row {row_index} of table {table_index} are not equal\n\tTransformed: {test_row}\n\tExpected   : {expected_row}"
+            assert len(test_row) == len(expected_row), msg
```
blublinsky: len(table) returns number of rows, not number of columns. I think you want to compare number of columns

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
daw3rd: 1.  we've already verified the schema is equivalent 
2. what i think we want here is a full table equality check, which was what we were doing originally I believe. see https://github.com/IBM/data-prep-kit/blob/3ee26c500f55abdd1681ee5586bc4e8ea73d83a9/data-processing-lib/python/src/data_processing/test_support/abstract_test.py#L99 in 0.2.0

data-processing-lib/python/src/data_processing/test_support/abstract_test.py
daw3rd: nice.  done.

blublinsky (approved):

Issue #451, opened 2024-07-25
[Feature] Create a demo notebook for RAG


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

The goal is to create a demo notebook for RAG that stitches the following components mentioned in the [issue](https://github.com/orgs/IBM/projects/86/views/1?filterQuery=&pane=issue&itemId=72179810). The notebook should reside in this folder https://github.com/IBM/data-prep-kit/tree/dev/examples/notebooks

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

dolfim-ibm:
A first draft of the notebook is here https://github.com/IBM/data-prep-kit/pull/461. At the moment it still depends on some internal transforms which are being pushed soon.

Issue #448, opened 2024-07-25
[Feature] Create vector embeddings 


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

The goal is to add a new module that can create embeddings from a given document. The module will take input parquet files, where each row may contain a chunk or the full document. For every row, it will convert the existing text to embeddings and add the embeddings as a new column in the parquet files. 

This should be added as a new module along with the other language transforms here https://github.com/IBM/data-prep-kit/tree/dev/transforms/language

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

dolfim-ibm:
Done in https://github.com/IBM/data-prep-kit/pull/461

Issue #447, opened 2024-07-25
[Feature] Capability to chunk text for RAG systems


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

The goal is to add a new transform that can take in the extracted text and chunk it. The input will be parquet files where every document is stored in one row. The output will be chunks, such that every chunk is stored in one row. Chunk size should be a parameter exposed to the user. 

This new transform should be added along with other language modules here https://github.com/IBM/data-prep-kit/tree/dev/transforms/language

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

dolfim-ibm:
Done in https://github.com/IBM/data-prep-kit/pull/461

PR #461, opened 2024-07-28
Language notebook for RAG ingestion w/ chunker and encoder transforms.

dolfim-ibm:

This PR adds a notebook which orchestrates multiple transforms for a RAG ingestion, as well as new transforms for 
- Doc chunking, https://github.com/IBM/data-prep-kit/issues/447
- Text encoders, https://github.com/IBM/data-prep-kit/issues/448

See issue https://github.com/IBM/data-prep-kit/issues/451


transforms/language/lang_id/python/pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "data-prep-toolkit==0.2.1.dev0",
     "fasttext==0.9.2",
     "langcodes==3.3.0",
-    "huggingface-hub==0.21.4",
+    "huggingface-hub >= 0.21.4, <1.0.0",
```
dolfim-ibm, line 15: Note: this is needed to avoid conflicts between transforms and other dependencies.

transforms/language/text_encoder/README.md
```diff
@@ -0,0 +1,12 @@
+# Text encoder Transform 
+The Text encoder transforms serves as a simple exemplar to demonstrate the development
```
daw3rd: Copied from the NOOP transform. can you provide correct description here.

examples/notebooks/language/launch.sh
```diff
@@ -0,0 +1,22 @@
+# Set REPOROOÞ
+REPO_ROOT=$(cd ../../../ && pwd && cd - > /dev/null )
+echo $REPO_ROOT
+
+
+# Set PYTHONPATH for `data_processing_ray` library
+export PYTHONPATH=$REPO_ROOT/data-processing-lib/python/src
+export PYTHONPATH=$PYTHONPATH:$REPO_ROOT/data-processing-lib/ray/src:$REPO_ROOT/data-processing-lib/python/src
+
+# Set PYTHONAPATH for transforms
```
daw3rd: Maybe this can be deleted?

examples/notebooks/language/launch.sh
```diff
@@ -0,0 +1,22 @@
+# Set REPOROOÞ
+REPO_ROOT=$(cd ../../../ && pwd && cd - > /dev/null )
+echo $REPO_ROOT
+
+
+# Set PYTHONPATH for `data_processing_ray` library
+export PYTHONPATH=$REPO_ROOT/data-processing-lib/python/src
```
daw3rd: Think this first one can be removed

examples/notebooks/language/prepare_env.sh
```diff
@@ -0,0 +1,48 @@
+
+
+# Set REPOROOT
+REPO_ROOT=$(cd ../../../ && pwd && cd - > /dev/null )
+echo $REPO_ROOT
+
+pip install notebook ipywidgets
+
+# requirement_files=(
+# "$REPO_ROOT/tools/ingest2parquet/requirements.txt"
```
daw3rd: maybe delete these comments

examples/notebooks/language/requirements.txt
daw3rd, line 1: should this file be deleted?

transforms/language/text_encoder/python/src/text_encoder_transform.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Add Transform-specific arguments to the given  parser.
+        This will be included in a dictionary used to initialize the TextEncoderTransform.
+        By convention a common prefix should be used for all transform-specific CLI args
+        (e.g, noop_, pii_, etc.)
+        """
+        parser.add_argument(
+            f"--{content_column_name_cli_param}",
+            default=default_content_column_name,
+            help="Column name to get content",
```
daw3rd: For help text, maybe "Name of the column containing the text to be encoded"

transforms/language/text_encoder/python/src/text_encoder_transform.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        parser.add_argument(
+            f"--{output_embeddings_column_name_cli_param}",
+            default=default_output_embeddings_column_name,
+            help="Column name to store the embeddings",
+        )
+        parser.add_argument(
+            f"--{model_name_cli_param}",
+            default=default_model_name,
+            help="Name of the HF model to use for encoding the text",
```
daw3rd: --help from ArgumentParsrer does not seem to show the default values.  Maybe it would be helpful to include the default hf model name in the help text.

transforms/language/text_encoder/python/src/text_encoder_transform_python.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        Initialization
+        :param base_configuration - base configuration class
+        """
+        super().__init__(transform_config=TextEncoderTransformConfiguration())
+
+
+if __name__ == "__main__":
+    launcher = PythonTransformLauncher(TextEncoderPythonTransformConfiguration())
+    logger.info("Launching noop transform")
```
daw3rd: noop.  or feel free to remove this comment.

transforms/language/text_encoder/python/test/test_text_encoder.py
```diff
@@ -0,0 +1,45 @@
...
+    Extends the super-class to define the test data for the tests defined there.
+    The name of this class MUST begin with the word Test so that pytest recognizes it as a test class.
+    """
+
+    def get_test_transform_fixtures(self) -> list[tuple]:
+        fixtures = [
+            (
+                TextEncoderTransform({}),
+                [table],
+                [expected_table],
```
daw3rd: your input and expected tables are the same, suggesting your transform is really a noop.

transforms/language/text_encoder/ray/test/test_text_encoder_ray.py
```diff
@@ -0,0 +1,43 @@
...
+
+from data_processing.test_support.launch.transform_test import (
+    AbstractTransformLauncherTest,
+)
+from data_processing_ray.runtime.ray import RayTransformLauncher
+from noop_transform import sleep_cli_param
+from noop_transform_ray import NOOPRayTransformConfiguration
+
+
+class TestRayNOOPTransform(AbstractTransformLauncherTest):
```
daw3rd: NOOP

transforms/language/text_encoder/ray/test/test_text_encoder_ray.py
```diff
@@ -0,0 +1,43 @@
...
+    Extends the super-class to define the test data for the tests defined there.
+    The name of this class MUST begin with the word Test so that pytest recognizes it as a test class.
+    """
+
+    def get_test_transform_fixtures(self) -> list[tuple]:
+        basedir = "../test-data"
+        basedir = os.path.abspath(os.path.join(os.path.dirname(__file__), basedir))
+        fixtures = []
+        # launcher = NOOPRayLauncher()
+        launcher = RayTransformLauncher(NOOPRayTransformConfiguration())
```
daw3rd: NOOP

daw3rd (changes requested):

transforms/language/text_encoder/python/test/test_text_encoder.py
dolfim-ibm: Right, I wanted to ask about this one.

To create some example result, I would have to run the initialize and run the encoder manually.
Is there a need for this test, instead of the the one with the Launcher? In case, we can simply do the same, compare input/expected.

dolfim-ibm:
@daw3rd the CI fails because the vectors computed by the encoder model have differences in some digits. You were already doing something about it in other transforms, right? Is there a parameter for doing some "shallow diff" or similar?

daw3rd:
> @daw3rd the CI fails because the vectors computed by the encoder model have differences in some digits. You were already doing something about it in other transforms, right? Is there a parameter for doing some "shallow diff" or similar?

Yes, there is something for float values in the table comparison to allow for 1% differences.  But, no support for vectors of floats :(.  Can probably be added though.

examples/notebooks/language/input_data/Granite%20Foundation%20Models.pdf
daw3rd, line 1: maybe too late now, but do we need 5mb file?

transforms/language/doc_chunk/README.md
```diff
@@ -0,0 +1,11 @@
+# chunk documents Transform 
+This transform is chunking documents from their JSON representation to a list of chunks.
+It relies on documents converted with the Docling library in the [pdf2parquet transform](../pdf2parquet).
```
daw3rd: Maybe just add, that pdf2parquet provides the required JSON

transforms/language/doc_chunk/python/README.md
```diff
@@ -0,0 +1,54 @@
+# chunk documents Transform 
+
+This transform is chunking documents from their JSON representation to a list of text chunks.
+It relies on .parquet files created by the [pdf2parquet transform](../pdf2parquet) with the
+`contents_type: "application/json"` argument, which the [Docling library](https://github.com/DS4SD/docling)
+to convert PDF documents to its rich JSON representation.
+
+The new .parquet table will contain one row for each document component in the source table.
```
daw3rd: maybe a bit more about the structure components chunked?  Chapters, sections, paragraphs, sentences?

transforms/language/doc_chunk/python/README.md
```diff
@@ -0,0 +1,54 @@
...
+## Running
+
+### Parameters
+
+The transform can be tuned with the following parameters.
+
+
+| Parameter  | Default  | Description  |
+|------------|----------|--------------|
+| `chunking_type`        | `dl_json` | Chunking type to apply. Valid options are `li_markdown` for using the LlamaIndex [Markdown chunking](https://docs.llamaindex.ai/en/stable/module_guides/loading/node_parsers/modules/#markdownnodeparser), `dl_json` for using the [Docling JSON chunking](). |
```
daw3rd: 1.  Need a link for the Docling JSON chunking
2. I need some education here, but it is. unclear what chunking really means here.  Is it by size, semantics, etc.  And are you always chunking at say paragraphs or are you providing multiple chunkings (sentence + paragraph) in a single pass.

transforms/language/doc_chunk/python/pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+license = {text = "Apache-2.0"}
+readme = {file = "README.md", content-type = "text/markdown"}
+authors = [
+    { name = "Michele Dolfi", email = "dol@zurich.ibm.com" },
+    { name = "Panos Vagenas", email = "pva@zurich.ibm.com" },
+    { name = "Christoph Auer", email = "cau@zurich.ibm.com" },
+]
+dependencies = [
+    "data-prep-toolkit==0.2.1.dev0",
+    # "llama-index-core==0.10.58",
```
daw3rd: Did you mean to comment this out, given the readme's reference to llamaindex?

transforms/language/doc_chunk/python/src/doc_chunk_chunkers.py
```diff
@@ -0,0 +1,57 @@
...
+from docling_core.types import Document as DLDocument
+from quackling.core.chunkers.hierarchical_chunker import HierarchicalChunker
+from llama_index.core import Document as LIDocument
+from llama_index.core.node_parser import MarkdownNodeParser
+
+
+class ChunkingExecutor(metaclass=ABCMeta):
+    @abstractmethod
+    def chunk(self, content: str) -> Iterator[dict]:
+        pass
```
daw3rd: maybe and exception here would be better for quick notification to developers of what mistake they've made?

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
```diff
@@ -0,0 +1,194 @@
...
+    configuration with CLI args.
+    """
+
+    def __init__(self):
+        super().__init__(
+            name=short_name,
+            transform_class=DocChunkTransform,
+        )
+
+        self.logger = get_logger(__name__)
```
daw3rd: It turns out that using the same logger name here, will currently (due to a "bug") result in double logging.  For now, just maybe use `__name__ + config`

daw3rd (changes requested):
Review of chunker done.  Looks really good with only some minor changes.

transforms/language/text_encoder/python/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Running
+
+### Parameters
+
+The transform can be tuned with the following parameters.
+
+
+| Parameter  | Default  | Description  |
+|------------|----------|--------------|
+| `model_name`                    | `BAAI/bge-small-en-v1.5` | The HF model to use for encoding the text. |
```
daw3rd, line 18: If this is currently restricted to hf models, maybe call this key `hf_model_name`?  This might allow more flexibility going forward to add new model types/providers.  I like the generic `model_name`, but what happens when we use other model types.   Another option is to require the value to include the indicator.  For example `hf:BAAI/bge-small-en-v1.5`

transforms/language/text_encoder/python/README.md
```diff
@@ -0,0 +1,47 @@
+# Text Encoder Transform 
+
+## Summary 
+This transform is using sentence encoder models to create embeddings of the text in each row of the input .parquet table.
```
daw3rd: Can you provide some reference or documentation for what an `embedding` is and how it might be used downstream?  Yes, I'm still learning, but others might be as well :).

transforms/language/text_encoder/python/src/text_encoder_transform.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(self):
+        super().__init__(
+            name=short_name,
+            transform_class=TextEncoderTransform,
+            # remove_from_metadata=[pwd_key],
+        )
+        from data_processing.utils import get_logger
+
+        self.logger = get_logger(__name__)
```
daw3rd: same comment on duplicate logging as for chunker.  Recommend using `__name__ + "cfg"` or some such to distinguish name from one used by the transform itself.

examples/notebooks/language/input_data/Granite%20Foundation%20Models.pdf
dolfim-ibm: should we have a script which does `curl` of some arxiv doc in the prepare script?

transforms/language/text_encoder/python/test/test_text_encoder_python.py
```diff
@@ -0,0 +1,43 @@
...
+        fixtures = []
+        launcher = PythonTransformLauncher(TextEncoderPythonTransformConfiguration())
+        fixtures.append(
+            (
+                launcher,
+                {},
+                basedir + "/input",
+                basedir + "/expected",
+                # this is added as a fixture to remove these columns from comparison
+                ["embeddings"],
```
daw3rd, line 40: Are the embedding supposed to be reproducible (modulo floating point precision)?  If so, how do we deal with this - if checked in as is, we should have an issue to go with it.  I think the right solution may be to truncate floating point values in the embeddings vector to some level or reasonable precision that will make the tests pass.

transforms/language/text_encoder/ray/test/test_text_encoder_ray.py
```diff
@@ -0,0 +1,43 @@
...
+        fixtures = []
+        launcher = RayTransformLauncher(TextEncoderRayTransformConfiguration())
+        fixtures.append(
+            (
+                launcher,
+                {"run_locally": True},
+                basedir + "/input",
+                basedir + "/expected",
+                # this is added as a fixture to remove these columns from comparison
+                ["embeddings"],
```
daw3rd, line 40: Same comment as on python test.

daw3rd (changes requested):
Great work!

transforms/language/doc_chunk/python/README.md
dolfim-ibm: 1. done, 2. depends on the method chosen. I tried to add some description in the piece above.

transforms/language/doc_chunk/python/pyproject.toml
dolfim-ibm: it is already installed as a dependency of quackling. adding again created some include loop..

transforms/language/doc_chunk/python/src/doc_chunk_chunkers.py
dolfim-ibm: done

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
dolfim-ibm: done

transforms/language/text_encoder/python/README.md
dolfim-ibm: if we use a different provider, very likely we need to have a different execution code, because the `model_name` is used by the HF `transforms` library directly.

If/when we get to other providers, I think we will need a "executor factory" and a corresponding parameter to change it.

transforms/language/text_encoder/python/test/test_text_encoder_python.py
dolfim-ibm: yes, we need an issue for being to address this at the core of the DPK, it cannot be the responsibility of each transform to implement such a code. (but it is its responsibility to know _when_ it could be used)

transforms/language/text_encoder/python/src/text_encoder_transform.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+
+    def __init__(self, config: dict[str, Any]):
+        """ """
+        # Make sure that the param name corresponds to the name used in apply_input_params method
+        # of TextEncoderTransform class
+        super().__init__(config)
+        from data_processing.utils import get_logger
+
+        self.logger = get_logger(__name__ + "config")
```
daw3rd: the "config" should probably go on the other logger from TextEncoderTransformConfiguration

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        Initialize based on the dictionary of configuration information.
+        This is generally called with configuration parsed from the CLI arguments defined
+        by the companion runtime, DocChunkTransformRuntime.  If running inside the RayMutatingDriver,
+        these will be provided by that class with help from the RayMutatingDriver.
+        """
+        # Make sure that the param name corresponds to the name used in apply_input_params method
+        # of DocChunkTransformConfiguration class
+        super().__init__(config)
+        self.logger = get_logger(__name__ + "config")
```
daw3rd: Same here, to be meaningful and not obfuscate things, "config" should be used in the DocChunkTransformConfiguration class's logger.

transforms/language/text_encoder/python/src/text_encoder_transform.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        parser.add_argument(
+            f"--{output_embeddings_column_name_cli_param}",
+            default=default_output_embeddings_column_name,
+            help="Column name to store the embeddings",
+        )
+        parser.add_argument(
+            f"--{model_name_cli_param}",
+            default=default_model_name,
+            help="Name of the HF model to use for encoding the text. The default model is BAAI/bge-small-en-v1.5",
```
daw3rd: You should use default_model_name here.

transforms/language/text_encoder/python/test/test_text_encoder_python.py
daw3rd: I think it is not reasonable to expect DPK to support fuzzy comparison of floats in arbitrarily structured python objects.  We start with float, the list of float, then dictionary of some floats, the nested dictionaries of some floats, etc.  If it can't be done in general, it seems the transform should be responsible.

transforms/language/doc_chunk/python/README.md
daw3rd: I'm still not clear on what is happening.  The document will have many ways to be chunked.  Is it by sentence, paragraph, section, etc. only. so that each row contains one of those.  Or does the checking happen at all those levels, so that some rows contains sentences and some paragraphs?  There does not seem to be a setting to control what the chunk size is, so perhaps it is the latter?  Sorry for my lack of familiarity here, but others may have the same issue.

examples/notebooks/language/output.zip
daw3rd, line 1: does this file need to be in the repo?

transforms/language/text_encoder/python/src/text_encoder_transform.py
daw3rd: i fixed this one

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
daw3rd: i fixed this one

daw3rd (approved):
With the caveat that soon issues #483 and #482 are addressed soon.

PR #464, opened 2024-07-30
Data profiling tool for dataset insights

pankajskku:
	This tool generates an output with different metrics that can help a user quickly understand insights from the data (code and text) and see if the data is fit for their use.
	
Data profiler can be usefulfor personas like Developers, Business Users, Data Scientists for use cases like,
* Understand the data before starting any data prep
* Understand data issues found during data prep
* Profile validation sets to understand where the model does well and where it does not do well
* Help customer build the right validation set based on their needs

Features:  
* Works on both code files and instruction pairs
* Data stats to help the user make recommendations during model tuning, eg * padding sizes
* Program Analysis based stats for data coverage
* Data prep-based stats on issues found in the data


Bytes-Explorer:
Thank you for your contribution! Can we rename this to code-profiler and add it in the [code folder ](https://github.com/IBM/data-prep-kit/tree/dev/transforms/code) ?

pankajskku:
> Thank you for your contribution! Can we rename this to code-profiler and add it in the [code folder ](https://github.com/IBM/data-prep-kit/tree/dev/transforms/code) ?

@Bytes-Explorer  I kept it generic so that in future we can some profiling models for the natural language data.

daw3rd:
Couple of thoughts

1.  Can this be made into a transform instead of a tool.  This will help scalability.
2. We currently have a ["profiler" transform](https://github.com/IBM/data-prep-kit/tree/dev/transforms/universal/profiler) that only counts words.  Would it be possible to extend this profiler transform with the capabilities here?

Bytes-Explorer:
Agree on making it a transform. I would prefer that we keep the code and natural language profilers separate, and keep them under natural language and code folders, so it is easy to find.

pankajskku:
> Agree on making it a transform. I would prefer that we keep the code and natural language profilers separate, and keep them under natural language and code folders, so it is easy to find.

@Bytes-Explorer 
There could be a case where a common profiler could be applied for both the code and text data. I created the folder hierarchy of the data-profiler as per the current folder convention defined in the data-prep-kit's transforms, where the transforms are categorized into three types code, text, and universal.

transform's hierarchy:
image

data-profiler hierarchy:
image

The data profiler is not transforming (changing/modifying/annotating) the data but instead providing insights into the dataset based on some metrics in the form of JSON metadata file. In this case is it still okay to call it a transform?

Current PR has support for Python runtime, in future PRs, we are planning to add support for more runtimes to make it scalable to run.

Bytes-Explorer:
@pankajskku We have taken a call that everything will be now under transforms, so will be good if we can add it accordingly. My understanding is that this code does not have any profiling information for NLP, so if it is just code, lets add it as a separate module under code folder. As and when we get NLP based profiling ready, we can extend the current available module that David pointed to. Does this work for you?

pankajskku:
In that case, it works great for me too. I'll move the code to transforms and update this PR with a modified structure. Till then I request @shivdeep-singh-ibm to pause the code review.

daw3rd:
As to language profiler, we currently have the doc_quality profiler/transform under language, see below.  What is the difference between these two?

https://github.ibm.com/ai-models-data/data-prep-kit-inner/tree/multimodal/transforms/language/doc_quality/ray

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+from data_processing.transform import AbstractTableTransform
+from langdetect import detect
+from pygments.lexers import get_all_lexers, guess_lexer
+from pygments.util import ClassNotFound
+
+
+class DatasetStatsProfiler(AbstractTableTransform):
+    """
+    Implements a simple copy of a pyarrow Table.
+    """
```
blublinsky, line 30: wrong comment?

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+        by the companion runtime, DatasetStatsProfilerRuntime.  If running inside the RayMutatingDriver,
+        these will be provided by that class with help from the RayMutatingDriver.
+        """
+        # Make sure that the param name corresponds to the name used in apply_input_params method
+        # of DatasetStatsProfilerConfiguration class
+        super().__init__(config)
+        from data_processing.utils import get_logger
+
+        self.logger = get_logger(__name__)
+
```
blublinsky, line 45: pls remove logger here. It is defined by super class

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+        if not isinstance(df, pd.DataFrame):
+            raise TypeError("Expected df to be a pandas DataFrame")
+
+        # Convert all columns to string type safely, handling NaN and other types
+        df = df.apply(
+            lambda col: col.map(lambda x: str(x) if not pd.isna(x) else "") if isinstance(col, pd.Series) else col
+        )
+
+        # Print column names for debugging
+        print("DataFrame columns:", df.columns)
```
blublinsky, line 69: please replace print to logger everywhere

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # 1. Rows and Unique Rows
+        total_rows = len(df)
+        unique_rows = len(df.drop_duplicates())
```
blublinsky, line 73: you should use dedup for this, not do it here

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+        super().__init__(config)
+        from data_processing.utils import get_logger
+
+        self.logger = get_logger(__name__)
+
+    def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Table], dict[str, Any]]:
+        """
+        Computes the dataset statistics
+        """
+        self.logger.debug(f"Transforming one table with {len(table)} rows")
```
blublinsky, line 50: You should verify that all required columns are peresent

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+        # Ensure df is a DataFrame
+        if not isinstance(df, pd.DataFrame):
+            raise TypeError("Expected df to be a pandas DataFrame")
+
+        # Convert all columns to string type safely, handling NaN and other types
+        df = df.apply(
+            lambda col: col.map(lambda x: str(x) if not pd.isna(x) else "") if isinstance(col, pd.Series) else col
+        )
+
+        # Print column names for debugging
```
blublinsky, line 68: This conversion to pandas is completely unnecessary. You can do the same operations on pyarrow table itself without virtually any changes but will make code smaller and faster

tools/data-profiler/universal/dataset_stats/python/src/dataset_stats_profiler.py
```diff
@@ -0,0 +1,316 @@
...
+            "sorted_prog_lang_count": sorted_prog_lang_count,
+            "language_code_sizes_list": language_code_sizes_list,
+        }
+
+    def calculate_word_counts(self, df: pd.DataFrame, column: str) -> dict[str, Any]:
+        """
+        Calculate word count statistics for a specific column in the DataFrame.
+        """
+        # Check if the column exists in the DataFrame
+        if column not in df.columns:
```
blublinsky, line 250: All this column chacks should be done upfront in one place

Issue #62, opened 2024-05-03
[Feature] New transform to do PII detection on code files


Bytes-Explorer:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

PII detection and redaction is important step while preparation of data for LLMs. The request is to build a new transform that can perform PII detection on code entities. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

SowmyaLR:
I am willing to work on this card. Is there any deadline to work on for this card? @Bytes-Explorer

Bytes-Explorer:
Thanks @SowmyaLR for your interest. Couple of weeks should be a good time frame.

Bytes-Explorer:
@SowmyaLR We are doing a demo in two weeks where this module can be showcased, if ready. Do you think you can raise a PR in the next couple of days? Please feel free to ask for help.

SowmyaLR:
@Bytes-Explorer I can raise PR by next week. Yeah had some issues and figured it out. I hope this can also go along for the demo 🤞

Bytes-Explorer:
Awesome @SowmyaLR ! 
cc @shivdeep-singh-ibm As and when this PR is raised, lets fast track the review process on it.

SowmyaLR:
@Bytes-Explorer @shivdeep-singh-ibm 
Raised [PR](https://github.com/IBM/data-prep-kit/pull/468) for this. I have added few comments in PR description. Need suggestions from people to know how to resolve that.

PR #468, opened 2024-07-31
Pii transform

SowmyaLR:
Included a new transform for pii redaction

https://github.com/IBM/data-prep-kit/issues/62

Need some suggestion for writing ray and spark version of pii transform:

1. [flair_recognizer](https://github.com/SowmyaLR/data-prep-kit/blob/pii_redaction_transform/transforms/universal/pii_redactor/python/src/flair_recognizer.py)
2. [pii_analyzer](https://github.com/SowmyaLR/data-prep-kit/blob/pii_redaction_transform/transforms/universal/pii_redactor/python/src/pii_analyzer.py)

the above files are common for all framework (python, ray, spark) of pii reactor transform. 
Shall we keep it in this repo at root level of this transform or do we want to move this to any other repo and include that as dependency here?


transforms/universal/pii_redactor/python/src/pii_redactor_transform.py
```diff
@@ -0,0 +1,137 @@
...
+        #     transformed_data if name == self.doc_contents_key else table[name] for name in table.column_names
+        # ]
+        # transformed_table = pa.Table.from_arrays(redacted_data, table.schema)
+        # metadata["transformed_table_rows"] = transformed_table.num_rows
+
+        for column in table.column_names:
+
+            if column == self.doc_contents_key:
+                column_data = pa.array(table[column].to_pandas().apply(self._redact_pii))
+                table = table.add_column(0, "new_contents", [column_data])
```
santoshborse, line 85: you can rewrite this block to make it more readable
```
if self.doc_contents_key in table.column_names:
    column_data = pa.array(table[self.doc_contents_key].to_pandas().apply(self._redact_pii))
    table = table.add_column(0, "new_contents", [column_data])
```

santoshborse (changes requested):

transforms/universal/pii_redactor/python/src/pii_redactor_transform.py
santoshborse: in fact you don't even need `if` line as you have ready have used `TransformUtils.validate_columns` make sure column `self.doc_contents_key` is in table

SowmyaLR:
@santoshborse seems  conflict issues. Will fix it and raise a new PR.

PR #469, opened 2024-07-31
pdf2parquet updates

dolfim-ibm:

Updates to the Docling package and updates of the test results.


dolfim-ibm:
@daw3rd I'm disabling the JSON test, because it fails on floating point precision. See screenshot.
To enable such a test we would need a test validator which would allow for small precision differences.

image

daw3rd (approved):
LGTM

PR #492, opened 2024-08-09
Adding support for python multiprocessing pool

blublinsky:
As more emphasis is towards pure Python runtime, usage of multiprocessing support can improve performance for this runtime


data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+        self.multiprocessing = False
+        self.num_processors = 0
+
+    def add_input_params(self, parser: argparse.ArgumentParser) -> None:
+        """
+        This method adds transformer specific parameter to parser
+        :param parser: parser
+        :return:
+        """
+        parser.add_argument(f"--{cli_prefix}multiprocessing", type=bool, default=False, help="use multiprocessing pool for execution")
```
daw3rd: can we make the default be multiprocessing?  Why not?  Yes, not quite backwards compatible, but not sure it would really break anything.

daw3rd: and then as a default number of processors something like `available-1`?

data-processing-lib/python/src/data_processing/runtime/transform_file_processor.py
```diff
@@ -21,17 +23,26 @@ class AbstractTransformFileProcessor:
...
+            self,
+            data_access_factory: DataAccessFactoryBase,
+            transform_parameters: dict[str, Any],
+    ):
         """
         Init method
+        :param data_access_factory: Data Access Factory
+        :param transform_parameters: Transform parameters
         """
-        self.data_access = None
```
daw3rd, line 28: As long as we're moving this around, can we name it what it is, that is data_access_factory?  Same for the key. it should be "data_access_factory", although for backwards compatibility, we need to keep "data_access".

data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
daw3rd: And/or, maybe we only need `num_processors` parameter and let the user set to 1 instead of setting multiprocessing=False.  For example, what if they set multiprocessing=True, but num_processors=1.  Yes this would be odd/broken usage, but the point is that it is ambiguous.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
```diff
@@ -54,34 +55,40 @@ def orchestrate(
...
-                    f"in {(time.time() - t_start)/60} min"
-                )
-        logger.debug(f"Done processing {completed} files, waiting for flush() completion.")
-        # invoke flush to ensure that all results are returned
-        start = time.time()
-        executor.flush()
-        logger.info(f"done flushing in {time.time() - start} sec")
+        if execution_config.multiprocessing:
+            # using multiprocessor pool for execution
+            statistics = process_transforms(files=files, size=execution_config.num_processors,
```
daw3rd: this is asymetric - in one case there is a single method and the other case a class plus lots of code.  What if we add a `process_files(...,files:list[string]...)` method to the `AbstractTransformFileProcessor` class and then have different implementations for serial and parallel?

daw3rd (changes requested):

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
daw3rd: Actually maybe we need an `AbstractPythonTransformFileProcessor` that adds this `process_files()` method and then two implementations for the two sub-classes?

data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
blublinsky: I kinda like for the user to decide whether he wants to use multiprocessing explicitly to make sure there are no surprises

data-processing-lib/python/src/data_processing/runtime/transform_file_processor.py
blublinsky: But its really data access:
```
self.data_access = data_access_factory.create_data_access()
```

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
blublinsky: changed layout

data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
blublinsky: becides, for small amount of files multiprocessing can be slower

data-processing-lib/python/src/data_processing/runtime/transform_file_processor.py
daw3rd: ah good. i stand corrected.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_file_processor.py
```diff
@@ -31,16 +31,14 @@ def __init__(self, params: dict[str, Any]):
...
-        # Create data access
-        self.data_access = params.get("data_access_factory", None).create_data_access()
-        # Add data access ant statistics to the processor parameters
-        transform_params = params.get("transform_params", None)
-        transform_params["data_access"] = self.data_access
+        super().__init__(
+            data_access_factory=params.get("data_access_factory", None),
+            transform_parameters=dict(params.get("transform_params", {})),
+        )
+        self.transform_params["statistics"] = params.get("statistics", None)
```
daw3rd: Should we be better about handling None?

data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+            return False
+        captured = CLIArgumentProvider.capture_parameters(args, cli_prefix, False)
+        # store parameters locally
+        self.multiprocessing = captured["multiprocessing"]
+        self.num_processors = captured["num_processors"]
+        # print them
+        if self.multiprocessing:
+            logger.info(f"multiprocessing {self.multiprocessing}")
+        else:
+            if self.num_processors <= 0:
```
daw3rd: This should also check for <= 1.  If they set to greatre than 1 but have multiprocessing=False, this is ambiguous and likely a user error of some sort.

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
daw3rd: i was not suggesting a formatting change :)

daw3rd (changes requested):

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_file_processor.py
blublinsky: We do

data-processing-lib/python/src/data_processing/runtime/pure_python/execution_configuration.py
blublinsky: fixed to 1. We are only checking num processes if the multiprocessing is specified

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_file_processor.py
daw3rd: Not sure I was clear.  If "statistics" is not provided, then self.stats is None, but then later it is used in _publish_stats() and this will cause an error.

daw3rd (changes requested):

blublinsky:
fixed

klwuibm:
@blublinsky Multiprocessing looks good. By the way, since we are modifying runtime, I wonder if we want to limit the print out of a floating number to have only 2 decimal digits, by using `round()`. 

For example, in `data-processing-lib/python/src/data_processing/runtime/pure_python/pooled_runtime_file_processor.py:`

Line 99: `f"Completed {completed} files ({100 * completed / len(files)}%) "` -> 
`f"Completed {completed} files ({round(100 * completed / len(files), 2)}%) "`

Line 100: `f"in {(time.time() - t_start)/60} min" `->
`f"in {round((time.time() - t_start)/60, 2} min"`

klwuibm:
Thanks, @blublinsky.  LGTM.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_file_processor.py
blublinsky: Its fixed

daw3rd: "statistics" is fixed, but there are lots of other places not checking for None.  But, that said, let's fix this as we go. I won't hold up this one over that

transforms/universal/ededup/ray/src/ededup_transform_ray.py
```diff
@@ -81,7 +81,7 @@ def _process_cached_hashes(self, hd: dict[str, str]) -> list[str]:
         return unique
 
 
-class EdedupRuntime(DefaultRayTransformRuntime):
```
daw3rd, line 84: Of course for backwards compatibility you should leave this class somehow

daw3rd (changes requested):
We need a test of the python runtime running parallel.

transforms/universal/ededup/ray/src/ededup_transform_ray.py
blublinsky: It is internal, so there is no compatibility issues

daw3rd (approved):
LG. Thanks for all the updates.

Issue #494, opened 2024-08-09
Work on the Table of the root README file 


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

Break the table up into universal, language, code, import/export tables.  Have links from the table to each transform directory/project.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #495, opened 2024-08-12
Enhancements to the Table of the root README file

shahrokhDaijavad:
Break the table up into universal, language, code, import/export tables. Have links from the table to each transform directory/project.

Issue #494

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,13 @@ Data modalities supported:
 Support for additional data modalities is expected in the future and additional data formats is welcome!
 
 ## Data Preparation Modules <a name = "modules"></a>
-The below matrix shows the the combination of modules and supported runtimes. All the modules can be accessed [here](/transforms) and can be combined to form data processing pipelines, as shown in [examples](/examples) folder. The modules are under three major categories of 1) Universal (apply to both code and language), 2) Language-only and 3) Code-only plus an additional category of Import/Export tables. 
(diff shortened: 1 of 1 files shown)


Bytes-Explorer:
Thanks @shahrokhDaijavad . I would suggest some minor changes. 

For universal, lets order by importance

My suggestion would be: exact dedup, fuzzy dedup, doc id annotation (lets call it unique id), and the rest can come in any order. 

For the language only part, lets rename [Split document to chunks] to Document Chunking for RAG, Chunk embeddings

In the import/export, why do we have a convert for PDF to parquet?

blublinsky (approved):

shahrokhDaijavad:
Thanks, @Bytes-Explorer. Done.

daw3rd (approved):

Bytes-Explorer:
@shahrokhDaijavad [Document Chunking for RAG, Chunk embeddings] are two separate modules. Chunk embeddings is what you have noted as text embeddings. So, we could have Document Chunking for RAG and Text embeddings. If somewhere we could make it clear that the text embeddings are used on the chunks, that will be great. 

 Also, may I suggest that we change repo level order to semantic file ordering. Can we add the link to this paper to the semantic file ordering module https://arxiv.org/pdf/2407.13739?

Bytes-Explorer:
@shahrokhDaijavad Pls raise a new PR to address these changes https://github.com/IBM/data-prep-kit/pull/495#issuecomment-2286580284

PR #496, opened 2024-08-12
Html2parquet changes made

sungeunan-ibm:
Updated html2parquet script to reflect changes mentioned in comments.


transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+# import data_processing
+
+
+from data_processing.transform import AbstractBinaryTransform, TransformConfiguration
+from data_processing.utils import CLIArgumentProvider, get_logger, TransformUtils
+
+
+class HtmlToParquetTransform(AbstractBinaryTransform):
+    def __init__(self, config: dict[str, Any]):
+        pass
```
blublinsky: this should be:
super().__init__(config)

transforms/universal/html2parquet/python/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+requires-python = ">=3.10"
+description = "HTML2PARQUET Python Transform"
+license = {text = "Apache-2.0"}
+readme = {file = "README.md", content-type = "text/markdown"}
+authors = [
+    { name = "Sungeun An", email = "sungeun.an@ibm.com" },
+    { name = "Syed Zawad", email = "szawad@ibm.com" },
+]
+dependencies = [
+    "data-prep-toolkit==0.2.1.dev0",
```
daw3rd, line 13: where is trafilatura dependency

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+        if TransformUtils.get_file_extension(file_name)[1] != ".zip":
+            self.logger.warning(f"Got unsupported file type {file_name}, skipping")
+            return [], {}
+        data = []
+        number_of_rows = 0
+
+        with zipfile.ZipFile(io.BytesIO(bytes(byte_array))) as opened_zip:
+            # Loop through each file member in the ZIP archive
+            for member in opened_zip.infolist():
+                if not member.is_dir() and '__MACOSX' not in member.filename:
```
daw3rd: don't you also want to make sure it is an html file?

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+                    with opened_zip.open(member) as file:
+                        try:
+                            # Read the content of the file
+                            content_bytes = file.read()
+                            # Use Trafilatura library
+                            content_string = trafilatura.extract(content_bytes)
```
daw3rd: You need to handle None return value, which is apparently an allowed/expected return.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+                            row_data = {
+                                "title": member.filename,
```
daw3rd: title and document are too vague (for me).  may title -> source_file and document -> basename.  And wonder why we need the basename separate from the full path.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self):
+        super().__init__(
+            name=short_name,
+            transform_class=HtmlToParquetTransform,
+        )
+    def add_input_params(self, parser: ArgumentParser) -> None:
+        pass 
+
+    def apply_input_params(self, args: Namespace) -> bool:
+        captured = CLIArgumentProvider.capture_parameters(args, cli_prefix, False)
```
daw3rd: Since you don't define any params in add_input_params() I think you just need to return True.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+#   RayTransformRuntimeConfiguration,
+# )
+# import data_processing
+
+
+from data_processing.transform import AbstractBinaryTransform, TransformConfiguration
+from data_processing.utils import CLIArgumentProvider, get_logger, TransformUtils
+
+
+class HtmlToParquetTransform(AbstractBinaryTransform):
```
daw3rd: For consistency, I would rename this and others from HtmlToParquet to Html2Parquet.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,84 @@
...
+        pass
+
+    def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
+        """
+        Converts raw data file (ZIP) to Parquet format
+        """
+        # We currently only process .zip files
+        if TransformUtils.get_file_extension(file_name)[1] != ".zip":
+            self.logger.warning(f"Got unsupported file type {file_name}, skipping")
+            return [], {}
```
daw3rd: This should throw an exception instead of returning the empty list.

daw3rd (changes requested):

transforms/universal/html2parquet/python/src/html2parquet_transform.py
sungeunan-ibm: The column names were from code2parquet. Would you recommend revising them? Additionally, the basename separation method was consistently used in both code2parquet and pdf2parquet. I think because they wanted to keep the file name rather than the full path?

blublinsky (approved):

shahrokhDaijavad:
@daw3rd Have all the changes you requested been implemented? 
@blublinsky Since you have approved, I assume all the changes you had requested are implemented.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
```diff
@@ -0,0 +1,117 @@
...
+            raise ValueError(error_message)  # Raising an exception with the error message
+        data = []
+        number_of_rows = 0
+
+        # Process ZIP archive of HTML documents
+        if(TransformUtils.get_file_extension(file_name)[1] == ".zip"):
+            with zipfile.ZipFile(io.BytesIO(bytes(byte_array))) as opened_zip:
+                # Loop through each file member in the ZIP archive
+                for member in opened_zip.infolist():
+                    if not member.is_dir() and '__MACOSX' not in member.filename and TransformUtils.get_file_extension(member.filename)[1] == ".html":
```
daw3rd: why would you need to MACOSX test if you're always requiring .html extension?

transforms/universal/html2parquet/python/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Output format
+
+The output format will contain the following colums
+
+```jsonc
+{
+	"title": "string"             // the member filename
+	"document": "string"          // the base of the source archive
+	"contents": "string"          // the content of the HTML
+    "document_id": "string",      // the document id, a random uuid4
```
daw3rd: document_id is not a randome uuid. it is a hash of the contents.

daw3rd (changes requested):
Please also add this transform to the import/export Table on the repo's top-level README.

transforms/universal/html2parquet/python/src/html2parquet_transform.py
sungeunan-ibm: You are right. This MACOSX test was added before requiring .html extension. Deleted it.

shahrokhDaijavad:
@sungeunan-ibm The root README file has changed in the dev branch, compared to the one you are using. That is the source of the conflict.

shahrokhDaijavad:
Thank you, @sungeunan-ibm . I think this PR is ready to be approved and merged, @daw3rd.

daw3rd (approved):
Thanks for the changes!

PR #499, opened 2024-08-13
fix lib doc .py links and update resize readme

daw3rd:
Fix links in library markdown.


shahrokhDaijavad (approved):
LGTM.

PR #500, opened 2024-08-14
fix lang_id, use parameters related to output columns

dtsuzuku-ibm:

There is a bug that output column names cannot be updated as specified in cli parameters


transforms/language/lang_id/python/README.md
```diff
@@ -16,6 +16,8 @@ configuration for values are as follows:
 * _lang_id_model_kind_ - specifies what kind of model you want to use for language identification. Currently, only `fasttext` is available.
 * _lang_id_model_url_ - specifies url that model locates. For fasttext, this will be repo name of the model, like `facebook/fasttext-language-identification`
 * _lang_id_content_column_name_ - specifies name of the column containing documents
+* _lang_id_output_lang_column_name_ - specifies name of the output column to hold predicted language code
```
daw3rd: as configuration keys, thes should all NOT have the `lang_id_` prefix, according to your  code anyway.

daw3rd: Also,  the defaults as implemented in `_init__()` should be documented.

daw3rd (changes requested):

daw3rd (approved):

PR #504, opened 2024-08-14
Update docling version and tests

dolfim-ibm:

1. Updates Docling to the latest version
2. Truncate bbox in the results, because of floating point differences between CI runs


transforms/language/pdf2parquet/python/src/pdf2parquet_local_python.py
```diff
@@ -36,9 +36,9 @@
     "runtime_job_id": "job_id",
     "runtime_code_location": ParamsUtils.convert_to_ast(code_location),
     # pdf2parquet params
-    # "pdf2parquet_do_table_structure": False,
-    # "pdf2parquet_contents_type": "application/json",
-    # "pdf2parquet_contents_type": "text/markdown",
+    #"pdf2parquet_do_table_structure": False,
+    #"pdf2parquet_contents_type": "application/json",
+    #"pdf2parquet_contents_type": "text/markdown",
 }
```
blublinsky, line 42: DO we need this commented-out stuff?

blublinsky (approved):
LGTM with 1 question

transforms/language/pdf2parquet/python/src/pdf2parquet_local_python.py
dolfim-ibm: it is actually there to simplify running the transform with some common options, which are also tested individually in the tests

transforms/language/pdf2parquet/python/src/pdf2parquet_transform.py
```diff
@@ -114,7 +112,7 @@ def _convert_pdf2parquet(self, doc_filename: str, ext: str, content_bytes: bytes
         if self.contents_type == pdf2parquet_contents_types.MARKDOWN:
             content_string = doc.render_as_markdown()
         elif self.contents_type == pdf2parquet_contents_types.JSON:
-            content_string = pd.io.json.ujson_dumps(doc.render_as_dict(), double_precision=2)
+            content_string = pd.io.json.ujson_dumps(doc.render_as_dict(), double_precision=0)
```
daw3rd, line 115: are these scores less than 1 that are being truncated to 0 precision after the decimal?  Do you still need this?

dolfim-ibm: The JSON format contains bbox. It seems that the precision on the CI workers is very unpredictable, so this is rounding them to integer.

PR #509, opened 2024-08-16
doc id python

blublinsky:
Adding a python version of doc_id.


transforms/universal/doc_id/python/src/doc_id_transform_python.py
```diff
@@ -0,0 +1,71 @@
...
+    DefaultRayTransformRuntime,
+    RayTransformLauncher,
+)
+from data_processing_ray.runtime.ray.runtime_configuration import (
+    RayTransformRuntimeConfiguration,
+)
+from doc_id_transform_base import IDGenerator, DocIDTransformBase, DocIDTransformConfigurationBase
+
+
+class DocIDPythonTransform(DocIDTransformBase):
```
daw3rd: None of our pure python transform classes contain `Python`.  For consistency, can you rename w/o `Python`?

transforms/universal/doc_id/python/src/doc_id_transform_python.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_starting_id(self, n_rows: int) -> int:
+        """
+        Get starting Id
+        :param n_rows - number of rows in the table
+        :return: starting id for the table
+        """
+        self.id_generator.get_ids(n_rows=n_rows)
+
+
+class DocIDPythonTransformConfiguration(TransformConfiguration):
```
daw3rd: None of our pure python transform classes contain `Python`.  For consistency, can you rename w/o `Python`?

transforms/universal/doc_id/python/src/doc_id_transform_python.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def __init__(self, config: dict[str, Any]):
+        """
+        Initialize based on the dictionary of configuration information.
+        """
+        # Make sure that the param name corresponds to the name used in apply_input_params method
+        super().__init__(config)
+
+    def get_starting_id(self, n_rows: int) -> int:
```
daw3rd: Maybe this should be `_get_starting_id` if only called from within this file or the super class?

daw3rd (changes requested):

transforms/universal/doc_id/python/src/doc_id_transform_python.py
blublinsky: Yes, because they are not derived from base. Only this one and ededup do

transforms/universal/doc_id/python/src/doc_id_transform_python.py
blublinsky: Yea, but its overwritten from base

transforms/universal/doc_id/python/src/doc_id_transform_python.py
daw3rd: By convention the python transforms run in the python runtime do not include `Python` in the name.  Consistency is key when we have so many transforms to manage.  Just because there is a common base class does not mean we need to violate the convention.

transforms/universal/doc_id/python/src/doc_id_transform_python.py
daw3rd: As I said, the super class.  If this method is only used within the class hiearachy (super or sub) it should be hidden from general use.

blublinsky: fixed

transforms/universal/doc_id/python/src/doc_id_transform_python.py
blublinsky: fixed

transforms/universal/doc_id/python/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+publish: publish-image
+
+publish-image:: .transforms.publish-image-python
+
+setup:: .transforms.setup
+
+# distribution versions is the same as image version.
+set-versions:
+	$(MAKE) TRANSFORM_PYTHON_VERSION=$(EDEDUP_PYTHON_VERSION) TOML_VERSION=$(EDEDUP_PYTHON_VERSION) .transforms.set-versions
```
daw3rd: Doesn't have an effect, but for strict correctness EDEDUP -> DOC_ID

transforms/universal/doc_id/python/test/test_doc_id_python.py
```diff
@@ -0,0 +1,43 @@
...
+)
+from doc_id_transform_python import DocIDPythonTransformRuntimeConfiguration
+from doc_id_transform_base import (doc_column_name_cli_param,
+                                   hash_column_name_cli_param,
+                                   int_column_name_cli_param,
+                                   start_id_cli_param,
+                                   )
+
+
+class TestPythonEdedupTransform(AbstractTransformLauncherTest):
```
daw3rd: Ededup

daw3rd (changes requested):

transforms/universal/doc_id/python/Makefile
blublinsky: Fixed

transforms/universal/doc_id/python/test/test_doc_id_python.py
blublinsky: Fixed

daw3rd (approved):

PR #511, opened 2024-08-18
Fix to try and avoid running out of disk space during ci/cd on transforms tests.

daw3rd:
Change .make.transforms so that the "test" target also does a clean after testing.

To try and avoid running out of disk while testing the transforms during ci/cd.


SowmyaLR (approved):
Thank you @daw3rd  for this clean up task. Looks Good to me.

daw3rd:
This fix was not an actual fix.  It applied to the `test` target but needed to apply to the `test-src` target. That is a bit more involved and will be fixed in a later PR.

PR #518, opened 2024-08-19
PII Transform: restablish test make file

touma-I:
## Why are these changes needed?

Re-create Makefile for full suite testing for pii transform

## Related issue number (if any).

#517 
#516

PR #522, opened 2024-08-20
Fix parameters type for pii transform pipeline.

Mohammad-nassar10:
The type of some of the input parameters of the pipelines was changed from `str` to `dict` but missing applying this change to the `pii` transform and the auto generator.

It should fix the errors of https://github.com/IBM/data-prep-kit/pull/518


daw3rd (commented):
LGTM

roytman (approved):
LGTM

Issue #513, opened 2024-08-19
[Bug] html2parquet needs a Makefile to enable tests


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### What happened + What you expected to happen

Currently the transforms/universal/html2parquet directory is missing the standard Makefile.  As a result, tests are not being run for this transform - an oversight on the reviewer (me?).  In addition, `make test-src` seems to be failing in the python directory, but that will show up in the PR once you add the Makefile.

### Reproduction script

```
$ cd transforms/universal/html2parquet/
$ make -n test
make: *** No rule to make target `test'.  Stop.
```

### OS

MacOS (limited support)

### Python

3.11.x

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

sungeunan-ibm:
I will work on this issue.

daw3rd:
There is now a Makefile and test-src and test-image pass (tested on rh)

PR #524, opened 2024-08-20
Html2parquet Makefile added

sungeunan-ibm:
Added a Makefile to enable tests

[Bug] html2parquet needs a Makefile to enable tests #513
https://github.com/IBM/data-prep-kit/issues/513


shahrokhDaijavad:
Thank you, @sungeunan-ibm for doing this so quickly! Please resolve the conflicts.

daw3rd:
@sungeunan-ibm please resolve the conflicts and make sure the ci/cd tests are running after that.  Thanks!

touma-I:
@sungeunan-ibm It will be easier for me to fix the CI/CD issues that this PR is running into and let you focus on the ray version of the transform. If you can confirm that all the Unit Tests for your python transform are passing, then I will take it from here to fix the remaining issues with this PR. I will tag you if needed. Please create a new PR for the Ray implementation and keep it in draft mode until all the Unit Test for the Ray code has passed. Thanks for all your help. This is great work!
cc: @daw3rd @shahrokhDaijavad

transforms/universal/html2parquet/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+docker-load-image::
+	@# Help: Recursively make $@ in all subdirs
+	$(MAKE) RULE=$@ .recurse
+
+docker-save-image::
+	@# Help: Recursively make $@ in all subdirs 
+	$(MAKE) RULE=$@ .recurse
+
+.PHONY: workflow-venv
```
roytman, line 56: You don't have KFP automation for the transform, so please leave the following Make rules empty.
```make
.PHONY: workflow-venv
workflow-venv:

PHONY: workflow-test
workflow-test:

.PHONY: workflow-upload
workflow-upload:

.PHONY: workflow-build
workflow-build:
```

This will fix the KFP CICD failures

daw3rd:
@sungeunan-ibm please sign your commits using `git commit -s ...` Note that git desktop does not seem to support signing.

daw3rd:
@sungeunan-ibm For now let's ignore the kfp test failures and focus on the html2parquet test failures.  You can run `make test-src` and `make test-image`  in the python directory to run your tests.

sungeunan-ibm:
The unit test runs with make test-src, but there’s an issue with the `date_acquired` column, which uses `datetime.now().isoformat()`. This causes timestamp changes with each test run, which is causing mismatches with the expected output. Should we remove the date_acquired column, or is there a way to run the test without it?

shahrokhDaijavad:
Thanks, @sungeunan-ibm 
@daw3rd Do we care about the timestamps not matching? Do other transforms behave the same way?

touma-I:
> The unit test runs with make test-src, but there’s an issue with the `date_acquired` column, which uses `datetime.now().isoformat()`. This causes timestamp changes with each test run, which is causing mismatches with the expected output. Should we remove the date_acquired column, or is there a way to run the test without it?

Thanks @sungeunan-ibm I will check to see if the framework allow us to exclude the time column from the validation step. I think we had a similar requirement from another transform and seems like a reasonable thing for the framework to do. I am not if @daw3rd had a chance to address that but I will investigate.

shahrokhDaijavad:
@sungeunan-ibm As you can see, @touma-I has now added issue #559 as a priority. This would take precedence over the Ray-scalable version of this transform. 
If you can resolve the conflict shown, I think we can live with the timestamp issue. We can then merge this branch and then you can look into adding markdown as the default option for what will be in the content column of the parquet file, with text being the non-default option (assuming Trafilatura gives this).

shahrokhDaijavad:
Still shows a conflict. @daw3rd Can you please help? It seems easy enough, but I don't want to make a mistake with the .make.versions file.

daw3rd:
@sungeunan-ibm it looks like maybe you need to update the test-data/expected test data to match the new date-acquired column.  The launcher-based tests can be made to ignore selected columns (see latest noop tests).  However,
the non-launcher pure transform tests do not allow this per issue #564.  If you're running into that, you may need to disable the non-transform tests - the launcher tests are a super set really so should be sufficient for now.

sungeunan-ibm:
@daw3rd I agree, and I want to disable the non-transform tests. Is this the correct way to do it? Is this how you would disable it in the Makefile` #test-src:: .transforms.test-src`?

daw3rd:
> @daw3rd I agree, and I want to disable the non-transform tests. Is this the correct way to do it? Is this how you would disable it in the Makefile` #test-src:: .transforms.test-src`?
@sungeunan-ibm 

No, just delete the test/test_html2parquet.py. We don't want to turn off all tests, just the non-launcher test.

See https://github.com/IBM/data-prep-kit/blob/dev/transforms/language/pdf2parquet/python/test/test_pdf2parquet_python.py for an example of ignoring columns in the launcher-based test.

touma-I (approved):
Looks good for this iteration. Thanks

Issue #531, opened 2024-08-23
[Bug] Remove Test from published library release to pypi


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core

### What happened + What you expected to happen

The wheel currently deployed to Pypi includes test folder

### Reproduction script

pip install data-prep-toolkit=0.2.1.dev0

### OS

MacOS (limited support)

### Python

3.11.x

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

touma-I:
Solution is to modify pypropject.toml to exclude test folder from package_dir

[options]
package_dir = ["src"]

touma-I:
Also modify .make.versions and publish distribution to pypi
```
make set-versions
make build-dist
make publish-dist

```

daw3rd:
I have looked at the wheel (zip file) and there is no test or test-data.  The tar.gz does contain this, but that file contains everything from the project directory.

daw3rd:
closing

PR #532, opened 2024-08-23
modify package_dir and use dev2 release for testing

touma-I:

Publish new wheel to pypi with src only (No Test code)

#531


touma-I:
superseded by #540

PR #539, opened 2024-08-24
Don't use pip cache in Dockerfile and fix exception reporting msg

daw3rd:
This should make python images smaller, but more importantly MAY fix `PACKAGES DO NOT MATCH THE HASHES` errors we're seeing in some image builds during ci/cd.


touma-I (approved):
Please merge it if you are done so I can apply it to PII redactor.

PR #528, opened 2024-08-22
pdf2parquet updates

dolfim-ibm:
## Why are these changes needed?

In this PR we have updates for the `pdf2parquet` transform.

- Update to the new Docling version
- Leverage the lazy OCR feature
- Simplify the test cases (also use the list of fixtures), which fixes https://github.com/IBM/data-prep-kit/issues/459 and https://github.com/IBM/data-prep-kit/issues/552

TODO (for a next PR)
- expose extraction of pages with their images
- expose more Docling parameters
- generalize the interface to support extensions of the transform

## Related issue number (if any).

- https://github.com/IBM/data-prep-kit/issues/459
- https://github.com/IBM/data-prep-kit/issues/552

PR #545, opened 2024-08-26
Added instructions for using pip install

touma-I:

Enhance quick-start documentation to explain how to use pip install for the data-prep-kit and transforms.


Bytes-Explorer:
Thanks for this PR. We should also add commands for user to create an environment and then use dpk in it. Something like below

conda create -n data-prep-kit-1 -y python=3.11

activate the new conda environment
conda activate data-prep-kit-1
make sure env is swithced to data-prep-kit-1

Check python version
python --version
should say : 3.11

install jupyter lab
pip3 install jupyterlab

In this environment dpk can be installed via pip. 

This will make sure user is working on py3.11 and  avoid py3.11 conflicts with any existing py version.

Bytes-Explorer:
Also as part of this PR, can we remove the additional NB https://github.com/IBM/data-prep-kit/blob/dev/examples/notebooks/code/sample-notebook_llama.ipynb which was meant for addition to the llama repo

doc/quick-start/quick-start.md
```diff
@@ -1,5 +1,39 @@
 # Quick Start for Data Prep Kit 
-Here we provided short examples of various uses of the Data Prep Kit.
+Here we provided short examples of various uses of the Data Prep Kit. Most users who want to jump right in can use standard pip install to deploy the data-prep-kit and the python or ray transforms to their virtual python environment.
```
daw3rd, line 2: Can we have a create-venv.md that breaks this out and then has separate sections for conda and venv.  It would also be good to have a link to install conda.  I believe venv comes with the base python install so no link needed there.

```
python -m venv venv
source venv/bin/activate
pip install ...
...
deactivate
```

daw3rd (changes requested):

daw3rd:
Also, we don't yet have the transforms/packaging directory so probably should not be telling folks how to install `data-prep-toolkit-transforms` until we have mainline support for that.  @touma-I

doc/quick-start/quick-start.md
```diff
@@ -1,5 +1,39 @@
...
+        `pip3 install jupyterlab`   
+<br><br>
+
+- Deploy the latest release of the data prep toolkit library<br>
+    `pip3 install data-prep-toolkit`
+
+or 
+-  deploy the latest releases of the data prep toolkit library and all python transforms
+
+    `pip3 install data-prep-toolkit-transforms`
```
daw3rd, line 43: I made this comment elsewhere, but just to locate it, this is not part of the `dev` branch  yet.  That is, the `transforms/packaging` directory needs to be present before we start referring to this wheel.

doc/quick-start/quick-start.md
```diff
@@ -1,5 +1,39 @@
...
+
+or 
+-  deploy the latest releases of the data prep toolkit library, all python transforms and all ray transforms
+
+    `pip3 install data-prep-toolkit-transforms-ray`
```
daw3rd, line 48: ditto

daw3rd (dismissed):

doc/quick-start/quick-start.md
shivdeep-singh-ibm: This step also requires gcc/g++, while building `fasttext` which is pulled by it.

On conda environemnts, it needs to be installed like:
```
conda install gcc_linux-64
conda install gxx_linux-64
```

touma-I: @shivdeep-singh-ibm do you know if this is specific to Ubuntu? I don't recall having to do that on my MacOS but I also did not use conda.

doc/quick-start/quick-start.md
touma-I: @daw3rd Let's have that as a separate issue/PR to revisit the documentation. As a new comer, and please I don't want anyone to take this the wrong way as I know a lot of effort and thoughts went into it, but all those different MD we have are quite confusing to navigate and increase difficult for maintaining it all. I don't know if adding one more will help or hurt. Maybe neither but there is a point to be made about producing the right amount of documentation that meets most users need at this point and iterate **as our users** (emphasis on users) ask for it.

doc/quick-start/quick-start.md
sujee: I am using Ubuntu 24.04 that has GCC v13 and can confirm it happens.  Seems like it compiles with GCC v11.  I have a fix here  #550

daw3rd:
To fix the language failures I think you will need to merge from dev, which now includes #528.

doc/quick-start/quick-start.md
sujee: I just did this and tested on both conda and venv environments,

https://github.com/sujee/data-prep-kit/blob/rag-example1/examples/notebooks/rag/setup-python-dev-env.md

happy to submit a PR for dev-env setup, if needed.

daw3rd: @touma-I Yep this is a style thing.  I personally do not like navigating through a long document to find the content I want and find it much easier to have a page full of links to focused content.

shahrokhDaijavad (approved):
I went ahead and tested noop transform successfully after following the instructions in quick-start.md file.

doc/quick-start/quick-start.md
touma-I: That should be ok for now. the packaging PR still need some more documentation but let's not hold back this PR for it.

PR #546, opened 2024-08-27
use .env file in running docker container if the file exists

dtsuzuku-ibm:

To pass some parameter (like `DPK_HUGGING_FACE_TOKEN`) to test runner kicked in docker, we need to pass it via environment variable.

### DIFF
--- a/.make.defaults
+++ b/.make.defaults
@@ -405,8 +405,12 @@ endif
 .PHONY: .defaults.test-image-pytest
 .defaults.test-image-pytest:: 
 	# Put this 2nd so its help showss up instead of .defaults.image help
-	@# Help: Test $(DOCKER_LOCAL_IMAGE) using test source inside the image. 
-	$(DOCKER) run -t --rm $(DOCKER_LOCAL_IMAGE) pytest -s test 
+	@# Help: Test $(DOCKER_LOCAL_IMAGE) using test source inside the image.
+	@if [ -f "$(REPOROOT)/.env" ];	then	\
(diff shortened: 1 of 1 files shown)


.make.defaults
```diff
@@ -405,8 +405,12 @@ endif
 .PHONY: .defaults.test-image-pytest
 .defaults.test-image-pytest:: 
 	# Put this 2nd so its help showss up instead of .defaults.image help
-	@# Help: Test $(DOCKER_LOCAL_IMAGE) using test source inside the image. 
-	$(DOCKER) run -t --rm $(DOCKER_LOCAL_IMAGE) pytest -s test 
+	@# Help: Test $(DOCKER_LOCAL_IMAGE) using test source inside the image.
+	@if [ -f "$(REPOROOT)/.env" ];	then	\
+	    $(DOCKER) run -t --env-file=$(REPOROOT)/.env --rm $(DOCKER_LOCAL_IMAGE) pytest -s test;	\
```
daw3rd, line 410: But this repo does not contain a .env at the top.  What do you want to put it it and what problem does it solve?

daw3rd (changes requested):

.make.defaults
dtsuzuku-ibm: I want the way to set environment variables to the container started by make command for pytest.

dtsuzuku-ibm: > But this repo does not contain a .env at the top

It doesn't matter. I want the way that can be used in local development or in CI/CD.

Issue #552, opened 2024-08-28
[Bug] pdf2parquet is now failing ci/cd builds 


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### What happened + What you expected to happen

Ci/CD builds are now failing for pdf2parquet in at least to unrelated PRs and I can reproduce the failure locally on my mac m1.  

1. https://github.com/IBM/data-prep-kit/actions/runs/10594844975/job/29359345957?pr=545
2. https://github.com/IBM/data-prep-kit/actions/runs/10599667381/job/29378015278?pr=548

### Reproduction script

```
cd transforms/language/pdf2parquet/python
make test-src
```

### Anything else

E                         [21] Peng Zhang, Can Li, Liang Qiao, Zhanzhan Cheng, Shiliang Pu, Yi Niu, and Fei Wu. Vsr: A unified framework for document layout analysis combining vision, semantics and relations, 2021.
E                         
E                         [22] Peter W J Staar, Michele Dolfi, Christoph Auer, and Costas Bekas. Corpus conversion service: A machine learning platform to ingest documents at scale. In Proceedings of the 24th ACM SIGKDD International Conference on Knowledge Discovery and Data Mining , KDD, pages 774-782. ACM, 2018.
E                         
E                         [23] Connor Shorten and Taghi M. Khoshgoftaar. A survey on image data augmentation for deep learning. Journal of Big Data , 6(1):60, 2019."]]
E                         num_pages: [[9]]
E                         num_tables: [[5]]
E                         num_doc_elements: [[147]]
E                         ext: [["pdf"]]
E                         hash: [["313bb7ef50bea94a1ef5ae4417f45923cb4ac383d49ba781c874eb9bfbc06be0"]]
E                         size: [[41244]]
E                         source_filename: [["2206.01062.pdf"]]
E                       assert <pyarrow.lib....\n    148\n  ]\n] == <pyarrow.lib....\n    147\n  ]\n]
E                         
E                         Use -v to get more diff

../../../../../data-processing-lib/python/src/data_processing/test_support/abstract_test.py:135: AssertionError
------------------------------------------------------------------------- Captured log call -------------------------------------------------------------------------
[...]

PR #553, opened 2024-08-28
quarantine/disable testing of pdf2parquet per issue #552

daw3rd:
Because this failure is disabling all other PRs.

#552


dolfim-ibm:
This should not be needed. Please review https://github.com/IBM/data-prep-kit/pull/528

daw3rd:
This does not seem needed now that we have PR #528 merged which fixes the tests.

PR #555, opened 2024-08-29
Improve noop test templates for reuse.

daw3rd:
update noop tests to use standardized way of loading test data for transforms from files and use empty ignore columns list.

So new transform authors have to do less when writing non-launcher tests and to show how to use ignore columns facility.


shivdeep-singh-ibm (approved):
lgtm

PR #558, opened 2024-08-30
Use cpu-only version of torch on linux builds

dolfim-ibm:

These transforms depend on the [PyTorch](https://pytorch.org/) library. Depending on your architecture, you might want to use a different distribution of torch. For example, you might want support for different accelerator or for a cpu-only version. All the different ways for installing torch are listed on their website https://pytorch.org/.

With this PR we switch to the simpler cpu-only version of `torch` for the docker images and the CI. Users can still get the full torch package from pypi, simply setting `LINUX_WITH_CPU_TORCH=false`.

### DIFF
--- a/.make.defaults
+++ b/.make.defaults
@@ -39,9 +39,9 @@ DOCKER_FILE?=Dockerfile
 DOCKER?=docker
 DOCKER_PLATFORM?=linux/amd64
 # Can be used by transforms or others to add args to the "docker build" command in .defaults.image target
-DOCKER_BUILD_EXTRA_ARGS=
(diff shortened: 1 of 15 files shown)


dolfim-ibm:
Note that while working on this fix I noticed the `.make.defaults` was missing the optional default. It showed as a difference in python vs ray versions, since the former was including the `.make.transforms` after defining the variable.

As discussed with @daw3rd, we changed the noop example to also flip the include order.

daw3rd (approved):

daw3rd (approved):

daw3rd (approved):

PR #563, opened 2024-09-03
Organise examples by use cases 

Bytes-Explorer:
Organise getting started examples by use cases. Have moved some duplicate files to archive for now. We can bring them back post discussion with the team


shivdeep-singh-ibm:
moving notebooks to internal nested folders will break `examples/notebooks/code_developers`,
`Makefile` and `prepare_env.sh`, also need to be updated.

Bytes-Explorer:
We are upgrading to the new notebook sample-notebook.ipynb and want everyone to use pip install. We are archiving the other notebooks for now and will eventually make them compatible with pip based installation or remove them.

shivdeep-singh-ibm (approved):

daw3rd (changes requested):
Why keep the archive directory?

Bytes-Explorer:
> Why keep the archive directory?

David, we will remove it once we upgrade all notebooks in a couple of days.

daw3rd (approved):
Prefer not to have leftover `archive` but in the interest of time/sharing...

PR #577, opened 2024-09-06
Custom column validator for pdf2parquet

dolfim-ibm:

With this PR we should resolve the instability of the `pdf2parquet` tests. Lately failing on tiny differences in the bbox predictions (452 vs 451).

To enable a custom validator for the pdf2parquet transform, we turn the `@staticmethod` to `@classmethod`, which allows for easier overloading.

Possible extension (in the future): allow transforms test to define custom validator for specific columns.


blublinsky (approved):
LGTM

daw3rd (approved):

PR #584, opened 2024-09-10
update docling dependencies to newer versions

dolfim-ibm:

Old versions of Docling were pinning third-party libraries with a large version range. This was fixed already a few days ago and we have to propagate here as well.


dolfim-ibm:
From running locally:

```
PYTHON=python3.10 make test-src
...
========================================================== 3 passed, 1 skipped, 6 warnings in 25.64s ===========================================================
...
```

daw3rd (approved):
lgtm

Issue #515, opened 2024-08-19
[Feature] Develop ability to run ci/cd testing only on the portion of repo that has changed and its dependencies


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

WOuld be nice to only run tests on those components impacted by the change.  For example,

* Readme changes don't do any tests
* core lib changes might have to run everything
* transform X change only runs tests on transform X
* etc.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

touma-I:
From Matsubara-san: " found an action to detect changed files https://github.com/marketplace/actions/changed-files
Using this you can branch out the execution of the test cases. If some of them are specified as PR check, then you can run a no-op test script instead to pass the check."

PR #589, opened 2024-09-13
disable test workflow when no code files change

daw3rd:
We would like to speed up the ci/cd process where possible.  
This disables retesting when non-code (e.g., *.md, examples/*, etc) change.

Issue 515 https://github.com/IBM/data-prep-kit/issues/515

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,6 @@ on:
             - "**.md"
             - "examples/**"
             - "**/doc/**"
-            - "data-processing-lib/doc/**"
             - "**/.gitignore"
             - "**/.dockerignore"
     pull_request:
@@ -23,7 +22,6 @@ on:
             - "**.md"
             - "examples/**"
             - "**/doc/**"
-            - "data-processing-lib/doc/**"
             - "**/.gitignore"
             - "**/.dockerignore"
 env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ on:
         paths-ignore:
             - "**.md"
             - "examples/**"
-            - "doc/**"
+            - "**/doc/**"
             - "data-processing-lib/doc/**"
             - "**/.gitignore"
             - "**/.dockerignore"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -8,12 +8,26 @@ on:
...
             - "releases/**"
+        paths-ignore:
+            - "**.md"
+            - "examples/**"
+            - "doc/**"
+            - "data-processing-lib/doc/**"
+            - "**/.gitignore"
+            - "**/.dockerignore"
 env:
-  KFP_BLACK_LIST: "doc_chunk-ray,pdf2parquet-ray,pii_redactor"
```
daw3rd, line 16: This seems to be a result of pre-commit.

shivdeep-singh-ibm (approved):
lgtm

.github/workflows/test.yml
```diff
@@ -8,12 +8,26 @@ on:
             - "releases/**"
         tags:
             - "*"
+        paths-ignore:
+            - "**.md"
+            - "examples/**"
+            - "**/doc/**"
+            - "data-processing-lib/doc/**"
```
roytman: could you please add `kfp/doc/**`

.github/workflows/test.yml
```diff
@@ -8,12 +8,26 @@ on:
...
+            - "**/.dockerignore"
     pull_request:
         branches:
             - "dev"
             - "releases/**"
+        paths-ignore:
+            - "**.md"
+            - "examples/**"
+            - "doc/**"
+            - "data-processing-lib/doc/**"
```
roytman: as above

roytman (commented):
good idea

revit13 (approved):

roytman (changes requested):
to prevent accidental merge

.github/workflows/test.yml
daw3rd: **/doc/** covers this i believe

.github/workflows/test.yml
daw3rd: I changed to `**/doc/**` should catch kfp too now I believe. Thanks.

.github/workflows/test.yml
roytman: if this is right, why do you need `data-processing-lib/doc/**` ?

roytman (approved):

PR #592, opened 2024-09-16
refactoring of data access code

blublinsky:
consolidated s3 and local code
fixed small issues in checkpointing


daw3rd (commented):
lgtm but would like another set of eyeballs

daw3rd (approved):

PR #595, opened 2024-09-17
Start triggering testing at finer granularity in the repo

daw3rd:
This PR splits out the tests from test.yml into separate tests for each transform and kfp and the core libraries.
Each workflow is now targeted to a subdirectory of the repository so fewer full builds should happen.
The workflows directory contains a Makefile and test-transforms.template file that is used to generate
the `test-*-*.yml` workflows for each of the transforms in the repository.

When adding a new transform, the Makefile will need to be edited to include the new transform and make will need to be run to create the transform's workflow.  

Also, this should help the space problems as each workflow only handles  1 transform.

Also :), this should make the build (of transform changes) faster since only the changed transform is processed.

To speed up the ci/cd builds.

https://github.com/IBM/data-prep-kit/issues/515


.github/workflows/README.md
```diff
@@ -0,0 +1,27 @@
...
+should be sufficient.
+
+## Transforms 
+We define a unique test workflow for each transform, based on a common template [test-transform.template](test-transform.template).
+The Makefile is used to (re)generate all workflows a necessary.  By design, workflows for a given transform should run when
+
+* anything of substance effecting operation is modified in the transform's directory tree.
+* anything in the core libraries in this repo (e.g., data-processing/lib) assuming the transform depends on these.
+* Help! the workflow should NOT run when documentation (e.g., !**.md) is changed, however this case does not seem to be working atm.
+
```
revit13, line 41: It might be worth mentioning that the kfp workflow of the transform is not tested in the unique test workflow.

.github/workflows/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+make
+git commit -a -s -m "update workflows"
+git push
+```
+
+should be sufficient.
+
+## Transforms 
+We define a unique test workflow for each transform, based on a common template [test-transform.template](test-transform.template).
```
revit13: Just for the testing please consider adding a small change to the data-prep-lib code in this PR in order to see to all the runs in parallel...

.github/workflows/test-kfp.yml
```diff
@@ -0,0 +1,174 @@
...
+            - "kfp/**"
+            - "!**.md"
+            - "!**/doc/**"
+            - "!**/.gitignore"
+    pull_request:
+        branches:
+            - "dev"
+            - "releases/**"
+        paths:
+            - "kfp/**"
```
revit13, line 22: kfp workflow test should actually be triggered every time a change is made in a transform, its dependent libs or kfp components... for example, to check that a change in transform parameters is ok so I would not trigger it only in `kfp` directory change. It is currently choosing the transform to run randomly and its not ideal, it might be worth checking if the kfp test for a specific transform could be integrated in each of the unique transform workflow.

.github/workflows/test-code-code2parquet.yml
```diff
@@ -0,0 +1,95 @@
...
+    pull_request:
+        branches:
+            - "dev"
+            - "releases/**"
+        paths:
+            - "transforms/code/code2parquet/**"
+            - "data-processing-lib/**"
+
+jobs:
+    check_if_push_image:
```
revit13, line 40: typo check_if_push_image-> check_if_push_image**s** (occurs in all the workflows)

.github/workflows/test-code-code2parquet.yml
```diff
@@ -0,0 +1,95 @@
...
+on:
+    workflow_dispatch:
+    push:
+        branches:
+            - "dev"
+            - "releases/**"
+        tags:
+            - "*"
+        paths:
+            - "transforms/code/code2parquet/**"
```
revit13, line 15: not sure about that... in a release the image of this transform will not be pushed if there was no change in it?

transforms/universal/filter/python/src/filter_local.py
```diff
@@ -55,4 +55,4 @@
     # Transform the table
     table_list, metadata = transform.transform(table)
     print(f"\noutput table has {table_list[0].num_rows} rows")
-    print(f"output metadata : {metadata}")
+    print(f"output metadata :: {metadata}")
```
roytman: sorry, for my curiosity, why do we need the second colon?

transforms/universal/noop/python/src/noop_transform.py
```diff
@@ -114,5 +115,5 @@ def apply_input_params(self, args: Namespace) -> bool:
             return False
 
         self.params = self.params | captured
-        self.logger.info(f"noop parameters are : {self.params}")
+        self.logger.info(f"noop parameters are::: {self.params}")
```
roytman: and now 3 columns?

.github/workflows/test.yml
roytman, line 1: can you rename the file extension and leave it for a while? it can be useful as a reference.

.github/workflows/Makefile
```diff
@@ -0,0 +1,20 @@
+
+# Directories in the transforms/universal directory for which we want to generate test workflows
+UNIVERSAL_TRANSFORMS=doc_id ededup fdedup filter html2parquet noop profiler resize tokenization
+# Directories in the transforms/code directory for which we want to generate test workflows
+CODE_TRANSFORMS=code2parquet code_quality header_cleanser malware proglang_select repo_level_ordering
+# Directories in the transforms/language directory for which we want to generate test workflows
+LANG_TRANSFORMS=doc_chunk doc_quality lang_id pdf2parquet pii_redactor text_encoder
```
roytman, line 7: hardcoded list of directories can be unscalable, can it be replaced by automatic list generation, as we did in the random KFP tests

.github/workflows/README.md
```diff
@@ -0,0 +1,27 @@
+# Workflow Management
+
+Here we have the start of a system to automatically generated workflows (currently only for transforms). 
+In general, the design is to use templates and `make` to generate/update the workflows.
+
+```
+make
+git commit -a -s -m "update workflows"
+git push
```
roytman: if the git target is not set, the command will fail.

.github/workflows/test-code-code2parquet.yml
daw3rd: i suppose this begs the need for a separate workflow for publishing and to not include publishing with testing.

.github/workflows/README.md
```diff
@@ -0,0 +1,27 @@
...
+The Makefile is used to (re)generate all workflows a necessary.  By design, workflows for a given transform should run when
+
+* anything of substance effecting operation is modified in the transform's directory tree.
+* anything in the core libraries in this repo (e.g., data-processing/lib) assuming the transform depends on these.
+* Help! the workflow should NOT run when documentation (e.g., !**.md) is changed, however this case does not seem to be working atm.
+
+When a new transform is added to the repository, 
+
+1. add it to the corresponding macro in the Makefile (in alphabetical order please). 
+1. Run `make` to create the new test .yml for the transform
```
roytman: can we add an automatic validation that it was done?

.github/workflows/build-library.yml
```diff
@@ -6,10 +6,20 @@ on:
         branches:
             - "dev"
             - "releases/**"
+        paths:
+            - "data-processing-lib/**"
```
roytman, line 10: we should add "kfp/kfp_support_lib/\**" and "kfp/kfp_ray_components/\*\*"

.github/workflows/build-library.yml
```diff
@@ -6,10 +6,20 @@ on:
...
+            - "!**.md"
+            - "!**/doc/**"
+            - "!**/.gitignore"
     pull_request:
         branches:
             - "dev"
             - "releases/**"
+        paths:
+            - "data-processing-lib/**"
```
roytman, line 19: "kfp/kfp_support_lib/**" and "kfp/kfp_ray_components/**"

.github/workflows/test-code-code2parquet.yml
```diff
@@ -0,0 +1,95 @@
...
+            - "data-processing-lib/**"
```
roytman, line 16: "kfp/kfp_support_lib/\**" and "kfp/kfp_ray_components/**"

.github/workflows/README.md
daw3rd: Can you explain at bit more?  As I've seen, all these transform tests do run in parallel (when a lib change is made).

transforms/universal/noop/python/src/noop_transform.py
daw3rd: this is just to trigger new builds to test which workflows are triggered.  Any .py or .md changes should hopefully have no real effect.

.github/workflows/README.md
daw3rd: this is working for me

.github/workflows/README.md
daw3rd: i gave this some thought, but I think there will need to be away to disable a transform's test, if only temporarily.  In the past this was done by renaming Makefile to something else so that the sub-directory would not be recursed into from the parent directory's Makefile.  We could do the same in the `test-transform.template` or just require that the Makefile is edited to include/uninclude a transform.  Hmm, but maybe maintaining this mode of renaming the Makefile is best to keep as it is then isolated to the transform.  If we do that, then we can do the automation you suggested above.

.github/workflows/build-library.yml
daw3rd: Why?  This workflow does not touch the kfp subdir

.github/workflows/build-library.yml
daw3rd: Why? This workflow does not touch the kfp subdir

.github/workflows/test-code-code2parquet.yml
daw3rd: I guess this is to get tests run in */kfp_ray directories?  This makes sense, although the test-src and test-image targets in kfp_ray/Makefile seem to be noops.  If you're suggesting this to get the images republished, I'm wondering if image publishing should be done in a separate workflow?

.github/workflows/test.yml
daw3rd: it is available in git history and can be found in the latest release branches - https://github.com/IBM/data-prep-kit/blob/v0.2.0/.github/workflows/test.yml

.github/workflows/README.md
daw3rd: done

.github/workflows/Makefile
daw3rd: done

.github/workflows/test.yml
daw3rd: done. added test.yml.old back

transforms/universal/filter/python/src/filter_local.py
daw3rd: this was just to try and trigger a test on filter.

.github/workflows/test-code-code2parquet.yml
daw3rd: thank you. that fixed some odd behavior i was seeing.

.github/workflows/test-kfp.yml
daw3rd: If we want randomness, I see no reason to impose this on a transform.  If we want every transform to always run its kfp test, then kfp_ray/Makefile needs to implement the test targets.  However, this would run all kfp transform tests on any change to the core dpk libs.  This might be too much.  I suggest instead we test kfp on a random transform whenever kfp/** changes.

daw3rd: And, we could add a target for transform/*/*/kfp_ray/** to trigger a test there if the kfp_ray files change.

.github/workflows/test-code-code2parquet.yml
daw3rd: Publishing of wheels and images for releases is handled manually per [this](https://github.com/IBM/data-prep-kit/blob/dev/RELEASE.md#publishing-wheels-and-images).  However, publishing the `latest` image can only happen on these changes, so maybe we're ok?

.github/workflows/README.md
daw3rd: ...but maybe I misunderstood `automatic validation`?

daw3rd: ah, ok will at a test yaml to make sure there are tests.ymls for all transforms in the repo.

.github/workflows/test-code-code2parquet.yml
daw3rd: We will run kfp tests separately and later add kfp tests to the test-transform.template to include kfp testing only when the transform kfp_ray files change

roytman (approved):
LGTM, but leave it to @revit13 to check her comments

.github/workflows/README.md
daw3rd: Add some details. not sure if its what you're looking for.

.github/workflows/README.md
daw3rd: done. in test-misc.yml

revit13 (approved):

Issue #585, opened 2024-09-11
[Bug] possible regression on ededupe code in release dev3


sujee:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/universal/tokenization

## In dev1 release (expected behavior)

 https://github.com/sujee/data-prep-kit/blob/rag-example1/examples/notebooks/rag/rag_1A_dpk_process_ray.ipynb

### Step-3: Chunking 

output

```text
Files processed : 3
Chunks created : 2,042
Input data dimensions (rows x columns)=  (3, 12)
Output data dimensions (rows x columns)=  (2042, 15)
```

###  Step-4 : EDedupe

Output

```text
Input data dimensions (rows x columns)=  (2042, 15)
Output data dimensions (rows x columns)=  (1324, 15)
Input chunks before exact dedupe : 2,042
Output chunks after exact dedupe : 1,324
Duplicate chunks removed :   718
```

## In Dev-3 (incorrect behavior)

https://github.com/sujee/data-prep-kit/blob/rag-example1-dev/examples/notebooks/rag/rag_1A_dpk_process_ray.ipynb

### Step-3: Chunking 

output

```text
Files processed : 3
Chunks created : 1,973
Input data dimensions (rows x columns)=  (3, 12)
Output data dimensions (rows x columns)=  (1973, 15)
```

### Step-4: EDedupe

output

```text
Input data dimensions (rows x columns)=  (1973, 15)
Output data dimensions (rows x columns)=  (2, 16)
Input chunks before exact dedupe : 1,973
Output chunks after exact dedupe : 2
Duplicate chunks removed :   1971
```

**Note: how resulting number of chunks is only 2**

This is the problem.

The result is not chunks but the documents!

This in turn breaks vector search and RAG responses.

### Reproduction script

dev1 (expected behavior) : https://github.com/sujee/data-prep-kit/blob/rag-example1/examples/notebooks/rag/rag_1A_dpk_process_ray.ipynb

dev3 (incorrect behavior) : https://github.com/sujee/data-prep-kit/blob/rag-example1-dev/examples/notebooks/rag/rag_1A_dpk_process_ray.ipynb

### OS

Ubuntu

### Python

3.11.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

daw3rd:
[...]

PR #598, opened 2024-09-18
doc_id and source_doc_id params in doc_chunk

dolfim-ibm:

This expands the doc_chunk parameters with the possibility to tune the propagation of the doc_id from the input tables to the results.

https://github.com/IBM/data-prep-kit/issues/585


blublinsky (approved):
LGTM

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
```diff
@@ -48,9 +52,11 @@ def __str__(self):
 
 
 default_content_column_name = "contents"
+default_doc_id_column_name = "doc_id"
```
daw3rd: Is this the right value?  I think ededup uses `document_id`.  Should we try and standardize?

daw3rd (changes requested):

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
dolfim-ibm: very good catch! let's align it here.

daw3rd (approved):

PR #612, opened 2024-09-22
Add delay before deleting ray cluster in kfp component.

revit13:

For debugging purposes add delay before deleting ray cluster in kfp component.

### DIFF
--- a/kfp/doc/simple_transform_pipeline.md
+++ b/kfp/doc/simple_transform_pipeline.md
@@ -100,7 +100,7 @@ The input parameters section defines all the parameters required for the pipelin
     # noop parameters
     noop_sleep_sec: int = 10,
     # additional parameters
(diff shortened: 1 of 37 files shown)


blublinsky:
I am not sure we need this. We already have delay in execution

roytman:
> I am not sure we need this. We already have a delay in execution

Hi @blublinsky , it is a different delay, if somebody wants to debug the Ray cluster, this delay will allow it before destroying the Ray cluster.

blublinsky:
> > I am not sure we need this. We already have a delay in execution
> 
> Hi @blublinsky , it is a different delay, if somebody wants to debug the Ray cluster, this delay will allow it before destroying the Ray cluster.

Oh, then yes. It was not clear. sorry

roytman (approved):
LGTM

PR #614, opened 2024-09-23
Update to exception processing doc

daw3rd:
clarity


data-processing-lib/doc/transform-exceptions.md
```diff
@@ -5,13 +5,20 @@ a given input data (e.g., pyarrow Table) does not have the expected format (.e.g
...
 This will cause only the error-causing datum to be ignored and not written out,
-but allow continued processing of tables by the transform.
-In both cases, the runtime will log the exception as an error.
\ No newline at end of file
+but allow continued processing of data by the transform.
+
+To terminate processing of all data by the runtime (2 above), 
+the transform should either 
+* throw any exception in `__init__()`, or 
+* throw an UnrecoverableException from AbstractBinaryTransform's `transform_binary`, `flush_binary(), or
```
blublinsky: throwing any exception in `__init__()` will not terminate the process. It will just continue re creating an actor

blublinsky (changes requested):

data-processing-lib/doc/transform-exceptions.md
daw3rd: so what happens, effectively, to the actor and the processing of data when init() throws an exception.  I would presume that no data is processed by an actor that gets an exception in init(),and since all actors likely get the same exception, all data processing is ended?

blublinsky: Added more precise doc in https://github.com/IBM/data-prep-kit/pull/616

daw3rd:
This was done in PR #616

daw3rd:
commited additional .md changes, so re-opening

PR #616, opened 2024-09-24
adding exception handling for transforms creation

blublinsky:
Without this we can get into infinite loop of errors, which overwhelms the logger


blublinsky:
Oh, there is an interesting thing with Ray, actor init exception is not propagated, as it is a separate process. Hopefully it is fixed now

daw3rd (approved):

PR #617, opened 2024-09-24
log error in case of exception

shivdeep-singh-ibm:
In case of exception while processing repo, log the error and move to next repo. Exception should not crash the actor.


Param-S (approved):
lgtm

Issue #618, opened 2024-09-24
Support for Python 3.12


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

We have asked DPK users to use Python 3.11, since Ray was not supported on Python 3.12. In workshops, we see people who have 3.12 and are hesitant to use 3.11. We need to investigate if Ray's issue with 3.12 has been solved and if yes, remove the restriction on using 3.12. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #619, opened 2024-09-24
support python 3.12

roytman:

There is a user requirement to support Python 3.12.
Ray started to support Python 3.12 from `2.36.0`. This PR upgrades Ray to `2.36.1` as it's bug fix release.

#618


.make.defaults
```diff
@@ -51,7 +51,7 @@ DOCKER_LOCAL_IMAGE=$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION)
 DOCKER_REMOTE_IMAGE=$(DOCKER_REGISTRY_ENDPOINT)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION)
 DOCKER_SPARK_BASE_IMAGE_NAME=data-prep-kit-spark-3.5.2
 DOCKER_SPARK_BASE_IMAGE=$(DOCKER_SPARK_BASE_IMAGE_NAME):$(DOCKER_IMAGE_VERSION)
-RAY_BASE_IMAGE?=docker.io/rayproject/ray:${RAY}-py310
+RAY_BASE_IMAGE?=docker.io/rayproject/ray:${RAY}-py312
```
blublinsky: Can we make Python version a separate parameter in make.versions? I am sure it can be used elsewhere

data-processing-lib/ray/pyproject.toml
```diff
@@ -12,7 +12,7 @@ authors = [
 ]
 dependencies = [
     "data-prep-toolkit>=0.2.1.dev3",
-    "ray[default]==2.24.0",
+    "ray[default]==2.36.1",
```
blublinsky, line 15: This should come from make.versions

kfp/kfp_support_lib/kfp_v1_workflow_support/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "data_prep_toolkit_kfp_v1"
 version = "0.2.1.dev3"
-requires-python = ">=3.10,<3.12"
+requires-python = ">=3.10"
```
blublinsky: Again, can we use values from make.default. I also think it should be 
equires-python = ">=3.10,<3.13"

kfp/kfp_support_lib/kfp_v2_workflow_support/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "data_prep_toolkit_kfp_v2"
 version = "0.2.1.dev3"
-requires-python = ">=3.10,<3.12"
+requires-python = ">=3.10"
```
blublinsky: see above

kfp/kfp_support_lib/shared_workflow_support/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "data_prep_toolkit_kfp_shared"
 version = "0.2.1.dev3"
-requires-python = ">=3.10,<3.12"
+requires-python = ">=3.10"
```
blublinsky: see above. Actually I would suggest requires-python = ">=3.10,<3.13"

blublinsky:
Why exactly we need python 3.12

roytman:
> Why exactly we need python 3.12

During Sujee's workshop, he was asked several times about Python 3.12

blublinsky:
Also see https://github.com/IBM/data-prep-kit/blob/dev/.make.versions#L115

roytman:
> Also see https://github.com/IBM/data-prep-kit/blob/dev/.make.versions#L115

you are checking the `dev` branch and not this PR

.make.defaults
roytman: We can, but this format (py312) is used only here, so I don't think we should propagate it into make.versions

data-processing-lib/ray/pyproject.toml
roytman: done

blublinsky:
> > Also see https://github.com/IBM/data-prep-kit/blob/dev/.make.versions#L115
> 
> you are checking the `dev` branch and not this PR

Sorry, my bad

kfp/kfp_support_lib/kfp_v1_workflow_support/pyproject.toml
roytman: it is, now

kfp/kfp_support_lib/shared_workflow_support/pyproject.toml
roytman: done

.make.defaults
blublinsky: Also this is kinda dangerous. what if I am on Python3.10? It really should get current Python version and adjust the base Ray image accordingly

roytman: No, here the question is the Python version of the images we deploy to the repository. This includes the RayKFP image. It doesn't connect to the Python version of a certain user. 
The original issue is to support Python v3.12, now we do support it. Should we change our Python default version to v3.12? probably not.

I'm going to return it to v3.10, but I'm open to other ideas. 
 
 CC: @daw3rd , @shahrokhDaijavad @yuanchi2807 @Bytes-Explorer @blublinsky

daw3rd: Aren't images different than venv?  Base images can be whatever works, as long as we do the transform tests in the image, which we are.  

Separately, we probably need to start using a github workflow matrix to test at least 3.11 and 3.12.  This requires changing .github/workflow/test-transform.template somehow and then regenerating them with `make`.

roytman: Ray is very sensitive to its and Python versions; they should be the same on clients and servers. We should provide a set of images, or users will have to build the images by themselves.

roytman: OK, I validated that all tests passed with Python 3.12, not I will return the default setting to 3.10.

data-processing-lib/ray/Makefile
```diff
@@ -17,6 +17,9 @@ setup::
 
 set-versions: .check-env
 	$(MAKE) TOML_VERSION=$(DPK_LIB_VERSION) .defaults.update-toml
+	cat pyproject.toml | sed -e 's/"ray[default]==.*",/"ray[default]==$(RAY)",/' > tt.toml
```
daw3rd, line 20: should this be `>=` which is similar to spark.

.make.defaults
```diff
@@ -573,6 +576,18 @@ MINIO_ADMIN_PWD= localminiosecretkey
...
 	fi
 
+# Updates the Python supported versions field of the pyproject.toml file to the given versions
+# Expects  REQUIRED_PYTHON_VERSIONS
+.PHONY: .defaults.__update-toml-python-versions
+.defaults.__update-toml-python-versions:
+	@# Help: Set the version= field of pyproject.toml
+	if [ -e pyproject.toml ]; then					\
+	    cat pyproject.toml | sed -e 				\
+	    's/^requires-python[ ]*=.*"/requires-python = "'${REQUIRED_PYTHON_VERSIONS}'"/' \
```
daw3rd, line 586: can you make the white space optional?

daw3rd (changes requested):

.make.defaults
roytman: what do you mean, I think it is.
`[ ]*`

data-processing-lib/ray/Makefile
roytman: I don't think it's a good idea. Ray is very sensitive to version mismatches between the client and server, even down to the last digit. Using the `>=` setting might lead to incompatibility between our Ray images (DPK and KFP-Ray)."

.make.defaults
daw3rd: ah i'm losing it. sorry, was looking at the substitution value.

data-processing-lib/ray/Makefile
daw3rd: ok

daw3rd (approved):

touma-I (changes requested):
Please slack me when ready to merge. Thanks

roytman:
> Please slack me when ready to merge. Thanks

Hi @touma-I , please take a look

.make.versions
```diff
@@ -118,9 +118,11 @@ KFP_v2=2.2.0
 KFP_v2_SDK=2.8.0
 KFP_v1=1.8.5
 KFP_v1_SDK=1.8.22
-RAY=2.24.0
+RAY=2.36.1
```
touma-I, line 121: @roytman just to confirm: Ray 2.36.1 will work with Python 3.10, 3.11 and 3.12. right?

touma-I (approved):

.make.versions
roytman: I ran all CI/CD tests with Python 3.12 and now it runs with Python 3.10. So yes, it works.

Issue #625, opened 2024-09-25
[Bug] Fail to build transform spark image when using a release tag x.y.z (without any .devN suffix)


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Makefiles

### What happened + What you expected to happen

When building release 0.2.1, the spark transform was failing. On close inspection, I found that the make file correctly builds data-prep-kit-spark-3.5.2:latest but when building the image for the transform, the makefile was referencing data-prep-kit-spark-3-5-2:0.2.1 (using the **0.2.1** tag instead of the **latest** tag)

The log shows the following when building the image for the base image:
```
docker build -t data-prep-kit-spark-3.5.2:latest 	\
	--platform linux/amd64  			\
	--build-arg EXTRA_INDEX_URL=	\
	--build-arg BASE_IMAGE=  		\
	--build-arg BUILD_DATE=2024-09-25T14:15:22Z  	\
	--build-arg GIT_COMMIT=fafecea . 
```

The log shows for following when building the image for the transforms.
```
docker build -t noop-spark:0.2.1 	\
	--platform linux/amd64  			\
	--build-arg EXTRA_INDEX_URL=	\
	--build-arg BASE_IMAGE=data-prep-kit-spark-3.5.2:0.2.1  		\
	--build-arg BUILD_DATE=2024-09-25T14:37:58Z  	\
	--build-arg GIT_COMMIT=fafecea . 
```

The complete log can be found here: https://github.com/IBM/data-prep-kit/actions/runs/11034666653/job/30648903714?pr=622

### Reproduction script

modify .make.versions and set DPK_VERSION_SUFFIX=
rebuild spark image and noop-transform image

### OS

Ubuntu

### Python

3.10.x, 3.11.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #626, opened 2024-09-25
set spark base image tag to that from .make.versions

daw3rd:
Releasing has problems since the tag for the base spark image was fixed to be `latest`.  
It now is set from .make.versions, like other images.

#625

### DIFF
--- a/data-processing-lib/spark/Makefile
+++ b/data-processing-lib/spark/Makefile
@@ -4,7 +4,6 @@ include $(REPOROOT)/.make.defaults
 SPARK_VERSION=3.5.2
 DOCKER_IMAGE_NAME=data-prep-kit-spark-$(SPARK_VERSION)
 DOCKER_IMAGE_LIB_NAME=data-prep-kit-spark
-DOCKER_IMAGE_VERSION := latest
 
 
 .check-env::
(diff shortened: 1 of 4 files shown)


blublinsky:
Can you also fix https://github.com/IBM/data-prep-kit/blob/dev/.make.defaults#L52. We should not use `data-prep-kit-spark-3.5.2` we already have it defined here https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/spark/Makefile#L4. Maybe move Spark version make version, similar to https://github.com/IBM/data-prep-kit/blob/dev/.make.versions#L115

blublinsky:
Also, I do not think it works as desired https://github.com/IBM/data-prep-kit/actions/runs/11036249651/job/30654242163?pr=626#step:4:9415

daw3rd:
> Also, I do not think it works as desired https://github.com/IBM/data-prep-kit/actions/runs/11036249651/job/30654242163?pr=626#step:4:9415

When using a version suffix (i.e. dev0), the image tags are always set to `latest` (see .make.versions), but when there is no suffix (i.e. when cutting a release), the tag should be the release versions (i.e. 0.2.1).  The ci/cd link above is for a dev0 suffix so has the `latest` tag.

blublinsky:
> > Also, I do not think it works as desired https://github.com/IBM/data-prep-kit/actions/runs/11036249651/job/30654242163?pr=626#step:4:9415
> 
> When using a version suffix (i.e. dev0), the image tags are always set to `latest` (see .make.versions), but when there is no suffix (i.e. when cutting a release), the tag should be the release versions (i.e. 0.2.1). The ci/cd link above is for a dev0 suffix so has the `latest` tag.

How can we prove this? It is soooo confusing

daw3rd:
> Also, I do not think it works as desired https://github.com/IBM/data-prep-kit/actions/runs/11036249651/job/30654242163?pr=626#step:4:9415

Can you be more specific?

blublinsky (approved):

PR #629, opened 2024-09-26
align docling versions among transforms

dolfim-ibm:

This transform aligns the docling-core version used in `pdf2parquet` and `doc_chunk`, which simplified the creation of the single package.


touma-I:
Thanks @dolfim-ibm I need to make one more change to move all the requirements out of the pyproject.toml and into a requirements.txt. will slack you when this is done.

touma-I (approved):
@dolfim-ibm Looks good. thanks for your patience! I will apply the same when we do the other transforms. I had to list the deep search libraries in the requirements for ray in order for the docker build to pass.

Issue #586, opened 2024-09-12
[Feature] Enable pure python transforms in new spark runtime.


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

With the r[ecent PR to the spark runtime](https://github.com/IBM/data-prep-kit/pull/406), to allow checkpointing and pure-python transforms to run in spark, we should begin enabling transforms listed below to run in the spark runtime.  Largely this means creating the boilerplate main() and associated Spark configuration classes.  See the [new filter transform ](https://github.com/IBM/data-prep-kit/blob/dev/transforms/universal/filter/spark/src/filter_transform_spark.py) for an example.  

The transforms suggested  this work are as follows:

Universal
- [ ] profiler
- [ ] resize
- [ ] tokenization

Code
- [ ] code2parquet
- [ ] code_quality
- [ ] header_cleanser
- [ ] malware
- [ ] proglang_select
- [ ] repo_level_ordering

Language
- [ ] doc_chunk
- [ ] doc_quality
- [ ] lang_id
- [ ] pdf2parquet
- [ ] pii_redactor
- [ ] text_encoder

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #630, opened 2024-09-26
Adding Resize Spark

blublinsky:
Adding additional transforms to Spark pipeline

https://github.com/IBM/data-prep-kit/issues/586


transforms/universal/resize/spark/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG BASE_IMAGE=quay.io/dataprep1/data-prep-kit/data-prep-kit-spark-3.5.2:0.2.1.dev0
```
daw3rd: 0.2.1.dev0 is no longer used in dev.  to be consistent with other spark transforms (recent change), use `latest` as the tag.  Note however, that this is generally overridden from the Makefile anyway by setting BASE_IMAGE when docker build is called.  But for consistency it would be nice to change to `latest`.

transforms/universal/resize/spark/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+build:: build-dist image
+
+publish: publish-image
+
+publish-image:: .transforms.publish-image-spark
+
+# set the version of python transform that this depends on.
+set-versions: 
+	$(MAKE) TRANSFORM_PYTHON_VERSION=${NOOP_PYTHON_VERSION} TOML_VERSION=$(NOOP_SPARK_VERSION) .transforms.set-versions
```
daw3rd: NOOP -> RESIZE?

transforms/universal/resize/spark/README.md
```diff
@@ -0,0 +1,64 @@
...
+```shell
+make build 
+```
+
+## Configuration and command line Options
+
+The set of dictionary keys holding [BlockListTransform](src/blocklist_transform.py)
+configuration for values are as follows:
+
+* _max_rows_per_table_ - specifies max documents per table
```
daw3rd, line 27: To better future-proof this file, shouldn't it defer to the python readme for configuration and CLI?

daw3rd (changes requested):

transforms/universal/resize/spark/Dockerfile
blublinsky: done

transforms/universal/resize/spark/Makefile
blublinsky: Done

daw3rd (approved):

PR #632, opened 2024-09-26
Define spark version in .make.versions

daw3rd:
To define the spark version in a single global place.


cmadam (approved):
Makes sense to me.

Issue #641, opened 2024-09-29
[Feature] Extend document chunker transform to support fixed-size token window chunker with overlap


juancappi:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

The current document chunker only supports Markdown and Docling JSON chunking. However, there are many other valid chunking criteria, and specifically, there is a need to support a fixed-size token window with token overlap. This functionality is important for classification-based transforms that operate on smaller chunks of data.

While there are multiple approaches to token counting, I propose leveraging the existing capabilities of the llama-index-core library (which is already a dependency) to implement this feature

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

dolfim-ibm:
The PR of @juancappi was merged. Closing the issue.

PR #642, opened 2024-09-29
feat: #641 - Extend document chunker transform to support fixed-size token window chunker with overlap

juancappi:

Partially implements #641 

This is a draft for pure python only. Ray is still pending


juancappi:
@touma-I would you please take a look to confirm I'm in the right track?

juancappi:
@dolfim-ibm this is ready for review

dolfim-ibm (approved):
Nice. LGTM

PR #646, opened 2024-09-30
Transform for Code Profiling

pankajskku:
This tranform extracts the base syntactic concepts from the multi-language source codes and represent these concepts in an unified langauge-agnostic representation that can be further used for multi-lnaguage data profiling. While programming languages expose similar syntactic building blocks to represent programming intent, such as importing packages/libraries, functions, classes, loops, conditionals, comments and others, these concepts are expressed through language-specific grammar, defined by distinct keywords and syntactic form.

Data profiling, in the context of machine learning, is the process of examining and analyzing data to create
useful statistics. These statistics are used both as an aid for better comprehension of the properties of data as
well as for a variety of downstream data processing tasks such as data valuation (assessing the value of data
relative to the business objectives at hand) and data curation (filtering and prioritizing training data based on
derived thresholds). In the Large Language Model (LLM) setting, training data is typically unstructured in
nature comprising natural language text, images, and code. In this work, we specifically focus on code-LLMs,
where the quality of code training data substantially affects the model accuracy of LLM-based coding tasks
such as code generation and summarization. Therefore, having the capabilities to characterize code data in
terms of programming language concepts aids in both deriving insights related to code training/evaluation
data and in the downstream curation of code training data. In this work, we address the problem of profiling
multi-lingual code datasets by extracting an extensible user-defined set of syntactic concepts
over arbitrary programming languages.


daw3rd (changes requested):
maybe I've missed it, but where is the transform class.

pankajskku:
> maybe I've missed it, but where is the transform class.
@daw3rd 
Thank you. 
The transform is at transforms/code/syntactic_concept_extractor/python/src/syntactic_concept_extractor_transform.py

daw3rd (changes requested):
You are checking in 100mb in the `input` directory.  This almost doubles the size of the clone repository.  Can we get these from somewhere else and not stored them in the repo and, presumably the published wheel containing the transform?

```
syntactic_concept_extractor$ du -sm *
1	Makefile
1	README.md
91	input
5	python
1	ray

```

pankajskku:
> You are checking in 100mb in the `input` directory. This almost doubles the size of the clone repository. Can we get these from somewhere else and not stored them in the repo and, presumably the published wheel containing the transform?
> 
> ```
> syntactic_concept_extractor$ du -sm *
> 1	Makefile
> 1	README.md
> 91	input
> 5	python
> 1	ray
> ```

@daw3rd I have removed the static .so from the PR and made sure that the tree-sitter-bindings folder gets cloned from a git repository while creating venv.

daw3rd:
> > You are checking in 100mb in the `input` directory. This almost doubles the size of the clone repository. Can we get these from somewhere else and not stored them in the repo and, presumably the published wheel containing the transform?
> > ```
> > syntactic_concept_extractor$ du -sm *
> > 1	Makefile
> > 1	README.md
> > 91	input
> > 5	python
> > 1	ray
> > ```
> 
> @daw3rd I have removed the static .so from the PR and made sure that the tree-sitter-bindings folder gets cloned from a git repository while creating venv.

I still see them in the PR under `input/tree-sitter-bindings`. 
And, if they are removed how will this work if pip installed from pypi?  Can they be downloaded at runtime.

pankajskku:
> > > You are checking in 100mb in the `input` directory. This almost doubles the size of the clone repository. Can we get these from somewhere else and not stored them in the repo and, presumably the published wheel containing the transform?
> > > ```
> > > syntactic_concept_extractor$ du -sm *
> > > 1	Makefile
> > > 1	README.md
> > > 91	input
> > > 5	python
> > > 1	ray
> > > ```
> > 
> > 
> > @daw3rd I have removed the static .so from the PR and made sure that the tree-sitter-bindings folder gets cloned from a git repository while creating venv.
> 
> I still see them in the PR under `input/tree-sitter-bindings`. And, if they are removed how will this work if pip installed from pypi? Can they be downloaded at runtime.

@daw3rd 
I'm sorry, I cleaned them now.
I decided to change the approach and rely on adding the utility code in the transform itself to clone the .so's during the transform runtime. This way, it could work at the runtime. Please let me know what you think.

.github/workflows/test-code-syntactic_concept_extractor.yml
daw3rd, line 1: Please merge dev into your fork and then recreate this file.  It will help to cancel ci/cd builds from stale commits.

daw3rd (changes requested):

transforms/code/syntactic_concept_extractor/python/src/syntactic_concept_extractor_transform.py
```diff
@@ -0,0 +1,195 @@
...
+        super().__init__(config)
+        
+        self.contents = self.config.get("contents")
+        self.language = self.config.get("language")
+
+        def ensure_tree_sitter_bindings():
+            bindings_dir = "tree-sitter-bindings"
+            if not os.path.exists(bindings_dir):
+                print("Cloning tree-sitter bindings...")
+                result = subprocess.run(["git", "clone", "https://github.com/pankajskku/tree-sitter-bindings.git", bindings_dir])
```
daw3rd: We have found a potential problem with loading like this.  The problem is when running multi-threaded/process on a common file system (this can happen with running local Ray with more than 1 worker).  This can result in conflicts and corruption, in this case in your cloned repo.  One solution would be to have a bindings_dir that is somehow unique to the transform instance.  A uuuid. would probably work.  

The other thing then is when, if ever, to remove this directory/repo.  If you go with the unique bindings dir, then cleaning on flush() could be done.

daw3rd: Not sure if python has it, but in java you can have a callback when the process exits.

daw3rd (dismissed):

transforms/code/syntactic_concept_extractor/python/src/syntactic_concept_extractor_transform.py
pankajskku: Thank you for identifying the bug.
Updated function to handle the multithreaded implementation
```
 def ensure_tree_sitter_bindings():
            # Get the directory where the script is located
            script_dir = os.path.dirname(os.path.abspath(__file__))
            # Generate a unique directory for the bindings based on a UUID
            bindings_dir = os.path.join(script_dir, f"tree-sitter-bindings-{uuid.uuid4()}")
            # Clone the bindings only if the unique directory does not exist
            if not os.path.exists(bindings_dir):
                print(f"Cloning tree-sitter bindings into {bindings_dir}...")
                result = subprocess.run(["git", "clone", "https://github.com/pankajskku/tree-sitter-bindings.git", bindings_dir])
                if result.returncode != 0:
                    raise RuntimeError(f"Failed to clone tree-sitter bindings into {bindings_dir}")
                return bindings_dir  
```
The deletion happens at the end of the transform function with the following code. 
```
        # Register cleanup for when the process exits
        atexit.register(shutil.rmtree, self.bindings_dir)
```

.github/workflows/test-code-syntactic_concept_extractor.yml
pankajskku: Done.

pankajskku:
@daw3rd Please let me know your opinion on the updated PR.

touma-I (changes requested):
@pankajskku Please slack me when you have a chance. Internal ID: touma@us.ibm.com

transforms/code/syntactic_concept_extractor/python/src/syntactic_concept_extractor_transform.py
daw3rd: You can now use the [MultiLock()](https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/python/src/data_processing/utils/multilock.py) class to make sure that the multiple (ray local) processes do not try to load this at the same time.

daw3rd: Unfortunately, for the deletion you really need a semaphore.

daw3rd: Unless you want to write a semaphore class (using MultiLock?), you probably should just not do the deletion, but make sure you put it in a temp dir with the assumption that the os will likely cleanup later.

transforms/code/code_profiler/python/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+WORKDIR /home/dpk
+
+# Copy and install data processing libraries 
+# These are expected to be placed in the docker context before this is run (see the make image).
+COPY --chown=dpk:root data-processing-lib-python/ data-processing-lib-python/
+RUN cd data-processing-lib-python && pip install --no-cache-dir -e .
+
+COPY --chown=dpk:root src/ src/
+COPY --chown=dpk:root pyproject.toml pyproject.toml
+COPY --chown=dpk:root README.md README.md
```
touma-I, line 20:
```suggestion
COPY --chown=dpk:root requirements.txt requirements.txt
```

How was it building the Dockerfile and passing the UT  if it was not loading the requirements??

touma-I (changes requested):

transforms/code/code_profiler/python/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
```
touma-I, line 22:
```suggestion
PY --chown=dpk:root requirements.txt requirements.txt

```

How was this building the Dockerfile and passing the test without the requirements.txt ???

touma-I (changes requested):

transforms/code/code_profiler/python/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    { name = "Pankaj Thorat", email = "pankaj.thorat@ibm.com" },
+]
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[build-system]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+build-backend = "setuptools.build_meta"
+
```
touma-I, line 17:
```suggestion

[tool.setuptools.dynamic]
dependencies = {file = ["requirements.txt"]}

```

transforms/code/code_profiler/python/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+version = "0.2.2.dev0"
+requires-python = ">=3.10"
+description = "Code Profiler Python Transform"
+license = {text = "Apache-2.0"}
+readme = {file = "README.md", content-type = "text/markdown"}
+authors = [
+    { name = "Pankaj Thorat", email = "pankaj.thorat@ibm.com" },
+]
+
+[tool.setuptools.dynamic]
```
touma-I:
```suggestion
```

transforms/code/code_profiler/python/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+dependencies = {file = ["requirements.txt"]}
```
touma-I:
```suggestion
dynamic = ["dependencies"]
```

touma-I (changes requested):
@pankajskku Please see additional changes. I was pulled how your UT was passing without the proper requirements.txt in the Dockerfile.

touma-I (approved):

PR #654, opened 2024-10-02
Change the calculation of the desired ray actors

revit13:

Recent [PR](https://github.com/IBM/data-prep-kit/pull/616) reveals that the number of live ray actors does not reach the number of desired ray actors which causes the kfp workflow to eventually fail. This PR includes a more accurate calculation of the desired ray actors to fix that.


kfp/kfp_support_lib/shared_workflow_support/src/runtime_utils/kfp_utils.py
```diff
@@ -138,7 +138,7 @@ def default_compute_execution_params(
         cluster_gpu = w_options["replicas"] * w_options.get("gpu", 0.0)
         logger.info(f"Cluster available CPUs {cluster_cpu}, Memory {cluster_mem}, GPUs {cluster_gpu}")
         # compute number of actors
-        n_actors_cpu = int(cluster_cpu * 0.85 / a_options.get("num_cpus", 0.5))
+        n_actors_cpu = int((cluster_cpu - 1) * 0.7 / a_options.get("num_cpus", 0.5))
```
blublinsky, line 141: To make it symmetric, can we try:

 n_actors_cpu = int((cluster_cpu - 1) * 0.85 / a_options.get("num_cpus", 0.5))

transforms/universal/ededup/kfp_ray/src/ededup_compute_execution_params.py
```diff
@@ -94,9 +95,9 @@ def ededup_compute_execution_params(
         )
         sys.exit(1)
     # Define number of workers
-    n_workers = int((0.85 * cluster_cpu - required_hash_cpu) / actor_cpu)
+    n_workers = int((0.85 * (cluster_cpu - 1) - required_hash_cpu) / actor_cpu)
     print(f"Number of workers - {n_workers}")
-    if n_workers < 2:
+    if n_workers < 0:
```
blublinsky: can we do:

if n_workers <= 0:

kfp/kfp_support_lib/shared_workflow_support/src/runtime_utils/kfp_utils.py
revit13: noop fails with this change...

transforms/universal/ededup/kfp_ray/src/ededup_compute_execution_params.py
revit13: sure ededup runs ok with it

kfp/kfp_support_lib/shared_workflow_support/src/runtime_utils/kfp_utils.py
blublinsky: lets put back

blublinsky (approved):
brilliant

PR #655, opened 2024-10-02
Updated RAG example for DPK version 0.2.1

sujee:

Updating RAG examples to work with 0.2.1


shahrokhDaijavad (approved):
I have tested all Notebooks in this PR successfully. My only comment is that in my testing, doing File --> Close and Shutdown Notebook did not release the db.lock and I had to do a hard shutdown of Jupyter (Ctrl-C from the terminal) to make sure db.lock is released.

Issue #639, opened 2024-09-29
[Feature] Relax release rules so MICRO version (0.0.X) can be increased for transforms inpdendent of library/runtime


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

The current release process increments the MICRO version for both transforms and library. Given the high rate of changes to the transforms, this process adds overhead an unnecessary churn by forcing us to increment the release number for the library each time we want to create a release cutoff for the transforms. We propose decoupling the two by:

- Allow a new release of the transforms to increment the transform MICRO number(_._.X) while keeping the library MICRO number the same (_._.Y) with Y <= X
- When a minor release for the transforms and  library is needed, increment the MINOR number and reset the MICRO to 0  before appending the devN suffix (_.X.0.dev0)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

daw3rd:
With the above in mind, we have the following:

**Guiding principals/goals**

1. github "releases" will be published to a new minor (not micro) version number. For example, 0.3.0
2. when there is a github release, we will publish pypi wheels and images
3. transforms can increment the micro version independently of the core library version and formal github release
4. We will (for now) keep the dev suffix, until we decide we can live w/o it.
5. It is ok and expected to republish a transform at a new minor version that may be the same functionally as the previous version (i.e 0.3.0 and 0.2.9)
6. Transforms can define dependencies on DPK in pypi or in the repo source (choice is already supported today via USE_REPO_LIB_SRC in .make.defaults)

**TODOs:**

1. Transform version numbers need to be pushed down from .make.versions to a transform-specific location (i.e. transforms/noop/version)
2. make set-versions neeeds to still work through the repo to reset to a new minor version (i.e. 0.3.0) (and be extended to work on transform's version file)
3. Update release-branch.sh script to bump minor version numbers instead of micro.
4. Update RELEASE.md to capture these goals and changes.

PR #657, opened 2024-10-02
Move transform version numbers out of .make.versions to transform-specific file

daw3rd:
## Summary of changes
1. For each transform create a transform.config file used for inclusion across the transform's Makefiles (for example, universal/noop/transform.config)
2. Include this file in transform's Makefile (python/Makefile, ray/Makefile,spark/Makefile,kfp_ray/Makefile)
3. Remove TRANSFORM_NAME from Makefiles (it is now in transform.config)
4. Remove/comment out transform version numbers from .make.versions
5. Expose USE_REPO_LIB_SRC macro in transform Makefiles to help developers switch to using pypi instead of rep source for dpk dependencies.
6. Fix a problem in .make.defaults so that it does not update the dpk dependencies in a transform's toml when USE_REPO_LIB_SRC=0

Per issue #639 we would like to allow transforms to update their own version numbers.  
However we don't want such a change to trigger ci/cd on all other transforms.  
As such we need to move the version number out of .make.versions to a file that is more local to the transform.

Note that this is only the first of the TODOs listed in issue #639, but can be done independent of the others. 

#639


daw3rd:
Its not critical, but would improve ci/cd if you merge with the latest dev (it now cancels ci/cd builds from stale commits).  I think you have some old ci/cd builds that are impacting you and others.

transforms/code/code2parquet/transform.config
```diff
@@ -0,0 +1,20 @@
...
+TRANSFORM_NAME=code2parquet
+
+################################################################################
+# This defines the transforms' version number as would be used
+# when publishing the wheel.  In general, only the micro version
+# number should be advanced relative to the DPK_VERSION. 
+#
+# If you change the versions numbers, be sure to run "make set-versions" to 
+# update version numbers across the transform (e.g., pyproject.toml).
+CODE2PARQUET_PYTHON_VERSION=$(DPK_VERSION)
```
touma-I, line 17: @daw3rd How are we getting DPK_VERSION if we are not including .make.versions ?

revit13:
@daw3rd scripts/transforms/update_workflow_tags.sh needs to be update to take the versions from the config files.

transforms/code/code2parquet/transform.config
roytman: The transformer Makefiles include `$(REPOROOT)/transforms/.make.transforms` it includes `$(REPOROOT)/.make.defaults` and the last one includes `$(REPOROOT)/.make.versions`

roytman (approved):
LGTM

touma-I (approved):
Million thanks

PR #658, opened 2024-10-02
Add .make.* to triggers for transform tests.

daw3rd:
Because changes to .make.* files need to also be tested.


.github/workflows/test-universal-html2parquet.yml
```diff
@@ -0,0 +1,128 @@
+#
+# DO NOT EDIT THIS FILE: it is generated from test-transform.template,  Edit there and run make to change these files
+#
+name: Test - transforms/universal/html2parquet
```
touma-I: Did't we move this to language ? or maybe my brain is tired... either case, better block the merge for now. Sorry

touma-I (changes requested):
Need to investigate why html2parquet is showing up under universal while it was moved to language

.github/workflows/test-universal-html2parquet.yml
daw3rd: Nice catch.  This happens because git does not seem to remove directories/files when switching to branches w/o them.  Something to watch out for.

touma-I (approved):

daw3rd:
Clossing to get jobs unqueued

Issue #659, opened 2024-10-02
A set of enhancements for the HAP tranform


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

1) Add this transform to the table of all transforms in the root README file. (https://github.com/IBM/data-prep-kit/blob/dev/README.md) Python-only column for now. 
2) There are some comments in the PR, AFTER it got merged. See the comments about having default values for the configuration for the key values (https://github.com/IBM/data-prep-kit/pull/638) 
3) Once these defaults are added, they should be mentioned in the README file for the transform. 
4) This transform defaults to using a different name for the **contents** column in the parquet file that all other transforms use. 
5) Other minor changes in the transform README file: Add a hyperlink for NLTK, when it is mentioned for the first time. Typo: compatable => compatible 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

ian-cho:
@shahrokhDaijavad sure, please see pull request https://github.com/IBM/data-prep-kit/pull/661 and https://github.com/IBM/data-prep-kit/pull/663

daw3rd:
- [ ] I'd also like to see the defaults (currently defined in the CLI params) to be implemented in the transform's init() so that they can take effect when the transform is run outside of a runtime.

ian-cho:
> * [ ]  I'd also like to see the defaults (currently defined in the CLI params) to be implemented in the transform's init() so that they can take effect when the transform is run outside of a runtime.

@daw3rd Would you mind sharing an example?

PR #663, opened 2024-10-03
changed doc_text into contents in related files

ian-cho:
The column name `doc_text` is changed into `contents` in all relevant scripts and parquet files

correct column name to be in line with other module.

https://github.com/IBM/data-prep-kit/issues/659


touma-I (approved):
@ian-cho - This is good work. Thank you. merging it now.

PR #670, opened 2024-10-04
Update requirements.txt in the Notebook branch

shahrokhDaijavad:
Fixing the version of DPK to avoid future problems

### DIFF
--- a/examples/notebooks/rag/requirements.txt
+++ b/examples/notebooks/rag/requirements.txt
@@ -1,5 +1,8 @@
 ## Data prep kit
 
+data-prep-toolkit==0.2.1
+data-prep-toolkit-ray==0.2.1
+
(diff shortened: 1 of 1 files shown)


touma-I (approved):

examples/notebooks/rag/requirements.txt
```diff
@@ -1,5 +1,8 @@
 ## Data prep kit
 
+data-prep-toolkit==0.2.1
```
daw3rd, line 3: data-prep-toolkit and data-prep-toolkit-ray should not be needed as the other wheels depend on these 2.

touma-I: David, the idea is to "enforce" the dependencies to the lowest possible level at the application level.

PR #671, opened 2024-10-04
Fix kfp workflows which were created with a bad template

daw3rd:
To reinstated kfp testing.


daw3rd:
will start with a new branch since it seems to be using past commits

PR #672, opened 2024-10-04
Fix kfp workflows which were created with a bad template

daw3rd:
## Summary of changes
1. Check for existence of kfp_ray/Makefile instead of just the kfp_ray directory so that individual transforms can use the same mechanism to disable kfp pipeline tests on their transform (i.e., renaming the Makefile).
2. fix typo from .make.workflow to .make.workflows
3. Remove test of kfp_support_lib in transform kfp pipeline test workflows/actions.
5. Disable a few transform kfp pipeline tests due to failures by renaming their kfp_ray/Makefile.  Note that I suspect these have always been failing.  We see them now since, **for the first time**, we are testing all transform kfp pipelines.  I will be creating issues after this PR is merged.
    * code/repo_level_ordering
    * code/license_select
    * language/text_encoder
6. Fix what seems like a real bug in repo_level_ordering transform with a float value set with a string value in the wf.py

To enable kfp tests across all transforms (likely for the first time).


transforms/.make.workflows
```diff
@@ -9,7 +9,7 @@ define set_env_var
 endef
 
 
-KFP_ENDPOINT ?= "http://localhost:8080/" 
+KFP_ENDPOINT?="http://localhost:8080/"
```
daw3rd, line 12: this change is only to trigger the transform kfp tests.

transforms/code/repo_level_ordering/kfp_ray/repo_level_order_wf.py
```diff
@@ -122,23 +122,30 @@ def repo_level_order(
...
     runtime_pipeline_id: str = "pipeline_id",
-    runtime_code_location: dict = {'github': 'github', 'commit_hash': '12345', 'path': 'path'},
+    runtime_code_location: dict = {"github": "github", "commit_hash": "12345", "path": "path"},
     # repo_level_order parameters
     repo_lvl_stage_one_only: bool = False,
     repo_lvl_grouping_column: str = "repo_name",
     repo_lvl_store_type: str = "ray",
     repo_lvl_store_backend_dir: str = "",
-    repo_lvl_store_ray_cpus: float = "0.5",
+    repo_lvl_store_ray_cpus: float = 0.5,
```
daw3rd, line 148: Most of the other changes here were pre-commit cleanups, but this one seems like a real bug fix.

roytman: yes, it was and it's fixed in #619

.github/workflows/test-code-code2parquet-kfp.yml
```diff
@@ -56,10 +56,10 @@ jobs:
                   sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android /usr/local/share/powershell /usr/share/swift /usr/lib/jvm /usr/local/.ghcup
                   sudo docker rmi $(docker image ls -aq) >/dev/null 2>&1 || true
                   df -h
-            - name: Test KFP libs (shared and v1) and run a workflow
+            - name: Test V1 KFP workflow for transforms/code/code2parquet
               timeout-minutes: 120
               run: |
-                  if [ -e "@TARGET_TRANSFORM_DIR/Makefile" -a -d "@TARGET_TRANSFORM_DIR/kfp_ray" ]; then
+                  if [ -e "transforms/code/code2parquet/Makefile" -a -e "transforms/code/code2parquet/kfp_ray/Makefile" ]; then
```
roytman, line 62: can we combine the lines 62-86 in a shared Make rule, which will be the same for all transformers, and just change its arguments?

revit13 (approved):

revit13:
@daw3rd according to https://github.com/IBM/data-prep-kit/actions/runs/11193934073/job/31119824449 it seems header_cleanser should be added to the kfp black list.

.github/workflows/test-code-code2parquet-kfp.yml
daw3rd: 100% agree with this, but can I put that on @revit13 for another PR.

Issue #673, opened 2024-10-05
[Feature] Single Pypi package for python, ray and spark and single pypi package for all transforms 


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core, Transforms/Other

### Feature

The current model for releasing the runtime and the transforms create a different package such as::

- data-prep-toolkit, data-prep-toolkit_ray, data-prep-transforms, data-prep-transfroms-ray, etc

We propose deploying a single package for all libraries/runtimes and a single package for all transforms with dependencies as follow:

- data-prep-toolkit[ray, spark] 
- data-prep-transform[all, ray, spark, pdf2parquet, doc_id, doc_chunk, etc.] 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #674, opened 2024-10-06
Packaging modules as single package with extra (data-prep-kit[ray,spark] and data-prep-kit-tranforms[all, ray, etc.])

touma-I:

In this release, we create a single pypi package for data-prep-kit[dev,ray,spark] and a single data-prep-kit-transforms[all,ray,code_quality,doc_id,etc]

https://github.com/IBM/data-prep-kit/issues/673


.make.defaults
```diff
@@ -627,7 +636,7 @@ MINIO_ADMIN_PWD= localminiosecretkey
 	rm -rf dist || true
 	rm -rf src/*egg-info || true
 	${PIP} install --upgrade build
-	${PYTHON} -m build
+	${PYTHON} -m build $(BUILD_WHEEL_ARG)
```
daw3rd, line 639: Can we renaming BUILD_WHEEL_ARG to PYTHON_BUILD_EXTRA_ARGS to match DOCKER_BUILD_EXTRA_ARGS?

daw3rd: and a default (empty?) should be defined above, similar to DOCKER_BUILD_EXTRA_ARGS

touma-I: Hi @daw3rd . No need for Default. The default is none anyway.

PR #684, opened 2024-10-08
Minor change to kfp workflow.

revit13:
Following request from @roytman and @daw3rd this pr tries to merge common parts in kfp workflow different steps.
In addition this PR builds the workflow even if its part of the workflow's black list.


roytman (approved):
LGTM

.github/workflows/test-code-code_quality-kfp.yml
```diff
@@ -56,33 +56,26 @@ jobs:
...
-                      #make -C kfp/kfp_support_lib test
-                      make -C transforms/code/code_quality workflow-build
-                      source $K8S_SETUP_SCRIPTS/common.sh
-                      make -C transforms/code/code_quality workflow-test
-                      echo "Run transforms/code/code_quality completed"
-                   else
-                      echo "Skipping transforms/code/code_quality kfp test for lack of Makefile and/or kfp_ray/Makefile"
+                     transform=$(basename "transforms/code/code_quality")
+                     if echo ${KFP_BLACK_LIST} | grep -qv ${transform}; then
+                        $PWD/scripts/workflow_helper.sh install-tools
```
daw3rd, line 72: can we have makefile targets that call these?  This would then allow the developer to run these tests locally, more easily and through mechanisms they are already familiar with (i.e. make).

daw3rd (changes requested):

.github/workflows/test-code-code_quality-kfp.yml
daw3rd: This would also force all the kfp tests to be run - they are not in the current PR .  See that only 3 checks were run.

revit13: Would it be okay to address this in a separate issue? Personally, I'm not in favor of this approach because this install-tools function works specifically for Ubuntu, as used in the GitHub Action. The installation process differs for other operating systems, like macOS, with its various versions. I believe it would be simpler for each person to deploy or use their own tools on their machine, rather than creating a new script that we would need to maintain. However, I could be wrong on this...

daw3rd (approved):

PR #685, opened 2024-10-08
Added ray-based version of hap transform

ian-cho:

Added ray-based version of hap transform to enable a ray runtime


transforms/universal/hap/ray/pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[project]
+name = "dpk_hap_transform_python"
```
touma-I:
```suggestion
name = "dpk_hap_transform_ray"
```
This may be causing most of the failure as it is conflict with the pure python module. We are working on combing both modules in a single package but for now, this should be dpk_hap_transform_ray

touma-I (changes requested):

ian-cho:
@touma-I thanks! I merged it. One previously failed test passed

transforms/universal/hap/ray/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+publish: publish-image
+
+publish-image:: .transforms.publish-image-ray
+
+setup:: .transforms.setup
+
+# distribution versions is the same as image version.
+set-versions:
+	$(MAKE) TRANSFORM_PYTHON_VERSION=$(HAP_PYTHON_VERSION) TOML_VERSION=$(HAP_RAY_VERSION) .transforms.set-versions
```
touma-I:
```suggestion
	$(MAKE) TRANSFORM_PYTHON_VERSION=$(HAP_PYTHON_VERSION) TOML_VERSION=$(HAP_PYTHON_VERSION) .transforms.set-versions 
```
@ian-cho : HAP_RAY_VERSION is not defined. We should use the same version for both PYTHON and RAY

touma-I: tagging @shahrokhDaijavad as he is tracking this closely. @ian-cho One more change that is needed. Please review and accept if you agree. I will do one more pass tomorrow morning.

touma-I (approved):
Thank you @ian-cho. Looks good now. All checks are passing- Meging this PR now. cc: @shahrokhDaijavad

shahrokhDaijavad:
Thank you, @ian-cho for doing this so quickly!

Issue #609, opened 2024-09-21
[Feature] Capability to specify the paths where multiple output tables will be saved


cmadam:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core

### Feature

In the context of Spark fuzzy dedup, we need to have a way to save the minhashes and band hashes calculated for each document in multiple output directories.  The result of running the signature calculation transform on a file located in the input folder at `input_path/relative_path/file_name`, must be mapped, in the output folder, to multiple files at locations: `output_path/bands/band=b/segment=s/relative_path/file_name`, where `b` and `s` are integers in the ranges `(0, _number_of_bands)` and `(0, number_of_segments)`, respectively. 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

cmadam:
@blublinsky : these are the hacks that provide to us this functionality now:
https://github.com/IBM/data-prep-kit/blob/2fb92f78b395ac53563b1885b6b78b2362de1075/data-processing-lib/python/src/data_processing/runtime/transform_file_processor.py#L167-L230
And we also need this logic in the transform (we are basically adding the paths where we want to save each table to the metadata):
https://github.com/IBM/data-prep-kit/blob/2fb92f78b395ac53563b1885b6b78b2362de1075/transforms/universal/fdedup/python/src/signature_calc_transform.py#L341-L354

blublinsky:
After looking carefully at this issue, there are 2 options for its resolution
1. Extend the File processor and transform API as suggested. This is a very invasive change, which will require changes in virtually any transform. The current implementation of the file processor is fairly light weight and meets the majority of the use cases and making it more complex for a single use case is not very appealing.
2. Implement saving of the custom files to the custom directories inside a transform itself. It makes a transform implementation a bit more complex, but keeps the framework lighter and keeps all the changes inside the transform itself.

In my opinion, the second approach is a much better solution.

Here is how this can be implemented. The base file processor adds current data access factory as a parameter to every transform - https://github.com/IBM/data-prep-kit/blob/dev/data-processing-lib/python/src/data_processing/runtime/transform_file_processor.py#L48
This means that a transform implementation can return no tables and do the writes itself.
[...]

PR #691, opened 2024-10-10
added folder_transform

blublinsky:
After several conversations with Constantin, it appears useful to have a transform for processing folders rather then file. 

https://github.com/IBM/data-prep-kit/issues/609
https://github.com/IBM/data-prep-kit/issues/705

### DIFF
--- a/data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
+++ b/data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
@@ -24,14 +24,13 @@ from data_processing.runtime.pure_python import (
     PythonTransformFileProcessor,
     PythonTransformRuntimeConfiguration,
 )
-from data_processing.transform import AbstractBinaryTransform, TransformStatistics, AbstractFolderTransform
+from data_processing.transform import AbstractTransform, TransformStatistics, AbstractFolderTransform
 from data_processing.utils import GB, get_logger
 
 
 logger = get_logger(__name__)
 
 
-@staticmethod
 def _execution_resources() -> dict[str, Any]:
     """
     Get Execution resource
@@ -47,6 +46,7 @@ def _execution_resources() -> dict[str, Any]:
         "memory": mused,
         "object_store": 0,
     }
+
 
 def orchestrate(
     data_access_factory: DataAccessFactoryBase,
@@ -76,7 +76,7 @@ def orchestrate(
     try:
         if is_folder:
             # folder transform
-            files = AbstractFolderTransform.get_folders(d_access=data_access)
+            files = runtime.get_folders(data_access=data_access)
             logger.info(f"Number of folders is {len(files)}")
         else:
             # Get files to process
@@ -145,7 +145,8 @@ def orchestrate(
             "job_input_params": input_params
             | data_access_factory.get_input_params()
             | execution_config.get_input_params(),
-            "execution_stats": _execution_resources() | {"execution time, min": round((time.time() - start_time) / 60.0, 3)},
+            "execution_stats": _execution_resources() |
+                               {"execution time, min": round((time.time() - start_time) / 60.0, 3)},
             "job_output_stats": stats,
         }
         logger.debug(f"Saving job metadata: {metadata}.")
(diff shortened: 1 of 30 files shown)


blublinsky:
It needs some tests, but this is a basic idea

data-processing-lib/python/src/data_processing/runtime/pure_python/transform_orchestrator.py
```diff
@@ -74,15 +72,21 @@ def orchestrate(
...
-        files, profile, retries = data_access.get_files_to_process()
-        if len(files) == 0:
-            logger.error("No input files to process - exiting")
-            return 0
-        if retries > 0:
-            statistics.add_stats({"data access retries": retries})
-        logger.info(f"Number of files is {len(files)}, source profile {profile}")
+        if is_folder:
+            # folder transform
+            files = AbstractFolderTransform.get_folders(d_access=data_access)
```
daw3rd: can this `get_folders()` be made part of DataAccess instead?  If not, then doesn't this have to be `runtime_config.get_transform_class().get_folders()`.

blublinsky: great minds, just moved it there

blublinsky: As for data access, discussed it with Constantin, and it is too application specific

cmadam: This is a great idea. `get_folders()` cannot be in the `DataAccess`, because sometimes we know exactly the set of folders that we want to process (also the number of levels of subfolders that we want to take into consideration), and we don't need `DataAccess` to produce a list of those folders. But putting `get_folders()` in the runtime config will work and will provide maximum flexibility, because we can either use `DataAccess` to retrieve specific files, or we can just provide a list of folders ourselves.

cmadam (commented):
Give me until tomorrow morning to try this `FolderTransform` with the cluster analysis transform in fuzzy dedup. If it works smoothly (and it should), I will approve this right away.

data-processing-lib/spark/src/data_processing_spark/runtime/spark/transform_orchestrator.py
```diff
@@ -92,13 +95,19 @@ def process_partition(iterator):
...
     try:
-        # Get files to process
-        files, profile, retries = data_access.get_files_to_process()
-        if len(files) == 0:
-            logger.error("No input files to process - exiting")
-            return 0
-        logger.info(f"Number of files is {len(files)}, source profile {profile}")
+        if is_folder:
+            # folder transform
+            files = AbstractFolderTransform.get_folders(d_access=data_access)
```
cmadam: Why does spark have a different logic? I think this line should be: 
```
            files = runtime_config.get_folders(d_access=data_access)
```
And the file `data-processing-lib/spark/src/data_processing_spark/runtime/spark/runtime_configuration.py` should contain the `get_folders()` function in the `SparkTransformRuntimeConfiguration` class:
```    
    def get_folders(self, data_access: DataAccess) -> list[str]:
        """
        Get folders to process
        :param data_access: data access
        :return: list of folders to process
        """
        raise NotImplemented()
```

cmadam (changes requested):

data-processing-lib/spark/src/data_processing_spark/runtime/spark/transform_orchestrator.py
blublinsky: Oops. Fixed

touma-I (dismissed):
@cmadam This PR does not seem to have an issue associated with it. Can you please create  an issue explaining what problem this PR is trying to solve and why we chose this approach to solve the problem,  and  an example on how this can be used by the transforms. Given that this is on the open source, we want to make sure other folks can consume the asset. Thanks

cmadam:
@touma-I : I have created Issue https://github.com/IBM/data-prep-kit/issues/705.

@blublinsky : can you please update the PR description, and associate it with Issue https://github.com/IBM/data-prep-kit/issues/705

blublinsky:
> @touma-I : I have created Issue #705.
> 
> @blublinsky : can you please update the PR description, and associate it with Issue #705

its linked to 609 and 705

cmadam (approved):
I have just used this code for fuzzy dedup, and was able to run correctly the fuzzy dedup code in python, spark and ray. In the context of fuzzy dedup, I have 4 transforms, 2 of those transforms (`ClusterAnalysisTransform` and `GetDuplicateListTransform`) are now inheriting from  `AbstractFolderTransform`, while the other two transforms(`SignatureCalculationTransform` and `DataCleaningTransform`) inherit from `AbstractTableTransform`.

Issue #667, opened 2024-10-04
[Bug] pdf2parquet ray version erroring out when downloading models for the very first time


sujee:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Tools/ingest2parquet

### What happened + What you expected to happen

Happens when running RAY version, with NUM_WORKERS > 1.
Reliably reproducible in google colab 
Running the cell again works.

But a **negative user experience**

```text
(orchestrate pid=1575) 05:41:45 ERROR - Failed to process request worker exception The actor died because of an error raised in its creation task, ray::RayTransformFileProcessor.__init__() (pid=1784, ip=172.28.0.12, actor_id=09c62ae6504057816b30599401000000, repr=<data_processing_ray.runtime.ray.transform_file_processor.RayTransformFileProcessor object at 0x7ee7e55fbc40>)
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/data_processing_ray/runtime/ray/transform_file_processor.py", line 46, in __init__
(orchestrate pid=1575)     self.transform = params.get("transform_class", None)(self.transform_params)
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/pdf2parquet_transform_ray.py", line 40, in __init__
(orchestrate pid=1575)     super().__init__(config)
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/pdf2parquet_transform.py", line 105, in __init__
(orchestrate pid=1575)     self._converter = DocumentConverter(
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/docling/document_converter.py", line 54, in __init__
(orchestrate pid=1575)     self.model_pipeline = pipeline_cls(
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/docling/pipeline/standard_model_pipeline.py", line 18, in __init__
(orchestrate pid=1575)     EasyOcrModel(
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/docling/models/easyocr_model.py", line 21, in __init__
(orchestrate pid=1575)     self.reader = easyocr.Reader(config["lang"])
(orchestrate pid=1575)   File "/usr/local/lib/python3.10/dist-packages/easyocr/easyocr.py", line 92, in __init__
[...]

PR #693, opened 2024-10-10
add MultiLock class

daw3rd:
This is needed to support inter-process and thread synchronization.  
With this, transforms can now run multi-process (via local ray actors, for example) and not get synchronization errors.

#667 will require pdf2parquet to make use of this lock.


dolfim-ibm (approved):
LGTM, thanks @daw3rd

PR #682, opened 2024-10-07
Single Packages (0.2.2.dev1) for data-prep-toolkit and data-prep-toolkit-transforms  with python3.12

touma-I:
## Why are these changes needed?

In this release, we create a single pypi package for data-prep-kit[dev,ray,spark] and a single data-prep-kit-transforms[all,ray,code_quality,doc_id,etc]

This replaces the multitude of packages we currently have and can be installed using:
pip install data-prep-toolkit==0.2.2.dev1
pip install 'data-prep-toolkit[ray]==0.2.2.dev1'
pip install 'data-prep-toolkit-transforms[ray, all]==0.2.2.dev1'

## Related issue number (if any).

#673

PR #698, opened 2024-10-11
Change the release script and documentation to bump the minor version instead of micro version.

daw3rd:
To allow intermediate transform and library releases that bump the micro version.
I have tested this to create the expected branches at the desired versions.

#639


touma-I (approved):
@daw3rd This Look's good. Once we merge #682 we will need to revisit this to force the transform release to resync with the library.

PR #702, opened 2024-10-13
Update test-kfp.yml workflow.

revit13:

This PR updates test-kfp.yaml workflow as was done for test-kfp-transform.template

### DIFF
--- a/.github/workflows/test-kfp.yml
+++ b/.github/workflows/test-kfp.yml
@@ -101,7 +101,7 @@ jobs:
                       set -- "${transforms[@]}" && transforms=("$@") && size=${#transforms[@]} && index=$(($RANDOM % $size))
                       transform=$(basename "${transforms[$index]}")
                       if [ -e "$transform/Makefile" -a -e "$transform/kfp_ray/Makefile" ] && echo ${KFP_BLACK_LIST} | grep -qv ${transform}; then
-                        header_text "Running ${transforms[$index]} workflow test"
+                        echo "Running ${transforms[$index]} workflow test"
                         break
                       fi
                   done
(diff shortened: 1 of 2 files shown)


.github/workflows/test-kfp.yml
```diff
@@ -86,41 +83,30 @@ jobs:
...
-                  chmod 700 /tmp/get_helm.sh
-                  HELM_INSTALL_DIR=/tmp/ /tmp/get_helm.sh -v v${HELM_VERSION} --no-sudo
-                  chmod 777 /tmp/helm
-                  curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /tmp/kubectl
-                  chmod 777 /tmp/kubectl
-                  curl https://dl.min.io/client/mc/release/linux-amd64/mc --create-dirs -o /tmp/mc
-                  chmod +x /tmp/mc
-                  export DEPLOY_KUBEFLOW=1
-                  make -C $K8S_SETUP_SCRIPTS setup
+                  KFP_BLACK_LIST=$(./scripts/check-workflows.sh -show-kfp-black-list)
```
daw3rd: Can you move the black list down to just before the while loop?  THis way I know that it is not being used by the next few lines (assuming that is the case).  Needs to be done in two places.

daw3rd (approved):

.github/workflows/test-kfp.yml
revit13: Done. Thanks

.github/workflows/test-kfp.yml
```diff
@@ -9,7 +9,6 @@ on:
         tags:
             - "*"
         paths:
-            - ".make.*"
```
daw3rd: kfp/Makefile includes .make.defaults, so this should be included.

daw3rd (changes requested):

.github/workflows/test-kfp.yml
revit13: Done. Thanks

daw3rd (approved):

Issue #703, opened 2024-10-14
[Bug] Github workflow fails


revit13:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

CI/CD

### What happened + What you expected to happen

`ubuntu:latest` seems to use now python3.12
the following errors occurs:
1)
https://github.com/IBM/data-prep-kit/actions/runs/11317492144/job/31470999796
```
../../../../data-processing-lib/python/src/data_processing/utils/pipinstaller.py:21: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
```

[pkg_resources is deprecated and removed in Python 3.12](https://github.com/mu-editor/mu/issues/2485)

2)  **FIXED**  https://github.com/IBM/data-prep-kit/actions/runs/11322510497/job/31483339004 
```
pip install jinja2
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

```

### Reproduction script

Run the workflows

### OS

Ubuntu

### Python

Other

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

revit13:
Trying to fix the first error by using importlib.metadata instead of pkg_resources I hit another error which relates to [python3.12](https://github.com/miguelgrinberg/python-socketio/issues/1265):
```
FAILED data_processing_tests/data_access/data_access_local_test.py::TestReadPyarrowTable::test_error_handling[malformed_parquet.parquet-ArrowException] - AttributeError: 'called_once_with' is not a valid assertion. Use a spec for the mock if 'called_once_with' is meant to be an attribute.. Did you mean: 'assert_called_once_with'?
```

revit13:
ubuntu-22.04 is now used in the github actions so python3.10 is used and the ci/cd is passing ok.

PR #704, opened 2024-10-14
Fix workflow failures due to usage of pip

revit13:
This PR aims to resolve the recent GitHub Action failure, which occurred due to the use of pip instead of the pip within a virtual environment (venv):
```
pip install jinja2
error: externally-managed-environment

× This environment is externally managed
```

partally solves #703

### DIFF
--- a/scripts/workflow_helper.sh
+++ b/scripts/workflow_helper.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -euo pipefail
+
 op=$1
 
 install_tools(){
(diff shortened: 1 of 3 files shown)


scripts/workflow_helper.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -euo pipefail
```
roytman, line 7: Please add a comment with some explanation
Maybe we should add in other places.

roytman (approved):

scripts/workflow_helper.sh
revit13: Done. Thanks

PR #709, opened 2024-10-14
Fix missing dev1 in README.md pip install command

touma-I:

Fix typo in README.md file as it was referencing a release version that does not yet exists. Set the tag to the current dev version: 0.2.2.dev1


shahrokhDaijavad (approved):
LGTM

Issue #697, opened 2024-10-11
Publish data-connector-lib to PyPI


hmtbr:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

There is a request from a consumer of the data-connector-lib that he would like to install the lib from PyPI by pip: `pip install dpk_connector`. We need to publish the module there.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

touma-I:
As discussed, the release package name will be data-prep-connector and the initial release tag will be aligned with the current DPK release tag: data-prep-connector==0.2.2 (with the initial test release being 0.2.2.dev0 )

hmtbr:
@touma-I Thanks for the review on my PR. Closing this as completed.

PR #712, opened 2024-10-16
workflow to publish data-connector-lib

hmtbr:

To address the request below, we want to publish the data-connector-lib with the project name `data_prep_connector`. Actually, the publish jobs are disabled following the other modules, but I would like to add it for the data-connector-lib for future changes towards automatic release management.

#697


touma-I (approved):
@hmtbr I don't see any issue with adding this to the repo. the name change is aligned with what we agreed to with the users. Also, currently, we are doing all the releases manually. Once we get this  initial testing period, I can see something like this to be helpful in the future.

PR #715, opened 2024-10-16
add license/copyright as appropriate to .py files and add check-licensing.sh script

daw3rd:
Need license headers on the code in the repo.
Also includes ci/cd check on license headers using new scripts/check-licensing.sh


dolfim-ibm (approved):
👍

matouma (approved):
@daw3rd Should we update you branch from dev before we do a merge of the PR ?

PR #717, opened 2024-10-16
Update README.md by adding Code_pofiler transform to the table

shahrokhDaijavad:
The new transform code_profiler has been added to the same table (both Python and Ray versions are available).


touma-I (dismissed):
Please hold off until the full implementation is complete. Right now we only have the syntactic profiler

shahrokhDaijavad:
html2parquet KFP has been added to the README, but Code_profiler is not ready.

Issue #719, opened 2024-10-17
[Bug] docid ray transformation errors when running on colab (release 0.2.2dev1)


sujee:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/universal/doc_id

### What happened + What you expected to happen

Ray version of docid transformation fails on colab with the following error

[full error](https://gist.github.com/sujee/2c39ffbb422ce3e3a3ef6f264aa8ee96)

```
(orchestrate pid=5751) created [Actor(RayTransformFileProcessor, 37cbb100b21c9012ce68444101000000), Actor(RayTransformFileProcessor, e063fda8d2ea6f56efc152e201000000)], alive [ActorState(actor_id='37cbb100b21c9012ce68444101000000', class_name='RayTransformFileProcessor', state='ALIVE', job_id='01000000', name='', node_id='08deddebac466e0100f6c1015c266849f52c8ff173595c8170cdee12', pid=5915, ray_namespace='aa965802-2b25-4b0b-83c8-3e939d83ab34', serialized_runtime_env=None, required_resources=None, death_cause=None, is_detached=None, placement_group_id=None, repr_name=None)]
(orchestrate pid=5751) Traceback (most recent call last):
(orchestrate pid=5751)   File "/usr/local/lib/python3.10/dist-packages/data_processing_ray/runtime/ray/transform_orchestrator.py", line 89, in orchestrate
(orchestrate pid=5751)     processors = RayUtils.create_actors(
(orchestrate pid=5751)   File "/usr/local/lib/python3.10/dist-packages/data_processing_ray/runtime/ray/ray_utils.py", line 121, in create_actors
(orchestrate pid=5751)     raise UnrecoverableException(f"out of {len(actors)} created actors only {len(alive)} alive")
(orchestrate pid=5751) data_processing.utils.unrecoverable.UnrecoverableException: out of 2 created actors only 1 alive
(orchestrate pid=5751) 05:26:08 ERROR - Exception during execution out of 2 created actors only 1 alive: None
(orchestrate pid=5751) Traceback (most recent call last):
(orchestrate pid=5751)   File "/usr/local/lib/python3.10/dist-packages/data_processing_ray/runtime/ray/transform_orchestrator.py", line 138, in orchestrate
(orchestrate pid=5751)     stats["processing_time"] = round(stats["processing_time"], 3)
[...]

PR #721, opened 2024-10-17
Fix metadata logging even when actors crash 

shivdeep-singh-ibm:

In case the ray actors crash, the key `processing_time` may not be updated to stats, so while rounding off this key, we may get an exception which stops metadata from being written.

Also seen in the crash here. https://github.com/IBM/data-prep-kit/issues/719

### DIFF
--- a/data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_orchestrator.py
+++ b/data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_orchestrator.py
@@ -62,7 +62,7 @@ def orchestrate(
         if is_folder:
             # folder transform
             files = runtime.get_folders(data_access=data_access)
-            logger.info(f"Number of folders is {len(files)}")        # Get files to process
+            logger.info(f"Number of folders is {len(files)}")  # Get files to process
         else:
             files, profile, retries = data_access.get_files_to_process()
             if len(files) == 0:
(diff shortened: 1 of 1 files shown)


data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_orchestrator.py
```diff
@@ -142,7 +142,8 @@ def orchestrate(
         # Compute execution statistics
         logger.debug("Computing execution stats")
         stats = runtime.compute_execution_stats(ray.get(statistics.get_execution_stats.remote()))
-        stats["processing_time"] = round(stats["processing_time"], 3)
+        if "processing_time" in stats:
```
daw3rd, line 145: should we set this to 0 if not found in case downstream code depends on it?

daw3rd (changes requested):

data-processing-lib/ray/src/data_processing_ray/runtime/ray/transform_orchestrator.py
shivdeep-singh-ibm: I think, value 0 for processing_time can be misleading.

daw3rd: So a missing key downstream means what?  I suspect this really means we're not properly handling the crashing of the actor and should indicate that in some other way, perhaps with a graceful shutdown.

blublinsky: The only way `stats["processing_time"] = round(stats["processing_time"], 3)` can throw an exception is in the case where either:
1. init method in file processor fails due to inability to create transform
2. All of the transform invocations failed and "processing time" was never reported by file processor

In both cases, there is really no statistics of interest to be collected. 
But if you really want to avoid this exception the right thing to do (in my opinion) is to do what Shivdeep is doing:

```
      if "processing_time" in stats:
            stats["processing_time"] = round(stats["processing_time"], 3)
```
This will return metadata with no processing time value, which is correct in this case

Note that if you want to go for this fix, please fix it in all 3 orchestrators - Python, Ray and Spark. Not just Ray

daw3rd (approved):

Issue #724, opened 2024-10-18
[Feature] [Connector] Apply subdomain focus based on the seed url


hmtbr:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

If the user provides https://research.example.com as a seed url for the data-prep-connector, there is a requirement that the user wants to automatically apply subdomain focus so we do not crawl other subdomains than research for the domain example.com.

The user may also provide multiple seed urls like
corporate.example.com
research.example.com

in which case we would like to crawl only two subdomains - corporate and research

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

jo-in:
@hmtbr Can you also confirm the behavior when you specify a seed URL like https://arxiv.org/year/astro-ph/2024 ? Will it be able to download papers from https://arxiv.org/list/astro-ph/2024-01 (note that the path starts with `/path` and not `/year`) or only those URLs starting with the exact seed URL ?

hmtbr:
@jo-in Yes, it can download the urls with different paths as such unless you specify `path_focus=True` (defaults to `False`).

PR #725, opened 2024-10-18
implement subdomain focus feature in data-prep-connector

hmtbr:

If the user provides https://research.example.com/ as a seed url for the data-prep-connector, there is a requirement that the user wants to automatically apply subdomain focus so we do not crawl other subdomains than research for the domain example.com.

This PR implements the subdomain focus feature.

#724

### DIFF
--- a/data-connector-lib/src/dpk_connector/core/crawler.py
+++ b/data-connector-lib/src/dpk_connector/core/crawler.py
@@ -74,6 +74,7 @@ def async_crawl(
     user_agent: str = "",
     headers: dict[str, str] = {},
     allow_domains: Collection[str] = (),
+    subdomain_focus: bool = False,
     path_focus: bool = False,
     allow_mime_types: Collection[str] = (
         "application/pdf",
@@ -96,6 +97,7 @@ def async_crawl(
         user_agent (str): The user agent string to use for the crawler. Defaults to "Scrapy/VERSION (+https://scrapy.org)".
         headers (dict[str, str]): A dictionary of additional headers to send with each request. Default is an empty dictionary.
         allow_domains (Collection[str]): A collection of domains to restrict the crawler to. Default is the domains of the seed URLs.
+        subdomain_focus (bool): If specified, only links under the subdomains of the input seed URLs will be extracted. Ignored if `allow_domains` is specified.
         path_focus (bool): If specified, only links under the paths of the input seed URLs will be extracted.
         allow_mime_types (Collection[str]): A collection of MIME types to allow during the crawl. Default is a collection containing "application/pdf", "text/html", "text/markdown", and "text/plain".
         disallow_mime_types (Collection[str]): A collection of MIME types to disallow during the crawl. Default is an empty collection.
@@ -140,6 +142,7 @@ def async_crawl(
         seed_urls=seed_urls,
         callback=on_downloaded,
         allow_domains=allow_domains,
+        subdomain_focus=subdomain_focus,
         path_focus=path_focus,
         allow_mime_types=allow_mime_types,
         disallow_mime_types=disallow_mime_types,
@@ -155,6 +158,7 @@ def crawl(
     user_agent: str = "",
     headers: dict[str, str] = {},
     allow_domains: Collection[str] = (),
+    subdomain_focus: bool = False,
     path_focus: bool = False,
     allow_mime_types: Collection[str] = (
         "application/pdf",
@@ -177,6 +181,7 @@ def crawl(
         user_agent (str): The user agent string to use for the crawler. Defaults to "Scrapy/VERSION (+https://scrapy.org)".
         headers (dict[str, str]): A dictionary of additional headers to send with each request. Default is an empty dictionary.
         allow_domains (Collection[str]): A collection of domains to restrict the crawler to. Default is the domains of the seed URLs.
+        subdomain_focus (bool): If specified, only links under the subdomains of the input seed URLs will be extracted. Ignored if `allow_domains` is specified.
         path_focus (bool): If specified, only links under the paths of the input seed URLs will be extracted.
         allow_mime_types (Collection[str]): A collection of MIME types to allow during the crawl. Default is a collection containing "application/pdf", "text/html", "text/markdown", and "text/plain".
         disallow_mime_types (Collection[str]): A collection of MIME types to disallow during the crawl. Default is an empty collection.
@@ -198,6 +203,7 @@ def crawl(
         user_agent,
         headers,
         allow_domains,
+        subdomain_focus,
         path_focus,
         allow_mime_types,
         disallow_mime_types,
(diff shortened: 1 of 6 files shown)


matouma:
@qiragg Please confirm that you can see this PR and comment on it. You can also tag me once your you approve it. I am also soliciting input from the broader community on this one. I know we did it before in the first part of the year and I want to make sure we capture lessons learned from the previous implementation (what worked and what did not work).

data-connector-lib/src/dpk_connector/core/crawler.py
```diff
@@ -74,6 +74,7 @@ def async_crawl(
     user_agent: str = "",
     headers: dict[str, str] = {},
     allow_domains: Collection[str] = (),
+    subdomain_focus: bool = False,
```
matouma, line 77: Should the default be set to True? Its is very rare that we need to do a common crawl of the web and not be restricted/focused on a specific domain or a specific URL...

data-connector-lib/src/dpk_connector/core/spiders/sitemap.py
```diff
@@ -88,11 +90,15 @@ def __init__(
                     self.focus_paths.add(path)
 
         # Domains and mime types filtering
-        self.allowed_domains = set(
-            allow_domains
-            if len(allow_domains) > 0
-            else [get_etld1(url) for url in seed_urls]
```
matouma, line 94: Let's say the URL is : http://www.research.ibm.com and domain focus is set to true. Will this restrict to *.research.ibm.com or will it restrict to *.ibm.com ?

matouma (commented):
@hmtbr would be great if you can review my comments and let me know your thoughts on how this should work.

touma-I:
@qiragg @yuanchi2807 Please review and comment as you see fit. @qqirag it may help if you can elaborate further on the rated issue based on previous experience with similar functionality in bluecrawl

data-connector-lib/src/dpk_connector/core/crawler.py
hmtbr: If we set it true by default, the default behavior the crawler would change. We should not introduce any braking change except for major version updates.

data-connector-lib/src/dpk_connector/core/spiders/sitemap.py
hmtbr: Note that the parameter I introduced this time is "sub"domain_focus. The crawler focuses the input domain by default: the seed url: https://www.research.ibm.com/ -> the crawl will be restricted to *.ibm.com
If `subdomain_focus` is set to true, the crawl will be restricted to *.www.research.ibm.com.

yuanchi2807:
> @Qiragg @yuanchi2807 Please review and comment as you see fit. @qqirag it may help if you can elaborate further on the rated issue based on previous experience with similar functionality in bluecrawl

I am not knowledgable enough to crawling requirements to make a comment.

data-connector-lib/src/dpk_connector/core/spiders/sitemap.py
touma-I: Nice! Thanks for the clarification @hmtbr -san

data-connector-lib/src/dpk_connector/core/crawler.py
touma-I: ok. that makes sense. I have the feeling that for most of the RAG use cases we will be setting this to true.

Qiragg:
> @Qiragg @yuanchi2807 Please review and comment as you see fit. @qqirag it may help if you can elaborate further on the rated issue based on previous experience with similar functionality in bluecrawl

In bluecrawl, we do provide the ability to do subdomain_focus automatically based on the input seed url but we cannot focus on multiple subdomains per job. For crawling, it would make sense to be able to launch a single job that focuses on multiple subdomains which is what this feature would provide. This is a functionality that was missing in DPK-connector and is much needed for launching certain targeted crawls.

There are a couple of points to discuss here:
1. It is more intuitive to set the default subdomain_focus to be true but we will not do that here based we do not want to depart from the default crawler behavior in the earlier version.
2. We expect the user to not provide `ibm.com` and `research.ibm.com` as seed urls at the same time if they want subdomain_focus to be applied to on `research`. It is not clear if we should reject jobs which are improperly configured in such a way or allow the user to rectify their mistake. While this is obvious in this case; in rare cases, the user may want to apply path_focus and subdomain_focus at the same time by providing: `research.ibm.com/help/` and `ibm.com/support/` as seeds. In such a case, my understanding is that they will end up crawling both `ibm.com/help/` and `ibm.com/support/`.

Ideally, we only want to crawl `research.ibm.com/help/` and `ibm.com/support/` for such a user-defined case.

The PR looks good to me for now. I think if we get feedback regarding a different design choice that the user wants, we can think about it at that point.

touma-I:
@Qiragg Thanks for the analysis. Good stuff: Can you click on "Files Changed" then click on the green button "review changes" then check "approve" then click submit to indicate your approval. 
@hmtbr: 
1- what happens if we use the seed url : http://www.research.ibm.com/ and set path focus to true (i.e. path focus is for the Root  (i.e. / ), will that produce the same result as sub-domain focus ?
2- Also,  what if the user specifies the "unexpected" in the case that @Qiragg identified above (ie. 2 seeds URLs  ibm.com and research.ibm.com and subdomain focus set to true? Will the code default  to the least restrictive or the more restrictive case ?

touma-I (changes requested):
Need to update pyproject.toml to reflect 0.2.2.dev1. (This will be the work-in-progress tag until it is released to pypi)

hmtbr:
> 1- what happens if we use the seed url : http://www.research.ibm.com/ and set path focus to true (i.e. path focus is for the Root (i.e. / ), will that produce the same result as sub-domain focus ?

Currently there is no support for path focus with subdomain. It will produce pages from domain `*.ibm.com`.

> 2- Also, what if the user specifies the "unexpected" in the case that @Qiragg identified above (ie. 2 seeds URLs ibm.com and research.ibm.com and subdomain focus set to true? Will the code default to the least restrictive or the more restrictive case ?

It will produce pages from domain `*.ibm.com`.

Qiragg (approved):

touma-I:
> > 1- what happens if we use the seed url : http://www.research.ibm.com/ and set path focus to true (i.e. path focus is for the Root (i.e. / ), will that produce the same result as sub-domain focus ?
> 
> Currently there is no support for path focus with subdomain. It will produce pages from domain `*.ibm.com`.
> 

@hmtbr @Qiragg @shahrokhDaijavad is this a bug in the logic then that need to be fixed ? If the application specifies **www.ibm.com/docs** with **path focus** it should not receive anything from **research.ibm.com/docs** . No?

touma-I (approved):

hmtbr:
@touma-I Thanks for your approval. The cases don't produce intuitive results actually. I will address them in another issue and PR soon. This PR itself won't introduce any regression, so I merge this now to unblock our work.

touma-I:
> @touma-I Thanks for your approval. The cases don't produce intuitive results actually. I will address them in another issue and PR soon. This PR itself won't introduce any regression, so I merge this now to unblock our work.

Thanks @hmtbr Agree. Now that this has matured to this point, it is appropriate to consider usability by outsiders who are not intimately familiar with our internal processes. I do think that if path focus is done right, we won't need subdomain focus and It will be appropriate to do a second iteration. thanks again.

PR #733, opened 2024-10-23
Update release number following release cutoff for http connector

touma-I:

After the release of http connector 0.2.2, need to bump up the patch and add dev0 to it


data-connector-lib/Makefile
```diff
@@ -2,6 +2,8 @@
 REPOROOT=..
 include $(REPOROOT)/.make.defaults
 
+DPK_CONNECTOR_VERSION=0.2.2
```
touma-I: this should be moved to .make.versions and incremented to 0.2.3.dev0 before re-running set-versions

hmtbr (approved):
Thanks. Looks good for this release.

Issue #739, opened 2024-10-24
[Feature] Example notebook for DPK connector


Qiragg:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

We do not have an example notebook demonstrating the use of DPK-connector for doing targeted crawls to support RAG. We would like to provide such an example.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

touma-I:
@Qiragg and @sujee to come up with a notebook and submit a PR

touma-I:
Replace with https://github.com/IBM/data-prep-kit/issues/751

PR #740, opened 2024-10-24
Add a notebook demonstrating the use of DPK connector for RAG

Qiragg:

These changes will allow us to demonstrate an end-to-end pipeline starting from the target acquisition of crawled content.

This PR will also update the requirements.txt and utils to support certain functions used during the acquisition.

The notebook uploaded demonstrates the download of only research papers (PDF files) published in NeurIPS 2017 using path_focus and mime_type extraction. The crawled PDFs can be fed further in our RAG pipeline and the rest of the steps remain the same starting from Step 2.2 of https://github.com/IBM/data-prep-kit/blob/dev/examples/notebooks/rag/rag_1A_dpk_process_python.ipynb

The only issue I see is somewhere in the notebook : https://github.com/IBM/data-prep-kit/blob/dev/examples/notebooks/rag/rag_2B_llamaindex_query.ipynb we do queries on the attention mechanism and the granite model. Papers pertaining to those are not downloaded in the notebook but I can make changes to have them be downloaded during the crawl. If I make the change and don't limit the number of downloads, around ~700 PDFs will get downloaded. I don't think it will cause hiccups during the rest of RAG steps but I haven't tested that yet.

Let's use this PR to discuss how can we this notebook fit with the rest of the RAG pipeline and whether any further changes are needed.

https://github.com/IBM/data-prep-kit/issues/739

### DIFF
--- a/examples/notebooks/rag/requirements.txt
+++ b/examples/notebooks/rag/requirements.txt
@@ -2,7 +2,7 @@
 
 data-prep-toolkit-transforms==0.2.1
 data-prep-toolkit-transforms-ray==0.2.1
-
+data-prep-connector==0.2.2.dev1
 deepsearch-toolkit
 
 
@@ -42,6 +42,7 @@ llama-index-vector-stores-milvus
 # --- Utils
 python-dotenv==1.0.0
 humanfriendly
+mimetypes-magic
 
 ## --- Jupyter Utils
 jupyterlab
@@ -51,3 +52,4 @@ ipywidgets
 IProgress
 chardet==5.2.0
 charset-normalizer==3.3.2
+nest-asyncio
--- a/examples/notebooks/rag/rag_1A_dpk_connector_python.ipynb
+++ b/examples/notebooks/rag/rag_1A_dpk_connector_python.ipynb
@@ -1,0 +1,270 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Before Running the notebook\n",
+    "\n",
+    "Please complete [setting up python dev environment](./setup-python-dev-env.md)"
+   ]
+  },
+  {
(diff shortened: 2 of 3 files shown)


Qiragg:
I noticed that there is a minor bug in the PR. 

This is being executed

```
    if saved_pages<20:
        print(f"Visited url: {url}")
```

even when the visited url count is greater than 20. I will fix that in the next commit after I get feedback on the notebook.

shahrokhDaijavad:
Thanks, @Qiragg. Nice job! I tested your branch by running the Notebook and got the expected output. 58 pages were retrieved, and 18 were downloaded to the input subdirectory. Together with the "attention" and "granite" papers, it is the 20-limit bug that you have noted above. This input directory is now ready to be used by @sujee's process Notebook. Of course, the top part of Notebook (overview section, including the picture) needs to be changed.

examples/notebooks/rag/requirements.txt
```diff
@@ -2,7 +2,7 @@
 
 data-prep-toolkit-transforms==0.2.1
 data-prep-toolkit-transforms-ray==0.2.1
-
+data-prep-connector==0.2.2.dev1
```
hmtbr, line 5: We may want to use the official release version.

```suggestion
data-prep-connector==0.2.2
```

hmtbr (commented):
Good job. One minor comment.

sujee:
My feedback is 

- This is a great utility!
- move it out of the RAG example (I prefer to keep that example simple)
- And release it as another example, downloading HTML files and feeding it to html2pq pipeline.

shahrokhDaijavad:
OK, @sujee. I appreciate that you want to keep the RAG example simple. @Qiragg We can discuss what to call the example, put it parallel to the RAG directory and what to add (pdf2parquet, html2parquet alone or add some other transform) to do something meaningful after the crawl.

PR #742, opened 2024-10-24
Investigate issue with hap failure

touma-I:

Investigate HAP kfp failure


touma-I:
Close PR without Merge. Image file is too big for kfp test

PR #746, opened 2024-10-26
Added exception handling to code quality transform

Param-S:
code quality transform failing when input data length is zero


agoyal26:
@Param-S is this still needed? Can you please ask Shivdeep to review - thanks

Issue #605, opened 2024-09-20
[Bug] pdf2parquet  must calculate hash and size on the file


sujee:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Tools/ingest2parquet

### What happened + What you expected to happen

I had duplicate documents (see attached).
I was expecting the exact same duplicate files to have same size and hash.
But seems like the hash is being calculated on 'contents' which is actual content + meta data (like file name)

I think the hash and size should be calculated on the actual file/document not on meta data.

![image](https://github.com/user-attachments/assets/26b91ae9-d69b-4cbe-a850-b24983e93491)

## Expected Behaviour

- `hash` should be identical to identical files
- `size` should be physical file size in bytes
- to avoid confusion, these columns can be renamed (or new columns can be created) with names like  `file_hash`  and `file_size`

### Reproduction script

[earth.pdf](https://github.com/user-attachments/files/17072188/earth.pdf)

Create a copy of the above file

execute the pdf2parquet section here : https://github.com/sujee/data-prep-kit-examples/blob/main/dpk-intro/dpk_intro_1_python.ipynb

### OS

Ubuntu

### Python

3.11.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

dolfim-ibm:
At the moment the `hash` column contains the hash of the actual `contents` column. This is the JSON representation of the output, which has the property `file-info.filename`, so different filenames will have different content.

Internally, the JSON has a property `file-info.document-hash` which is the actual hash of the binary input file.

It could indeed make sense to expose that one as well. Where? Should it be the `document_id`? Another field? Happy for an open discussion here.

sujee:
I do see `document_hash` in the contents.

I would like to see this propagated up as a top-level column in the output parquet.  Along with actual file size.

![image](https://github.com/user-attachments/assets/31ca764a-0a98-415e-81d0-056f7a922602)

sujee:
@dolfim-ibm with the new Docling integration, will this be addressed as well?

dolfim-ibm:
Reading again above, there were some open questions about which field to expose and with which names. The fact of exposing both is for sure a good idea, since they serve different purposes.

dolfim-ibm:
Should be fixed in https://github.com/IBM/data-prep-kit/pull/756.

Bytes-Explorer:
[...]

Issue #573, opened 2024-09-05
[Bug] improve performance of pdf2parquet


sujee:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### What happened + What you expected to happen

Extracting text from PDF into parquet seems slow.  It is processing 1 page / second.  So if a PDF has 300 pages, it takes 300 seconds (5 mins)

This negatively affects the user experience, as PDF2PQ is usually one of first few steps in many workflows.

### Reproduction script

data : https://github.com/sujee/data-prep-kit/tree/perf-1-pdf2pq/test/perf-pdf2pq/input
(These PDFs are about 100 pages each)

Instructions and minimal code to reproduce the problem are here : https://github.com/sujee/data-prep-kit/tree/perf-1-pdf2pq/test/perf-pdf2pq

instructions (README.md) : https://github.com/sujee/data-prep-kit/blob/perf-1-pdf2pq/test/perf-pdf2pq/README.md

A py-spy generated speedscope file is attached.  It can be viewed at https://www.speedscope.app/ 

[test_pdf2pq_py.speed.txt](https://github.com/user-attachments/files/16885512/test_pdf2pq_py.speed.txt)

### OS

Ubuntu

### Python

3.11.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Bytes-Explorer:
@dolfim-ibm Can you share any updates on this pls?
cc @touma-I

dolfim-ibm:
Soon we will update DPK to use the new Docling v2. As part of the new feature (together with support for docx, html, pptx, etc) we have a new parse which is about 10x faster. See https://github.com/DS4SD/docling-parse/?tab=readme-ov-file#performance-benchmarks. Note, this is not the speed up of the full pipeline, but one of the important pieces.

Medium term, we are actually running heavy benchmarks to identify the characteristic timing and compare with other tools.

Bytes-Explorer:
Thank you @dolfim-ibm When do you expect to integrate this change?

dolfim-ibm:
> Thank you @dolfim-ibm When do you expect to integrate this change?

Should be doable this week.

Bytes-Explorer:
Thanks!

From: Michele Dolfi ***@***.***>
Date: Monday, 28 October 2024 at 8:05 PM
To: IBM/data-prep-kit ***@***.***>
[...]

PR #756, opened 2024-10-30
Update pdf2parquet to Docling v2

dolfim-ibm:

The new Docling v2 allows to
1. Process more input formats: PDF, DOCX, PPTX, HTML, Markdown, ASCII Docs
2. Faster PDF backend
3. Improvements in the generated document
4. New DoclingDocument
5. Additional export format to plain text

In progress
- [x] upgrade dependencies and adapt code
- [x] add safe-guard for the download of model weights on multi-processes
- [x] implement flushing mechanism for creating batches of files
- [x] update doc_chunk tansform
- [x] propagate parameters to kfp

- expose document hash of the input document. Refs https://github.com/IBM/data-prep-kit/issues/605
- use new faster PDF backend. Refs https://github.com/IBM/data-prep-kit/issues/573 
- safe-guard the download of model weights on multi-processes. Refs https://github.com/IBM/data-prep-kit/issues/667


daw3rd:
I believe with the use of MultiLock this also fixes #667

transforms/language/doc_chunk/python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dpk_doc_chunk_transform_python"
-version = "0.2.2.dev1"
+version = "0.3.0"
```
daw3rd, line 3: not sure how we handle these version suffixes on transforms now.  But since you depend on a dev1 version of the core library, should this also be a dev1 suffix?

dolfim-ibm: @touma-I which version should be used? I followed your advice here, but I can also change it.

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
```diff
@@ -26,7 +26,6 @@
 content_column_name_key = "content_column_name"
 doc_id_column_name_key = "doc_id_column_name"
 chunking_type_key = "chunking_type"
-dl_min_chunk_len_key = "dl_min_chunk_len"
```
daw3rd, line 29: I guess this is not backwards compatible with previous notebooks, scripts, etc?

transforms/language/pdf2parquet/python/requirements.txt
```diff
@@ -1,6 +1,6 @@
 data-prep-toolkit==0.2.2.dev1
```
daw3rd, line 1: Should this be >= 0.2.2.dev1.  dev1 may eventually go away.

transforms/language/pdf2parquet/python/src/pdf2parquet_transform.py
```diff
@@ -93,18 +138,75 @@ def __init__(self, config: dict):
...
+        if not isinstance(self.pdf_backend_name, pdf2parquet_pdf_backend):
+            self.pdf_backend_name = pdf2parquet_pdf_backend[self.pdf_backend_name]
         self.double_precision = config.get(
             pdf2parquet_double_precision_key, pdf2parquet_double_precision_default
         )
 
         logger.info("Initializing models")
-        pipeline_options = PipelineOptions(
+        pipeline_options = PdfPipelineOptions(
+            artifacts_path=self.artifacts_path,
```
daw3rd, line 161: Is it ok for artifacts_path to be None?  seems possible.

daw3rd (changes requested):
Do any of. your tests exercise self.buffer and flush_binary()?

transforms/language/doc_chunk/python/pyproject.toml
touma-I: @dolfim-ibm you are ok with what you have. The release process will fix all this...

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
dolfim-ibm: indeed. the parameter doesn't exist anymore in the library.
if you think it is better to keep it in the arguments and simply ignore it, we could add it again.

transforms/language/pdf2parquet/python/requirements.txt
dolfim-ibm: here I rely on @daw3rd  and @touma-I expertise. let me know what is the correct value to use

transforms/language/pdf2parquet/python/src/pdf2parquet_transform.py
dolfim-ibm: yes, it is actually the default value.

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
daw3rd: does this old parameter relate in anyway to the new parameters and could it be mapped somehow to the new parameter value (with a WARNING message of some sort).  Or, is this function no longer supported?  If the latter, then I would lean towards accepting it and warning about it so we don't break old notebooks?  I'm more worried about what example notebooks we have in the repo though, so if none use it, less of a problem for me to get rid of.

transforms/language/pdf2parquet/python/requirements.txt
daw3rd: @touma-I I defer to you on this one.

transforms/language/doc_chunk/python/src/doc_chunk_transform.py
dolfim-ibm: The parameter does not exists at all anymore. The new behavior, is sort of just covering what was needed.

On the other hand, for the notebook failures. Do DPK launchers complain if an argument is not existing? Or it is simply ignored?

dolfim-ibm: I did a few tests (with and without). I think it should be covered now with the warning.

daw3rd (approved):
LGTM modulo @touma-I's call on the >= versioning.

transforms/language/pdf2parquet/python/requirements.txt
touma-I: @dolfim-ibm @daw3rd this will be reset during the release process or whenever we push to pypi. for now, I wouldn't worry about it. You're good.

Issue #761, opened 2024-10-31
[Feature] Create new pre-release 0.2.2.dev2 with recent code for the transformers


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

Current pre-release version 0.2.2.dev1 does not include the ray implementation for hap and a number of outher improvements for other  transforms

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

touma-I:
Additional enhancements to the library will most likely require a new release 0.2.2.dev2 of the library as well.

touma-I:
@daw3rd @santoshborse I will be sumbitting a PR for your review/approval. Once approved, I will run:
```
make -C transforms/ build-pkg-dist && make -C transforms/ publish-dist 
make -C data-processing-lib/ build-pkg-dist && make -C data-processing-lib/ publish-dist 
```
This will build the wheels and push them to pypi.

touma-I:
Include latest pdf2parquet and docchunk with docling v2

touma-I:
Current pre-release 0.2.2.dev2 is now in pypi. Closing this issue.

PR #765, opened 2024-10-31
Create new dev2 pre-releases for both tansforms and library with latest from docling 2.0

touma-I:

Build and deploy a pre-release dev2 version for both transforms and library. While most recent requirements are only for a new release of the transforms, the library has undone several improvements that may be required.

https://github.com/IBM/data-prep-kit/issues/761


touma-I:
@daw3rd @santoshborse if you get a chance, please do a review and approve as appropriate. Lots of files have been changed but only because set-versions needs to update the pyprojects.toml and requirements.txt for all transforms. There were a lot of changes to the library as well so rebuilding and publishing both library and transforms.

transforms/transform.config
```diff
@@ -14,4 +14,4 @@ TRANSFORM_NAME=data-prep-kit-transforms
 #
 # If you change the versions numbers, be sure to run "make set-versions" to 
 # update version numbers across the transform (e.g., pyproject.toml).
-TRANSFORMS_PKG_VERSION=$(DPK_VERSION)
+TRANSFORMS_PKG_VERSION=0.2.2.dev2
```
daw3rd, line 17: Does `make set-versions` update this or is this a hand-edit?  If a hand-edit, wondering why not just stay with $(DPK_VERSION) or some other globally defined version that is used by `make set-versions`.

transforms/language/doc_chunk/transform.config
```diff
@@ -14,7 +14,7 @@ TRANSFORM_NAME=doc_chunk
 #
 # If you change the versions numbers, be sure to run "make set-versions" to 
 # update version numbers across the transform (e.g., pyproject.toml).
-DOC_CHUNK_PYTHON_VERSION=$(DPK_VERSION)
+DOC_CHUNK_PYTHON_VERSION=0.3.0
```
daw3rd, line 17: While I understand he wants to set his own versions, how do we update this (not by hand) when we do a major upgrade.  Do we need 2 `set-version` targets or will `set-versions` update this?

daw3rd (changes requested):

transforms/transform.config
touma-I: @daw3rd this follows a similar pattern to the other transforms where each transform can define its own revision tag. Our starting point was DPK_VERSION for all the transforms but expect to diverge going forward

transforms/language/doc_chunk/transform.config
touma-I: We may need to traverse all the transform.config and bump up the last digit or append .devX to it when need to.

daw3rd: ok, kick the can down the road :)

daw3rd (approved):

Issue #753, opened 2024-10-29
Uniform documentation and example Notebooks for all transforms! 


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

This is a "super-issue" that will affect all transforms! Each transform owner will be assigned to do two tasks, for the transform they won:
1) Better documentation of each transform, based on a given template (higher priority task)
2) An example notebook for every transform. The notebooks should be simple to use by taking in the user data, calling the API, and showing the output result. All other code (extra imports, parameter settings) should be hidden away. This will be done based on a notebook template. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

touma-I:
@shahrokhDaijavad can you please follow-up with the individual transform owners.

Bytes-Explorer:
@shahrokhDaijavad @agoyal26 Here is my suggestion on first batch of transforms to be updated. What do you guys think?

- [ ] Exact Dedup
- [ ] Fuzzy Dedup
- [ ] PDF2Parquet
- [ ] HTML2Parquet
- [ ] Doc_chunk
- [ ] Doc_quality
- [ ] text_encoder
- [ ] doc_id
- [ ] HAP

shahrokhDaijavad:
This is a good first batch, @Bytes-Explorer. The owners of these transforms are: Boris, Michele, Sung, Tsuzuku-san, and Yang Zhao. We can go ahead and assign them, but we need to discuss what (and a timeline) to expect if they are doing other projects at the moment.

Bytes-Explorer:
I believe you already have the template for documentation which should help all the owners fill in. Btw, Constantin has taken over all the work from Boris.

shahrokhDaijavad:
@cmadam , @dolfim-ibm , @sungeunan-ibm , @dtsuzuku-ibm and @ian-cho We have started a significant effort to simplify the use of DPK for the first-time users. 
A high-priority item is to have a better and more unified documentation (README files) for each transform. Beyond that, we want an example Jupyter notebook for each transform. For the second step, we will develop a template notebook and share it with you later. But the first step is first! 

For the first step, we have a template for what we think should be in the README for each transform (attached).
[DPK_Documentation_template.docx](https://github.com/user-attachments/files/17653334/DPK_Documentation_template.docx)

[...]

PR #800, opened 2024-11-13
Update README docs for language transforms

dolfim-ibm:

Updates for the `pdf2parquer`, `doc_chunk` and `text_encoder` transforms.

https://github.com/IBM/data-prep-kit/issues/753


transforms/language/doc_chunk/python/README.md
touma-I, line 1: @dolfim-ibm @shahrokhDaijavad What do you guys think of adding a section like this one below to show how a user can invoke the transform once they have done a pip install (alternative to cloning the repo)::

import ast
import sys

from data_processing.runtime.pure_python import PythonTransformLauncher
from data_processing.utils import ParamsUtils
from pdf2parquet_transform_python import Pdf2ParquetPythonTransformConfiguration

local_conf = {
    "input_folder": “input”,
    "output_folder": “output”,
}
params = {
    "data_local_config": ParamsUtils.convert_to_ast(local_conf),
    "data_files_to_use": ast.literal_eval("['.pdf','.docx','.pptx','.zip']"),
}
sys.argv = ParamsUtils.dict_to_req(d=params)
launcher = PythonTransformLauncher(runtime_config=Pdf2ParquetPythonTransformConfiguration())
launcher.launch()

shahrokhDaijavad: Nice job, @dolfim-ibm! Great job with the README files for all three transforms. They follow the template. 

What @touma-I is suggesting would be to add these lines of code in the section that says "Code example" and has the link to the upcoming Notebook example. These lines, together with the pip install, will be used in the Notebook, but they could also be used in a Python example that is not Notebook. I am ok either way: 1) Wait for the Notebook or 2) Add the lines now. 

@dolfim-ibm Please don't pick option 1 because it will make it easier on you! Maroun's question is how useful it is to have these lines.

dolfim-ibm: @touma-I @shahrokhDaijavad  I was actually adding the code block already, but then I realized it was 1-to-1 exactly the content of the example script. Instead of having to maintain multiple versions of it (with the high-risk) of being outdated, I think that linking to the example is still ok.

Honestly, I think the best is to plan in terms of a documentation engine which can embed working code examples, and to ensure in CI that those example codes are being executed.

shahrokhDaijavad (approved):
LGTM.

touma-I (approved):

PR #806, opened 2024-11-17
Fix set_s3_env_vars_to_component in kfp v2.

revit13:

`set_s3_env_vars_to_component` in kfp v2 accepts `env2key` dict  that is shared between the function invocation. This implementation can lead to errors, especially when the function is called multiple times, sometimes with a prefix. This PR fixes that.


roytman (approved):
/LGTM

Issue #807, opened 2024-11-17
[Bug] Ray clusters are created in CI/CD with `imagePullPolicy: Always`


revit13:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP workflows

### What happened + What you expected to happen

**Expected Behavior:** 
When creating the Ray cluster, the default ImagePullPolicy should be IfNotPresent. In CI/CD, the transform image is built and loaded into the kind cluster before the test run, ensuring that the local image code is tested.
**What happen:**
The Ray cluster was created with ImagePullPolicy set to Always, causing it to always pull the transform image from the remote repository. This resulted in an image pull error when the image was not available in the remote repository, and preventing the local version from being tested.

### Reproduction script

run `make workflow-test` in `transforms/universal/noop`.

### OS

Ubuntu

### Python

3.10.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

revit13:
closed with https://github.com/IBM/data-prep-kit/pull/808

PR #808, opened 2024-11-17
Update Kuberay api server version in requirements.env.

revit13:
The image tag for the KubeRay API server set in [ray_api_server_values.yaml](https://github.com/IBM/data-prep-kit/blob/dev/scripts/k8s-setup/ray_api_server_values.yaml) was not used because it was overridden in [install_kuberay.sh ](https://github.com/IBM/data-prep-kit/blob/dev/scripts/k8s-setup/tools/install_kuberay.sh#L12) script with the version specified in requirements.env. This PR fixes that.

#807


roytman (approved):
/LGTM

Issue #774, opened 2024-11-05
[Feature] Restructure transforms as their own modules


touma-I:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

Restructer the repo toreduce burden/constraints on contributors:

1. Each transform will have its own module name dpk_<transform> (ie. dpk_pdf2Parquet, dpk_doc_quality, etc) . Since all of our contributors are currently Python code, the module name is also the folder name where the contributors add their code.
2. The Python Folders, the Ray folder and the Spark folder will no longer be needed. We will be giving the contributor freedom on deciding whether ray and spark are submodules (i.e. dpk_<transform>/ray) or if they want to deliver it as a separate module (dpk_<transform>_ray)
3. The Src subfolder is no longer required. This is an area that is also left to the contributor as we each have our own style
4. Pyproject.toml will no longer be required. Developers can opt to have one using setuptools, poetry, hatch, etc but is no longer a requirement. We still need a requirements.txt file that lists all the dependencies for the code that is being contributed
5. Reduce (if not eliminate) the number of Makefiles for each transform. Right now we are down from 3 to 1. In anycase, contributor will not be required to maintain the Makefile and we simplified it quite to reduce complexity.  (Contributor can still create their own Makefile and use make within their folder anyway they want to automate simle tasks)
6. Both Dockerfiles for ray and Python will be removed. We will use a standard docker template for those transforms that need to be integrated with the KFP. I also see this area evolving over time as more and more of our users are running their pipelines and transforms inside of self contained docker container (even without KFP). So stay tuned for some evolution in this area but for all intended purposes, contributors of the transforms will not have to deal with Dockerfiles unless they want to for their own needs.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

touma-I:
For a short time, we will need to support both layout (old and new). During that time, the CI/CD will get a bit more complicated but expect we will do a pass at cleaning things up once all the transforms have been moved to the new layout. Also, the packaging will get a bit trickier if we decide to rollout new transform as their own modules while the other transforms are still using a flat structure.

shahrokhDaijavad:
When PR #1023 is merged, we will close this.

PR #809, opened 2024-11-17
Restructure Html2Parquet with its own dpk_ namespace

touma-I:

This is a first of a series of restructuring changes that are done to have each transform built as its own module (e.g. dpk_html2parquet) with a ray submodule (dpk_html2parquet.ray ).
- Removed python and ray folders and keep Dockerfile.python and Dockerfile.ray
- remove pyproject.toml
- move python code under dpk_html2parquet
- move ray code under dpk_html2parquet/ray
- change import statement to include module name
- replace recursive Makefile and use targets from .make.cicd.targets
- adapt kfp_ray/Makefile and other make target


.make.defaults
```diff
@@ -222,14 +222,15 @@ __check_defined = \
...
 	fi
-	$(DOCKER) build -t $(DOCKER_LOCAL_IMAGE) $(DOCKER_BUILD_EXTRA_ARGS)	\
+	$(DOCKER) build -f $(DOCKER_FILE) -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) $(DOCKER_BUILD_EXTRA_ARGS)	\
 		--platform $(DOCKER_PLATFORM)  			\
 		--build-arg EXTRA_INDEX_URL=$(EXTRA_INDEX_URL)	\
 		--build-arg BASE_IMAGE=$(BASE_IMAGE)  		\
 		--build-arg DPK_WHEEL_FILE_NAME=$(DPK_WHEEL_FILE_NAME) \
 		--build-arg BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')  	\
 		--build-arg GIT_COMMIT=$(shell git log -1 --format=%h) . 
-	$(DOCKER) tag $(DOCKER_LOCAL_IMAGE)  $(DOCKER_REMOTE_IMAGE)
```
daw3rd: I would keep DOCKER_LOCAL/REMOTE_IMAGE.  In particular, the REMOTE one is way to complicated/delicate.  In addition, these are both being. used elsewhere.  I think you need to revert to these for consistency/correctness.

.make.defaults
```diff
@@ -244,24 +245,25 @@ __check_defined = \
...
 	cp $(REPOROOT)/data-processing-lib/dist/*.whl data-processing-dist
 
+
+# Build and image using the local Dockerfile 
+# Assumes wheel has already been created
+.PHONY: .defaults.lib-whl-image 
+.defaults.lib-whl-image::
+	# Must be called with a DOCKER_LOCAL_IMAGE= settings.
+	@# Help: Build the Python $(DOCKER_LOCAL_IMAGE) using the the wheel file for the library
+	@$(eval LIB_WHEEL_FILE := $(shell find data-processing-dist/*.whl))
```
daw3rd: SInce DOCKER_LOCAL_IMAGE is required, should add

`$(call check_defined, DOCKER_LOCAL_IMAGE)`

.make.defaults
```diff
@@ -270,28 +272,9 @@ __check_defined = \
 	@# Help: Build the Python $(DOCKER_LOCAL_IMAGE) using the the wheel file for the library
 	@$(eval LIB_WHEEL_FILE := $(shell find data-processing-dist/*.whl))
```
daw3rd: Again,
`
$(call check_defined, DOCKER_LOCAL_IMAGE)`

.make.defaults
```diff
@@ -306,7 +289,7 @@ __check_defined = \
 	if [ -e ../python ]; then								\
```
daw3rd: And again,

`$(call check_defined, DOCKER_LOCAL_IMAGE)`

transforms/.make.cicd.targets
```diff
@@ -7,10 +7,15 @@
...
-## Default setting for TRANSFORM_RUNTIME uses folder name-- Old layout
-TRANSFORM_PYTHON_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).transform
-TRANSFORM_RAY_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).ray.transform
-TRANSFORM_PYTHON_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).spark.transform
+## Default setting for TRANSFORM_RUNTIME entry point: 
+# python -m dpk_html2parquet.ray.transform --help
+# or
+# python -m dpk_html2parquet.transform_python --help
+#
+TRANSFORM_PYTHON_SRC?="-m dpk_$(TRANSFORM_NAME).transform_python"
```
daw3rd: I guess there's a reason for not inclucing .py in the value?  SRC in the var name sort of suggests (to me) that it is the name of a source file.  Actually, this looks like a package name.  Does changing _SRC to _PKG make sense for these?

transforms/language/html2parquet/dpk_html2parquet/transform.py
daw3rd, line 1: For consistency in naming, this files should be named `transform_ray.py`.  That said, maybe a change in convention is warranted?  Each of the python, ray, spark main()-containing files could be `transform_main.py`, distinguished by their different package locations.  But, if not that, then `transform_ray.py`

transforms/.make.cicd.targets
```diff
@@ -28,62 +32,113 @@ clean:: .transforms.clean
...
-image::
-	@if [ -e Dockerfile ]; then \
-		$(MAKE) image-default ; \
-	else \
-		echo "Skipping image for $(shell pwd) since no Dockerfile is present"; \
+build:: image
+
+publish:
+	@if [ -e Dockerfile.python ]; then \
+		$(MAKE) DOCKER_REMOTE_IMAGE=$(DOCKER_REGISTRY_ENDPOINT)/$(TRANSFORM_NAME)-python:$(DOCKER_IMAGE_VERSION) \
```
daw3rd: Instead of setting DOCKER_REMOTE_IMAGE, which is delicate, I would use DOCKER_IMAGE_NAME and let DOCKER_REMOTE_IMAGE be set from that in .make.defaults.  This way it seems less delicate to me.

daw3rd: In fact, you've done this below for `test-image` and others.

transforms/.make.cicd.targets
```diff
@@ -28,62 +32,113 @@ clean:: .transforms.clean
...
+				.defaults.publish-image ; \
+	fi
+	@if [ -e Dockerfile.spark ]; then \
+		$(MAKE) DOCKER_REMOTE_IMAGE=$(DOCKER_REGISTRY_ENDPOINT)/$(TRANSFORM_NAME)-spark:$(DOCKER_IMAGE_VERSION) \
+				.defaults.publish-image ; \
+	fi
+
+test-image-sequence:: .defaults.lib-whl-image .transforms.test-image-help .transforms.clean
+
+test-image:: .default.build-lib-wheel
```
daw3rd, line 63: Can we split this into make test-image-python/ray/spark so I can only call and build the image I want to test.

transforms/.make.cicd.targets
```diff
@@ -28,62 +32,113 @@ clean:: .transforms.clean
...
 
-image::
-	@if [ -e Dockerfile ]; then \
-		$(MAKE) image-default ; \
-	else \
-		echo "Skipping image for $(shell pwd) since no Dockerfile is present"; \
+build:: image
+
+publish:
+	@if [ -e Dockerfile.python ]; then \
```
daw3rd: maybe split this into publish-python/ray/spark.

daw3rd (changes requested):

.make.defaults
```diff
@@ -244,24 +245,25 @@ __check_defined = \
...
-#	@# Help: Build the Python $(DOCKER_LOCAL_IMAGE) using the $(DOCKER_FILE), requirements.txt and data-processing-lib/python source
-#ifeq ($(USE_REPO_LIB_SRC), 1)
-#	$(MAKE) LIB_PATH=$(DPK_PYTHON_LIB_DIR) LIB_NAME=data-processing-lib-python .defaults.copy-lib
-#endif
-#	$(MAKE) DOCKER_IMAGE=$(DOCKER_LOCAL_IMAGE) .defaults.image
-#	-rm -rf data-processing-lib-python
 	
 .PHONY: .default.build-lib-wheel
 .default.build-lib-wheel:
 	make -C $(REPOROOT)/data-processing-lib build-pkg-dist
```
daw3rd: Can  you change `make` to `$(MAKE)`

transforms/language/html2parquet/kfp_ray/Makefile
```diff
@@ -2,10 +2,20 @@ REPOROOT=${CURDIR}/../../../../
...
 # Include the common configuration for this transform
-include ../transform.config
+#include ../transform.config
+
+SRC_DIR=${CURDIR}/../
+# Use the docker image that is built for ray runtime
+TRANSFORM_RUNTIME=ray
+## override settings in .make.default as they assume old structure with ray being the current folder
+DOCKER_IMAGE_NAME=$(TRANSFORM_NAME)-$(TRANSFORM_RUNTIME)
+DOCKER_LOCAL_IMAGE=$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION)
```
daw3rd, line 14: DOCKER_LOCAL_IMAGE is defined the same in .make.defaults, so I think you can remove it here.

transforms/language/html2parquet/kfp_ray/Makefile
```diff
@@ -2,10 +2,20 @@ REPOROOT=${CURDIR}/../../../../
...
+
 # Include the common configuration for this transform
-include ../transform.config
+#include ../transform.config
+
+SRC_DIR=${CURDIR}/../
+# Use the docker image that is built for ray runtime
+TRANSFORM_RUNTIME=ray
+## override settings in .make.default as they assume old structure with ray being the current folder
+DOCKER_IMAGE_NAME=$(TRANSFORM_NAME)-$(TRANSFORM_RUNTIME)
```
daw3rd, line 13: TRANSFORM_NAME does not seem to be set.  It was originally in ../transform.config.

transforms/language/html2parquet/kfp_ray/html2parquet_wf.py
```diff
@@ -20,7 +20,7 @@
 task_image = "quay.io/dataprep1/data-prep-kit/html2parquet-ray:latest"
 
 # the name of the job script
-EXEC_SCRIPT_NAME: str = "html2parquet_transform_ray.py"
+EXEC_SCRIPT_NAME: str = "-m dpk_html2parquet.ray.transform"
```
daw3rd, line 23: SCRIPT should be changed to MODULE

transforms/language/html2parquet/kfp_ray/pipeline_definitions.yaml
```diff
@@ -1,7 +1,7 @@
 pipeline_parameters:
     name: "html2parquet"
     description: "Pipeline for html2parquet task"
-    script_name: "html2parquet_transform_ray.py"
+    script_name: "-m dpk_html2parquet.ray.transform"
```
daw3rd, line 4: `script` to `module`

transforms/language/html2parquet/dpk_html2parquet/transform.py
```diff
@@ -51,7 +50,7 @@ def __init__(self, config: dict[str, Any]):
         if not isinstance(self.favor_recall, html2parquet_favor_recall):
             self.favor_recall = html2parquet_favor_recall[self.favor_recall]
 
-    def _convert_html2parquet(self, member_filename: str, file_name: str, content_bytes: bytes) -> dict:
+    def convert_html2parquet(self, member_filename: str, file_name: str, content_bytes: bytes) -> dict:
```
daw3rd, line 53: why the name change?  It is not used outside this file and so can be considered at least protected, if not private.  In general, methods should not be exposed if not necessary.  I know we're not consistent on this, but better to catch it when we can

.make.defaults
touma-I: Hi @daw3rd. My plan is to get rid of as may globally defined variables as I can. DOCKER_LOCAL and DOCKER_RMOTE in particular.  We have too many global variables that reference each other. Agree things are delicate but they don't have to be and I am using this opportunity to do gradual cleanup. This change was not done by accident.

.make.defaults
touma-I: I will be removing DOCKER_LOCAL_IMAGE gradually over time as part of the cleanup process.

transforms/language/html2parquet/dpk_html2parquet/transform.py
```diff
@@ -236,3 +236,35 @@ def apply_input_params(self, args: Namespace) -> bool:
         self.params = self.params | captured
         logger.info(f"html2parquet parameters are : {self.params}")
         return True
+
+
+class Html2Parquet(Html2ParquetTransform):
```
daw3rd, line 241: This should also inherit from AbstractTableTransform to indicate that it implements the transform() method below.

transforms/language/html2parquet/dpk_html2parquet/transform.py
daw3rd: And sub-classes generally are considered to have access to their parent _ methods.

transforms/.make.cicd.targets
touma-I: Actually, having it defined in the .make.defaults is not good practice. Over time, DOCKER_REMOTE will be removed from the globally defined variables.

transforms/.make.cicd.targets
touma-I: @daw3rd this was initially meant for CI/CD but I supposed the developers may be interested in using it to run targeted testing. Agree, will make the change. Thanks.

transforms/language/html2parquet/dpk_html2parquet/transform.py
daw3rd, line 241: This class needs some documentation here and in a README.

transforms/language/html2parquet/dpk_html2parquet/transform.py
touma-I: @daw3rd it is already the the ray scope (dpk_html2parquet.ray.transform() why do you want to add ray to the end as well ?

daw3rd (changes requested):
This transform needs a README.md

transforms/language/html2parquet/Makefile
```diff
@@ -1,78 +1,24 @@
...
-	@# Help: Recursively make $@ all subdirs 
-	$(MAKE) RULE=$@ .recurse
+################################################################################
+# This defines the transforms' version number as would be used
+# when publishing the wheel.  In general, only the micro version
+# number should be advanced relative to the DPK_VERSION. 
+#
+# If you change the versions numbers, be sure to run "make set-versions" to 
+# update version numbers across the transform (e.g., pyproject.toml).
+#TRANSFORM_VERSION=$(DPK_VERSION)
```
daw3rd: Maybe delete this?

transforms/language/html2parquet/dpk_html2parquet/ray/transform.py
```diff
@@ -10,24 +10,14 @@
...
+from data_processing.utils import get_logger
 from data_processing_ray.runtime.ray import RayTransformLauncher
 from data_processing_ray.runtime.ray.runtime_configuration import (
     RayTransformRuntimeConfiguration,
 )
-from html2parquet_transform import (
-    Html2ParquetTransform,
-    Html2ParquetTransformConfiguration,
-)
+from dpk_html2parquet.transform import Html2ParquetTransform, Html2ParquetTransformConfiguration
```
daw3rd, line 19: You need to run pre-commit.

transforms/language/html2parquet/python/README.md
daw3rd, line 1: This file can form the basis of a new readme in ../

daw3rd (changes requested):

transforms/language/html2parquet/dpk_html2parquet/transform.py
touma-I: I don't see the reason why this particular method was done as a private method in the first place. Once you create a valid object, it might be helpful to have access to this method.

transforms/language/html2parquet/dpk_html2parquet/transform.py
touma-I: @daw3rd I admit, This is one that I struggled a lot with. The conclusion I came to: Only methods that are invoked from the runtime need to implement the abstraction.  For now, this method is not intended to be called from the runtime so I don't see the benefit making derived from the AbstractTableTransform.

touma-I: BIG TIME :-). working on it. Thanks

.make.defaults
touma-I: Done! Thanks

transforms/.make.cicd.targets
touma-I: @daw3rd When is publish being used. Do you know ?

transforms/language/html2parquet/kfp_ray/pipeline_definitions.yaml
touma-I: @daw3rd this will unecessary break the downstream code. This is still a python script that is invoked differently. Anyway, I will leave it to @eres / @roytman  to decide.

roytman: Yes, it is a Python file (script) that is executed as a Ray job

roytman (commented):
If we are doing so massive refactoring, should we combine `test` and `test-data` into one dir, e.g. 
- test
   - data
        - input
        - expected

roytman:
if each transformer builds its own module, should we add __init.py__ files and create a unified namespace?
For example: instead 
`from dpk_html2parquet.transform import Html2ParquetTransformConfiguration` 
use 
`from dpk_html2parquet import Html2ParquetTransformConfiguration`

the same for the ray runtime.

transforms/language/html2parquet/kfp_ray/pipeline_definitions.yaml
touma-I: In this case, let's keep it as script_name:. The fewer changes we make for this first iteration, the fast we will be able to move, test and iterate. :-)

touma-I:
> If we are doing so massive refactoring, should we combine `test` and `test-data` into one dir, e.g.
> 
> * test
>   
>   * data
>     
>     * input
>     * expected

@roytman Why not leave it to the transform owner developer to decide if they want to nest the test-data under test. All we care about that we have a test folder for running the pytest. no ? where the developer puts their data is up to them. no ?

transforms/language/html2parquet/dpk_html2parquet/transform.py
daw3rd: Always ere on the side of private/protected until someone asks for it.  Otherwise, you unnecessarily create additional support and backwards compatibility issues.  It unnecessarily ties our hands going forward.

touma-I:
> if each transformer builds its own module, should we add **init.py** files and create a unified namespace? For example: instead `from dpk_html2parquet.transform import Html2ParquetTransformConfiguration` use `from dpk_html2parquet import Html2ParquetTransformConfiguration`
> 
> the same for the ray runtime.

How did I miss that? Done. Thanks @roytman

transforms/language/html2parquet/dpk_html2parquet/ray/transform.py
touma-I: @daw3rd can you elaborate ? what is pre-commit ?

transforms/language/html2parquet/Makefile
touma-I: Done. Thanks

transforms/.make.cicd.targets
touma-I: Done

transforms/.make.cicd.targets
touma-I: I don't think we need to do this. _SRC is fine. again, I will be taking another pass at all the Makefile once we get all the current work in place.

.make.defaults
touma-I: I need to keep changes to the Makefile to a minimum as most of it will change in a second iteration once I get a good handle on it. If this is not causing a bug right now or breaking the code, I won't be changing it in this iteration.

.make.defaults
touma-I: Not in this iteration. Most of this will be changing in a second iteration. For now, If it is not breaking the code, let's skip it. Thanks

transforms/language/html2parquet/kfp_ray/Makefile
touma-I: @daw3rd this is not received as an argument for the make -C:

$(MAKE) -C kfp_ray TRANSFORM_NAME=$(TRANSFORM_NAME) workflow-test;

transforms/language/html2parquet/kfp_ray/Makefile
touma-I: no. this is a bit more complicated than this. Let's keep it for now. I will do a second pass on the makefile and trim some of the fat...

transforms/language/html2parquet/kfp_ray/html2parquet_wf.py
touma-I: no. let's keep it as script for now.

revit13 (approved):

transforms/.make.cicd.targets
```diff
@@ -7,10 +7,15 @@
 include $(REPOROOT)/transforms/.make.transforms
 
 ######################################################################
-## Default setting for TRANSFORM_RUNTIME uses folder name-- Old layout
-TRANSFORM_PYTHON_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).transform
-TRANSFORM_RAY_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).ray.transform
-TRANSFORM_PYTHON_RUNTIME_SRC_FILE=-m dpk_$(TRANSFORM_NAME).spark.transform
+## Default setting for TRANSFORM_RUNTIME entry point: 
+# python -m dpk_html2parquet.ray.transform --help
```
roytman: do we really need the `dpk` prefix for each transformer? If you want to prevent possible name conflicts, what about ` dpk` namespace? e.g. `dpk.html2parquet.ray.transform`

touma-I: @roytman : @dolfim-ibm made a passionate argument for this. I am good either way. @dolfim-ibm @daw3rd Can yo guys weigh in ? thanks

dolfim-ibm: my argument was about having something, i.e. either prefix or namespace. I think it is very important to have a library which doesn't provide lots of global names which might overlap with others.
it also brings a bit of "branding" and makes it clearer to users that all these transforms come from the same place, i.e. data-prep-kit.

roytman: so why not `dpk.html2parquet.ray.transform` ?

roytman: It's just a suggestion; I'm OK with any decision.

transforms/language/html2parquet/notebooks/html2parquet.ipynb
```diff
@@ -0,0 +1,184 @@
...
+      "11:38:33 INFO - Completed 1 files (100.0%) in 0.003 min\n",
+      "11:38:33 INFO - Done processing 1 files, waiting for flush() completion.\n",
+      "11:38:33 INFO - done flushing in 0.0 sec\n",
+      "11:38:33 INFO - Completed execution in 0.003 min, execution result 0\n"
+     ]
+    }
+   ],
+   "source": [
+    "from dpk_html2parquet.transform_python import Html2ParquetRuntime\n",
+    "x=Html2ParquetRuntime(input_folder= \"input\", \n",
```
sungeunan-ibm: "input/ai-alliance-index.html"

transforms/language/html2parquet/notebooks/html2parquet.ipynb
```diff
@@ -0,0 +1,184 @@
...
+    "               output_folder= \"output\", \n",
```
sungeunan-ibm: Do we need an output folder for this to work?

transforms/language/html2parquet/notebooks/html2parquet.ipynb
```diff
@@ -0,0 +1,184 @@
...
+    "               data_files_to_use=['.html']).transform()"
```
sungeunan-ibm: I know input file is just one html file but ['.zip', '.html'] in case user wants to test html zip in the notebook.

sungeunan-ibm: or add a comment something like: '.zip' in case your input file is zip of html files.

sungeunan-ibm (approved):
I've reviewed all the changes, and they all make sense to me. I left some comments. Thank you for the significant effort you put into restructuring the code! @touma-I

roytman (approved):

shahrokhDaijavad:
@touma-I I don't see a README file for this one!

Issue #752, opened 2024-10-29
Update README file by adding links to the RAG and fine-tuning examples. 


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

Root REDAME page should be updated for the user to know where to go for RAG and fine-tuning examples. They are currently in the examples folder and may be hidden. 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

touma-I:
@shahrokhDaijavad to own the actual deliverable on this one and submit a PR .

PR #810, opened 2024-11-18
A few changes in the root README

shahrokhDaijavad:
In addition to addressing #752, three other changes were made:
1) Adding Web2Parquet to the table, after the merging of #797 
2) Updating the Citations, because the order of authors in the DPK paper has been changed 
3) HTML2parquet has had a KFP implementation for a while, but it was never check-marked in the table.


transforms/universal/web2parquet/README.md
```diff
@@ -49,4 +49,4 @@ Web2Parquet(urls= ['https://thealliance.ai/'],
                     depth=2,
```
daw3rd, line 57: Something looks wrong here.  Maybe replace  `= ` with `(`

daw3rd (dismissed):

touma-I (commented):
Thanks shahorkh. Looks good.

touma-I (approved):

PR #813, opened 2024-11-19
Restructure Pdf2Parquet with its own dpk_ namespace

touma-I:

This is a first of a series of restructuring changes that are done to have each transform built as its own module (e.g. dpk_pdf2parquet) with a ray submodule (dpk_pdf2parquet.ray ).

Removed python and ray folders and keep Dockerfile.python and Dockerfile.ray
move python code under dpk_pdf2parquet
move ray code under dpk_pdf2parquet/ray


transforms/language/pdf2parquet/dpk_pdf2parquet/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .transform import *
+from .local_python import *
+from .transform_python import *
+from .local import *
```
dolfim-ibm: I think here we should import only from the transform. the local and local_python shouldn't export any symbol needed, or?

transforms/language/pdf2parquet/dpk_pdf2parquet/ray/s3_ray.py
```diff
@@ -16,7 +16,7 @@
 
 from data_processing.utils import ParamsUtils
 from data_processing_ray.runtime.ray import RayTransformLauncher
```
dolfim-ibm: `data_processing_ray` is available only if dpk is installed with the `[ray]` extra, right?

we could wrap these imports in `try: ... except ImportError: raise ImportError("Please install data_prep_kit[ray]")` or similar.

transforms/language/pdf2parquet/dpk_pdf2parquet/__init__.py
touma-I: Done! Please resolve conversation and approve. Thanks

transforms/language/pdf2parquet/dpk_pdf2parquet/ray/s3_ray.py
touma-I: Done! Please resolve conversation and approve. Thanks

dolfim-ibm (approved):

roytman (approved):
@touma-I thank you for your work.

shahrokhDaijavad:
@touma-I Please check what I have done before I make changes to the others. You seem to have only the run-cli-sample as a target for `make`, so I got rid of everything else that had to do with `make `other targets. 
Do we need to provide an example of how to run the ray code?

Issue #803, opened 2024-11-16
[Bug] Cannot run KFP pipeline for fuzzy dedup with more than 100 actors


cmadam:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP workflows, Library/core, Transforms/universal/fdedup

### What happened + What you expected to happen

I am not able to run a KFP pipeline for fuzzy dedup that creates 102 actors with the following configuration:
```
{
  "cpu": 8,
  "image": "us.icr.io/cil15-shared-registry/preprocessing-pipelines/fuzzydedup-ray:0.2.8-v5",
  "imagePullPolicy": "Always",
  "image_pull_secret": "****-***-***-**",
  "max_replicas": 35,
  "memory": 60,
  "min_replicas": 35,
  "replicas": 35
}
``` 
The error I am getting while trying to run the pipeline is:
```
(orchestrate pid=382, ip=172.17.29.195) 11:01:01 INFO - Cluster resources: {'cpus': 280, 'gpus': 0, 'memory': 2180.0, 'object_store': 652.3036788804457}
(orchestrate pid=382, ip=172.17.29.195) 11:01:01 INFO - Number of workers - 102 with {'num_cpus': 2.3333333333333335, 'max_restarts': -1} each
(orchestrate pid=382, ip=172.17.29.195) Traceback (most recent call last):
(orchestrate pid=382, ip=172.17.29.195)   File "/home/ray/anaconda3/lib/python3.10/site-packages/data_processing_ray/runtime/ray/transform_orchestrator.py", line 96, in orchestrate
(orchestrate pid=382, ip=172.17.29.195)     processors = RayUtils.create_actors(
(orchestrate pid=382, ip=172.17.29.195)   File "/home/ray/anaconda3/lib/python3.10/site-packages/data_processing_ray/runtime/ray/ray_utils.py", line 121, in create_actors
(orchestrate pid=382, ip=172.17.29.195)     raise UnrecoverableException(f"out of {len(actors)} created actors only {len(alive)} alive")
(orchestrate pid=382, ip=172.17.29.195) data_processing.utils.unrecoverable.UnrecoverableException: out of 102 created actors only 100 alive
(orchestrate pid=382, ip=172.17.29.195) 11:03:06 ERROR - Exception during execution out of 102 created actors only 100 alive: None
```

### Reproduction script

[...]

PR #814, opened 2024-11-19
update to actor list limit to fix issue 803

blublinsky:
Default limit in the actor list API is 100, so if the number of actors is more than that, then the code was thinking that not all actors were created

https://github.com/IBM/data-prep-kit/issues/803


cmadam:
Can you please increase the limit?  Make it 100000 instead of 1000.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -113,7 +113,7 @@ def operator() -> ActorHandle:
         actors = [operator() for _ in range(n_actors)]
         for i in range(120):
             time.sleep(1)
-            alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")])
+            alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=1000)
```
daw3rd: list_actors() is now in a different package.  See https://docs.ray.io/en/latest/ray-observability/reference/doc/ray.util.state.list_actors.html.  Should we be using that one instead of the `experimental` package?

daw3rd: Why is there any limit here?  Should this be a configuration parameter?

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -113,7 +113,7 @@ def operator() -> ActorHandle:
         actors = [operator() for _ in range(n_actors)]
         for i in range(120):
             time.sleep(1)
-            alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")])
+            alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=n_actors + 10)
```
daw3rd: Why is 10 the right number. what if a transform's ray runtime starts its own actors?  Maybe this should be n_actors*2 or the above with the ability to override with a config value (env var?)

cmadam: I agree.  The idea to use the number of actors to compute the limit is great, but l think it would be better to use 
```
num_actors * 10
```
instead of 
```
num_actors + 10
```

daw3rd: what is the downside of setting the limit to 1E6

touma-I (changes requested):
Looks like the code is still changing. Please move the PR to draft mode and then promote to ready for review once it is ready. Thanks

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
blublinsky: Its version dependent
We have to move it once we move to the next version

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
blublinsky: Gents, If we create 5 actors whats the reason asking for 100000? We are asking only for what we need and padding it with 10

daw3rd: @blublinsky 1) what is the downside of using 1E6 and 2) what if the transform creates its own actors?  Please advise.

blublinsky: take a closer look at this line of code.
We are listing only actors, based on filters - class name and running. I do not see transform creating actors using the same class. I do not want to put a number of 1e6, which is an absolutely arbitrary number. This said, I assume that if an transform creates his own actors, it will use the same API, but listing in this case will use a different filter.

cmadam: @blublinsky : I have two questions about the code we are discussing here:
https://github.com/IBM/data-prep-kit/blob/91f22eb7f321442e882899804b9cc043ff8eba54/data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py#L114-L121
1. It looks like the only way for this code to execute successfully is if the `list_actors()` method returns **exactly** `n_actors` not more and not less.  Then, I think the limit should equal `n_actors`
2. I might be missing something, but why is the condition on line 117 requiring that all the actors are up?  Wouldn't this work even if, say, 80% of the actors were up?  In other words, what the drawback on replacing the condition on line 117 with this one?
```
if len(alive) >= 0.8 * len(actors):
    return actors
# failed - ....
```

blublinsky: The documentation is quite clear:
```
def list_actors(
    address: Optional[str] = None,
    filters: Optional[List[Tuple[str, PredicateType, SupportedFilterType]]] = None,
    limit: int = DEFAULT_LIMIT,
    timeout: int = DEFAULT_RPC_TIMEOUT,
    detail: bool = False,
    raise_on_missing_output: bool = True,
    _explain: bool = False,
) -> List[ActorState]:
    """List actors in the cluster.

    Args:
        address: Ray bootstrap address, could be `auto`, `localhost:6379`.
            If None, it will be resolved automatically from an initialized ray.
        filters: List of tuples of filter key, predicate (=, or !=), and
            the filter value. E.g., `("id", "=", "abcd")`
            String filter values are case-insensitive.
        limit: Max number of entries returned by the state backend.
        timeout: Max timeout value for the state APIs requests made.
        detail: When True, more details info (specified in `ActorState`)
            will be queried and returned. See
            :class:`ActorState <ray.util.state.common.ActorState>`.
        raise_on_missing_output: When True, exceptions will be raised if
            there is missing data due to truncation/data source unavailable.
        _explain: Print the API information such as API latency or
            failed query information.
```
limit is Max number of entries returned by the state backend.
If the limit is 100 and you only have 5, you will still get 5

cmadam: I agree, but I don't understand why we need to add 10 to the number of actors in the limit. Why not just set the limit to be equal to the number of actors, since this is what is tested in line 117. 

For example, if there are 5 actors, why would we ever need a limit greater than 5? Can it happen that `list_actors()` will return 6 actors, if we only intended to create 5?  And, in that case, do we need the run to fail because there are more actors available than what we requested to create? 

And again, I am not clear why are we testing for exact equality in line 117.  Would we be ok to proceed even if only 80% of the actors are alive?  Or is there a reason why we need all the actors to be up when we start?  I am thinking of a scenario when we request 10 worker groups, but only 9 are allocated, and one is pending due to lack of resources.  Would it be ok to proceed with the execution in this case?

blublinsky: Just in case, I am  a bit concerned to use the same number

blublinsky: Another thing that I found today, max limit supported is 10000

blublinsky: @cmadam, @daw3rd any progress on this?

cmadam: If the number of created actors is greater than the limit, the function `create_actors()` should still return the actor handles

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -110,15 +112,21 @@ def operator() -> ActorHandle:
...
-        # failed - raise an exception
-        print(f"created {actors}, alive {alive}")
-        raise UnrecoverableException(f"out of {len(actors)} created actors only {len(alive)} alive")
+        overall = current + n_actors
+        if overall < MAX_LIST:
+            n_list = min(overall + 10, MAX_LIST)
+            for i in range(120):
+                time.sleep(1)
+                alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=n_list)
+                if len(actors) >= len(alive) - current:
```
cmadam: The inequality operator here should be either `<=` or `==`

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,30 @@ def operator() -> ActorHandle:
...
+        overall = c_len + n_actors
+        if overall < MAX_LIST:
+            n_list = min(overall + 10, MAX_LIST)
+            for i in range(120):
+                time.sleep(1)
+                alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=n_list)
+                if len(alive) >= n_actors + c_len:
+                    return actors
+            # failed
+            if len(alive) >= n_actors/2 + c_len:
```
cmadam: @blublinsky : the `alive` variable might not be initialized at this point.  Can you please initialize, after line 120, this variable?  Something like:
```
    alive = []
```

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,30 @@ def operator() -> ActorHandle:
...
+               # At least half of the actors were created
+                logger.info(f"created {n_actors}, alive {len(alive)} Running with less actors")
+                return alive - current
```
cmadam, line 131: This will not work, and should be replaced with this code:
```
                return [actor for actor in alive if actor not in current]
```

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
blublinsky: It can not. You can't get to line 128 without going through the loop, that sets it

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
cmadam: This line has 2 problems:
1. There is no such operator `-` between 2 python lists.
2. This function is supposed to return `list[ActorHandle]`, but `alive` and `current` are of type `list[ActorStatus]`

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -16,11 +16,13 @@
 
 import ray
 from ray.experimental.state.api import list_actors
-from data_processing.utils import GB, UnrecoverableException
+from data_processing.utils import GB, UnrecoverableException, get_logger
 from ray.actor import ActorHandle
 from ray.exceptions import RayError
 from ray.util.actor_pool import ActorPool
 
+MAX_LIST = 10000  # Max number of actors returned by list
```
cmadam: `MAX_LIST` is necessary, because there is an upper limit on the number of `ActorState`s returned by the `list_actors()` method:
https://github.com/ray-project/ray/blob/569f7df9067c5654fb57ba7bc4792b3ba5aaa846/python/ray/util/state/common.py#L50-L53

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,30 @@ def operator() -> ActorHandle:
             time.sleep(creation_delay)
             return clazz.options(**actor_options).remote(params)
 
+        logger = get_logger(__name__)
         cls_name = clazz.__class__.__name__.replace('ActorClass(', '').replace(')','')
+        current = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=MAX_LIST)
```
cmadam: The `current` variable is needed to handle the case when actors of a specific class already exist in the ray cluster, and the current `create_actors()` request is adding more actors of a specific kind.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,30 @@ def operator() -> ActorHandle:
...
-        for i in range(120):
-            time.sleep(1)
-            alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")])
-            if len(actors) == len(alive):
-                return actors
-        # failed - raise an exception
-        print(f"created {actors}, alive {alive}")
-        raise UnrecoverableException(f"out of {len(actors)} created actors only {len(alive)} alive")
+        overall = c_len + n_actors
+        if overall < MAX_LIST:
```
cmadam: If the `overall` number of actors is less than `MAX_LIST`, we can get their state from the ray server. Otherwise, we cannot determine their state, and the code execution becomes unpredictable. Therefore, we throw an `UnrecoverableException` if we have more than 10000 actors of the same type instantiated.  As a reference, the largest transforms we run now, have about 400 actors of a given type.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,30 @@ def operator() -> ActorHandle:
...
+            n_list = min(overall + 10, MAX_LIST)
+            for i in range(120):
+                time.sleep(1)
+                alive = list_actors(filters=[("class_name", "=", cls_name), ("state", "=", "ALIVE")], limit=n_list)
+                if len(alive) >= n_actors + c_len:
+                    return actors
+            # failed
```
cmadam: @blublinsky : I would suggest that we remove the code in lines 128-131, and we replace it with raising an `UnrecoverableException` for two reasons:
1. It is better to just let the user know if a ray cluster could not be allocated that matches the desired specifications, rather than run with less resources. 
2. I do not see a proper way to implement this solution.  The current code is broken, it will not run, because you cannot just subtract two lists using `-` and it is returning `list[ActorStatus]` instead of `list[ActorHandle]`.  It would be possible, maybe, to run something like this:
```
               alive_actor_ids = [a.actor_id for a in alive]
               current_actor_ids = [a.actor_id for a in current]
               return [
                   actor for actor in actors
                   if actor._actor_id in alive_actor_ids and actor._ray_actor_id not in current_actor_ids
               ]
```
But this is not a proper solution, as it is accessing a protected variable of the `ActorHandle` class.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
blublinsky: You are correct. rewrote this (and tested)

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -109,16 +111,35 @@ def operator() -> ActorHandle:
...
+                created_ids = [item.actor_id for item in alive if item not in current]
+                return [
+                    actor for actor in actors
+                    if (str(actor._ray_actor_id).
+                        replace('ActorID(', '').replace(')','') in created_ids)]
+            else:
+                # too few actors created
+                raise UnrecoverableException(f"Created {n_actors}, alive {len(alive)}. Too few actors were created")
+        else:
+            raise UnrecoverableException(f"Overall number of actors of class {clazz.__name__} is {overall}. "
```
cmadam: This line of code does not work. Please replace
```
raise UnrecoverableException(f"Overall number of actors of class {clazz.__name__} is {overall}. "
```
with
```
raise UnrecoverableException(f"Overall number of actors of class {cls_name} is {overall}. "
```

cmadam:
@touma-I - the code is ready for review.

Issue #816, opened 2024-11-20
[Bug] parquet files with columns containing large list of byte arrays can not be read by pyarrow.


daw3rd:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/core

### What happened + What you expected to happen

I have a parquet file that has a column containing a list of images, as byte arrays.  Under some circumstances, such files are not readable by pyarrow.

### Reproduction script

grab https://ibm.ent.box.com/file/1684883605503?s=9qcne0iubeji6t6a77gh2sxgi29k1tjp as test.parquet
```bash
cd transforms/universal/noop/python
mkdir input
cp .../test.parquet input
make venv
source venv/bin/activate
python src/noop_transform_python.py --data_local_config  "{'input_folder': 'input', 'output_folder':'output'}"
09:37:35 INFO - Launching noop transform
09:37:35 INFO - noop parameters are : {'sleep_sec': 1, 'pwd': 'nothing'}
09:37:35 INFO - pipeline id pipeline_id
09:37:35 INFO - code location None
09:37:35 INFO - data factory data_ is using local data access: input_folder - input output_folder - output
09:37:35 INFO - data factory data_ max_files -1, n_sample -1
09:37:35 INFO - data factory data_ Not using data sets, checkpointing False, max files -1, random samples -1, files to use ['.parquet'], files to checkpoint ['.parquet']
09:37:35 INFO - orchestrator noop started at 2024-11-20 09:37:35
09:37:35 INFO - Number of files is 1, source profile {'max_file_size': 341.400185585022, 'min_file_size': 341.400185585022, 'total_file_size': 341.400185585022}
09:37:39 ERROR - Failed to convert byte array to arrow table, exception Nested data conversions not implemented for chunked array outputs. Skipping it
09:37:39 WARNING - Transformation of file to table failed
09:37:39 INFO - Completed 1 files (100.0%) in 0.07 min
09:37:39 INFO - Done processing 1 files, waiting for flush() completion.
09:37:39 INFO - done flushing in 0.0 sec
09:37:39 INFO - Completed execution in 0.07 min, execution result 0
Segmentation fault: 11
```

### Anything else
```
(venv) dawood@MacBookPro:~/git/data-prep-kit/transforms/universal/noop/python$ parquet-tools inspect input/test.parquet 

############ file meta data ############
created_by: parquet-cpp-arrow version 16.1.0
[...]

PR #817, opened 2024-11-20
Fix for inability to read some parquet files (issue #816)

daw3rd:
To allow proper reading and processing of some parquet files that are otherwise unreadable.

#816


data-processing-lib/python/src/data_processing/utils/transform_utils.py
```diff
@@ -144,8 +145,18 @@ def convert_binary_to_arrow(data: bytes, schema: pa.schema = None) -> pa.Table:
             table = pq.read_table(reader, schema=schema)
             return table
         except Exception as e:
-            logger.error(f"Failed to convert byte array to arrow table, exception {e}. Skipping it")
-            return None
+            logger.warning(f"Could not convert bytes to pyarrow: {e}")
+
```
blublinsky, line 149: Can you, please put a comment here about why polars. Just copy the blur from the where you found this solution

blublinsky (approved):
Approving, but please, add the comment

data-processing-lib/python/src/data_processing/utils/transform_utils.py
daw3rd: Done.

transforms/universal/filter/python/src/filter_transform.py
```diff
@@ -67,6 +67,10 @@ def __init__(self, config: dict):
         self.logical_operator = config.get(filter_logical_operator_key, filter_logical_operator_default)
         self.columns_to_drop = config.get(filter_columns_to_drop_key, filter_columns_to_drop_default)
 
+        # Temporarily here to test if this can allow use to process files that are required to be read by polars for mm
```
touma-I: consider removing the 4 commented lines or restore file to its original commit.

data-processing-lib/python/requirements.txt
```diff
@@ -4,3 +4,4 @@
   argparse
   mmh3
   psutil
+  polars
```
touma-I: If this is the first module to introduce polars, consider adding a >= version constraints

touma-I (changes requested):
Let me know if the two comments can be addressed. Thanks

data-processing-lib/python/src/data_processing/utils/transform_utils.py
```diff
@@ -144,8 +145,21 @@ def convert_binary_to_arrow(data: bytes, schema: pa.schema = None) -> pa.Table:
...
+        # "Nested data conversions not implemented for chunked array outputs"
+        # See issue 816 https://github.com/IBM/data-prep-kit/issues/816.
+        logger.info(f"Attempting read of pyarrow Table using polars")
+        try:
+            import polars
+
+            df = polars.read_parquet(io.BytesIO(data))
+            table = df.to_arrow()
+        except Exception as e:
+            logger.warning(f"Could not convert bytes to pyarrow using polars: {e}. Skipping.")
```
touma-I: should this be logger.error ?

data-processing-lib/python/requirements.txt
daw3rd: set to >=1.16.0

data-processing-lib/python/src/data_processing/utils/transform_utils.py
daw3rd: Good catch.  Thanks.

transforms/universal/filter/python/src/filter_transform.py
daw3rd: done

touma-I (approved):
Thanks

PR #825, opened 2024-11-24
Restructure doc_chunk with its own dpk_ namespace

touma-I:

This is a first of a series of restructuring changes that are done to have each transform built as its own module (e.g. dpk_doc_chunk) with a ray submodule (dpk_doc_chunk.ray ).

move python code under dpk_doc_chunk
move ray code under dpk_doc_chunkt/ray


touma-I:
merged with alpha-1.0

PR #839, opened 2024-11-28
Simplified fix for issue 803

cmadam:
This provides a simplified, minimal fix for issue #803 

#803

### DIFF
--- a/data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
+++ b/data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
@@ -22,6 +22,9 @@ from ray.experimental.state.api import list_actors
 from ray.util.actor_pool import ActorPool
 
 
+# This value matches the constant `RAY_MAX_LIMIT_FROM_API_SERVER` defined in the ray source code here:
+# https://github.com/ray-project/ray/blob/569f7df9067c5654fb57ba7bc4792b3ba5aaa846/python/ray/util/state/common.py#L50-L53
+
 RAY_MAX_ACTOR_LIMIT = 10000
 
 
--- a/data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
+++ b/data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
@@ -15,11 +15,14 @@ import time
 from typing import Any
 
 import ray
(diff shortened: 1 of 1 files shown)


data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
```diff
@@ -15,13 +15,16 @@
...
 import ray
-from ray.experimental.state.api import list_actors
 from data_processing.utils import GB, UnrecoverableException
 from ray.actor import ActorHandle
 from ray.exceptions import RayError
+from ray.experimental.state.api import list_actors
 from ray.util.actor_pool import ActorPool
 
 
+RAY_MAX_ACTOR_LIMIT = 10000
```
touma-I, line 28: Please consider using environment variable as below. This keep the default at 100 (as it is now) but then allows you to set an environment variable to increate the number of actors if needed
```suggestion
RAY_MAX_ACTOR_LIMIT = int (os.environ.get("RAY_MAX_ACTOR_LIMIT", '100'))
```

touma-I (commented):
Thanks @cmadam This is exactly what we need so we can just address the issue we are facing. I encourage you to consider the comment to use an environment variable though. This way we can use a default and if we want to overwrite the default, we can do so without rebuilding the code. Let me know what you think.

data-processing-lib/ray/src/data_processing_ray/runtime/ray/ray_utils.py
cmadam: I do not see a benefit from using an additional environment variable. First, setting 100 as the default means issue #803 will occur by default, requiring users to read the documentation, locate the issue, and define a new environment variable to fix it. Second, Ray already uses an environment variable [to define the `RAY_MAX_LIMIT_FROM_API_SERVER` constant here](https://github.com/ray-project/ray/blob/569f7df9067c5654fb57ba7bc4792b3ba5aaa846/python/ray/util/state/common.py#L50-L53). Playing with this environment variable might allow us to change the max limit value, but I do not see what is the advantage of doing so.

cmadam: @touma-I : I can see that @dtsuzuku-ibm has opened an identical issue (#858) and a pull request (#859) with exactly the same fix as proposed here.

touma-I (approved):

PR #849, opened 2024-12-03
Enhance header cleanser module with multi-processing and timeout

takuyagt:

We observed scanning codes with scancode-toolkit was slow. Sometimes, the processing will get stuck. We should improve the throughput with the multi-processing, and avoid getting stuck with a timeout logic.


Param-S (approved):
lgtm

Issue #844, opened 2024-11-30
Add the first Google Colab Compatible Notebook as a template for all transforms 


shahrokhDaijavad:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

We will use the Google Colab compatible notebook by @Ryan-Gordon-314159 for pdf2parquet as a template for all other transforms to the examples folder. 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #851, opened 2024-12-03
Notebook template for transforms to run on Google Colab

Ryan-Gordon-314159:
This PR contains a notebook that can serve as a guide for porting transforms over to work on Google Colab

Issue #844

### DIFF
--- a/examples/notebooks/pdf2parquet_colab-version.ipynb
+++ b/examples/notebooks/pdf2parquet_colab-version.ipynb
@@ -1,0 +1,268 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "afd55886-5f5b-4794-838e-ef8179fb0394",
+   "metadata": {},
+   "source": [
+    "### Open this notebook in Google Colab"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9f3dbe08",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


shahrokhDaijavad (approved):
Thanks, @Ryan-Gordon-314159. As we discussed, the open colab icon will work only when this PR is merged. Other than that, I have tested this by manually running it on Google Colab.

examples/notebooks/pdf2parquet_colab-version.ipynb
touma-I, line 1: We need to change the notebook to tell a story that would be interesting to the end user. What do you think of this?
- Step 1: Enable Collab
- Step 2: Pip install
- Step 3: Identify interesting pdf files and use web crawler to fetch them. If possible, identify pdf contant that has OCR in it
- Step 4: Configure the transform to generate MD, run the transform
- Step 5: Show output from the transform, not only the new content but any other fields that the transfrom may have produced that may be of interest based on the content
- Step 6: Repeat step 4 and 5 above but with a different configuration of the tansform for example, generating Json instead of  MD or extracting OCR, etc.

touma-I (changes requested):
Please see comments. We need to make this notebook tell a story (not just copy from other notebooks)

examples/notebooks/pdf2parquet_colab-version.ipynb
shahrokhDaijavad: We want to consider getting rid of my_utils.py file, if possible.

touma-I: please do not download my_utils.py. I won't be able to merge this PR with this in it.

touma-I (changes requested):
We should not be download utils.py ?

PR #855, opened 2024-12-04
Update README.md for check-marking the table with Python and Spark versions of fdedup

shahrokhDaijavad:
We now have the Python and Spark versions of fdedup in the repo.


touma-I (changes requested):
Should we also update the pip install insruction to make sure they match with the new fdedup ?

shahrokhDaijavad:
@touma-I I tested the fdedup notebook by using the pip install of the latest releases of both toolkit and toolkit-transform inside the notebook and verified that it works.

touma-I (approved):

PR #857, opened 2024-12-04
Create pre-release wheels for code_profiler using transform 0.2.3.dev1 and toolkit 0.2.3.dev0

touma-I:

Need to deploy new code_profiler code as pip install


daw3rd (commented):
How about the top readme which has

```
pip3 install  'data-prep-toolkit[ray]==0.2.3.dev0'
pip3 install  'data-prep-toolkit-transforms[ray,all]==0.2.3.dev1'
```

PR #866, opened 2024-12-10
Refactor hap transform as its own dpk_ module

touma-I:

first pass at restructuring the transform built as its own module (e.g. dpk_hap) with a ray submodule (dpk_hap.ray ).

move python code under dpk_hap
move ray code under dpk_hap/ray


shahrokhDaijavad:
@touma-I I fixed the README file.

PR #886, opened 2024-12-17
refactor tokenization transform as its own named dpk_ module

matouma:
First pass at restructuring the transform built as its own module (e.g. dpk_tokenization) with a ray submodule (dpk_tokenization.ray ).

move python code under dpk_tokenization
move ray code under dpk_tokenization/ray


shahrokhDaijavad (approved):
README file looks good to me.

dangxuanhong (approved):
The re-structuring looks good to me! Also, using `>=` is better than using `==` for specifying required version in most requirements.txt. Thanks

PR #893, opened 2024-12-18
FDedup Refactored as its own dpk_ module

matouma:


shahrokhDaijavad (approved):
The latest README and makefile look good to me.

cmadam (approved):
It might be good to be consistent - if the base image is an argument in the Ray Dockerfile, then it should also be an argument in the Spark Dockerfile, or the other way around.

touma-I:
> It might be good to be consistent - if the base image is an argument in the Ray Dockerfile, then it should also be an argument in the Spark Dockerfile, or the other way around.

Thanks @cmadam . Actually, the plan now is to remove it from ray Dockerfile in a separate PR. I want to get this iteration done before starting some discussions on this.

PR #916, opened 2025-01-06
Refactoring code profiler transform to new pythonic code layout #913

pankajskku:
Refactoring code profiler transforms to new pythonic code layout to fix file path issue in the DPK package.

Issue Description
While working with the .whl distribution of the dpk_code_profiler_transform_python package (version 0.2.3.dev0), I noticed that certain .csv files required by the code are missing from the installed package. Specifically, the following file paths could not be resolved:
semantic-ruleset/ikb_model.csv
semantic-ruleset/null_libs.csv
When attempting to run the code, it fails with a FileNotFoundError because these files are not included in the wheel distribution.
Impact
The missing .csv files are essential for the proper functioning of the transform. Without them, the code cannot execute as intended, resulting in runtime errors.
Possible Cause
It seems that while these files are present in the source code, they are not being included in the .whl file during the packaging process. Locally I verified and these csv files are present in the files dpk_code_profiler_transform_ray-0.2.3.dev0.tar.gz and the dpk_code_profiler_transform_ray-0.2.3.dev0-py3-none-any.whl files created in dist folder after make build.
When I installed the packages for these transforms to use the DPK-inner venv, a folder named semantic-ruleset was created, but it does not contain the .csv files. To address this issue, I suggest ensuring that the .csv files are explicitly included in the wheel distribution.


transforms/code/code_profiler/dpk_code_profiler/output_data-prep-kit.sl.cloud9.ibm.com_20250106-003334-783.html
touma-I, line 1: Does this file need to be in git ?

touma-I (commented):
@pankajskku Very nice work and very thorough! There are a number of files that start with output_data-prep-kit.sl.cloud9.ibm.com*.json and *.html. Do they need to be checked to git? Thanks

transforms/code/code_profiler/requirements.txt
touma-I, line 1: is there a chance to relax some of the dependencies using <= and/or >=  ? 

Also is it possible not to have a tag for `pandas` and `request` but simply use the latest ?

transforms/code/code_profiler/dpk_code_profiler/output_data-prep-kit.sl.cloud9.ibm.com_20250106-003334-783.html
pankajskku: These are the sample output files that I kept for reference.

transforms/code/code_profiler/requirements.txt
pankajskku: I need to test them individually, hoping it will not break the application. For tree-sitter related packages, it will be an issue, because I have tested it previously.

Due to additional PRs on top of this one, my fork is broken now. I could not push the change for pandas and requests.

pankajskku: In current requirements, only these pydantic and pydantic_core are causing the issue with "==". I changed them to pydantic>=2.7.4, pydantic_core>=2.18.4.
Also used the latest versions for pandas and parameterized.

touma-I (approved):
Looks good. I'd like to merge this to have a baseline and do any additional debuging/enhancement in a follow-up PR. cc: @pankajskku

PR #929, opened 2025-01-08
Added TRANSFROM_NAME to docker build arg

matouma:
…f docker file template

While doing the refactoring of the transforms, we observed that now, most of the Dockerfile have the same scripts except for the module name. This change adds TRANSFORM_NAME build arg and dockerfile templates for python, ray and spark so transforms owner can simply copy/paste the template. In most cases, the transform owner will not need to modify the file unless their transform require additional non-common components that are unique to the run time environment for this transform

[Feature] Restructure transforms as their own modules #774


transforms/Dockerfile.python.template
```diff
@@ -0,0 +1,33 @@
...
+ARG TRANSFORM_NAME
+
+# Copy and install data processing libraries 
+# These are expected to be placed in the docker context before this is run (see the make image).
+COPY --chown=dpk:root data-processing-dist data-processing-dist
+RUN  pip install data-processing-dist/${DPK_WHEEL_FILE_NAME}
+
+# END OF STEPS destined for a data-prep-kit base image 
+
+COPY --chown=dpk:root dpk_${TRANSFORM_NAME}/ dpk_${TRANSFORM_NAME}/
```
roytman: I think it should be `--chown=dpk:users`

transforms/Dockerfile.spark.template
```diff
@@ -0,0 +1,34 @@
...
+# Copy and install data processing libraries 
+# These are expected to be placed in the docker context before this is run (see the make image).
+COPY --chown=spark:root data-processing-dist data-processing-dist
+RUN  pip install data-processing-dist/${DPK_WHEEL_FILE_NAME}[spark]
+
+
+# Install project source
+
+## Copy the python version of the tansform
+COPY --chown=spark:root dpk_${TRANSFORM_NAME}/ dpk_${TRANSFORM_NAME}/
```
roytman: as above ` --chown=spark:root`

transforms/Dockerfile.ray.template
```diff
@@ -0,0 +1,31 @@
+ARG BASE_IMAGE=docker.io/rayproject/ray:2.24.0-py310
```
roytman, line 1: we use this BASE_IMAGE mechanism with Ray only, should it be extended to Python and Spark as well?

transforms/Dockerfile.python.template
touma-I: Done. Thanks

transforms/Dockerfile.spark.template
touma-I: Done. Thanks

transforms/Dockerfile.ray.template
touma-I: @roytman I'd like to defer this to another PR when I do the full cleanup of the Makefile and other CI/CD related artifacts. https://github.com/IBM/data-prep-kit/issues/930 For this PR, I just want to setup the templates so developers of new transforms don't have to edit the Dockerfiles and simply use the template as is unless they have some complex custom requirements. I will need to do another round on the templates when I do the Makefile here https://github.com/IBM/data-prep-kit/issues/930

roytman (approved):

PR #934, opened 2025-01-12
Support the case where an arbitrary user id runs the ray docker images

revit13:
This PR Support the case where an arbitrary user id runs the ray docker images as is done in openshift 4.16+
The solution is described in https://docs.openshift.com/container-platform/4.17/openshift_images/create-images.html#use-uid_create-images


roytman (approved):
LGTM.

please mention in the PR description that we need it for execution on OpenShift clusters.

Issue #909, opened 2025-01-02
[Bug] path issues when running superworkflow pipeline sample for kfp v2


juancappi:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP workflows

### What happened + What you expected to happen

When running super-pipeline example from [`examples/kfp-pipelines/superworkflows/ray/kfp_v2/`](https://github.com/IBM/data-prep-kit/tree/6a06d8763ede388a93e956635af530df3494a9c8/examples/kfp-pipelines/superworkflows/ray/kfp_v2#how-to-compile-the-superpipeline) (only after fixing the issues proposed in PR #906), the build fails due to incorrect relative path computation:
```
FileNotFoundError: [Errno 2] No such file or directory: '../../../../kfp/kfp_ray_components/createRayClusterComponent.yaml'
```

### Reproduction script

```
make workflow-build
...
Traceback (most recent call last):
  File "[...]/git/ai-foundation/data-prep-kit/examples/kfp-pipelines/superworkflows/ray/kfp_v2/superpipeline_noop_docId_v2_wf.py", line 18, in <module>
    from universal.doc_id.kfp_ray.doc_id_wf import doc_id
  File "[...]/git/ai-foundation/data-prep-kit/transforms/universal/doc_id/kfp_ray/doc_id_wf.py", line 94, in <module>
    create_ray_op = comp.load_component_from_file(component_spec_path + "createRayClusterComponent.yaml")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/git/ai-foundation/data-prep-kit/transforms/venv/lib/python3.11/site-packages/kfp/components/load_yaml_utilities.py", line 53, in load_component_from_file
    with open(file_path, 'r') as component_stream:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '../../../../kfp/kfp_ray_components/createRayClusterComponent.yaml'
make[2]: *** [.workflows.compile-pipeline] Error 1
make[1]: *** [superpipeline_noop_docId_v2_wf.yaml] Error 2
make: *** [workflow-build] Error 2
```
If I update the relative path in `doc_id_wf.py` to include the extra level `../`:
[...]

PR #935, opened 2025-01-13
Fix path issues when running superworkflow pipeline sample for kfp v2

revit13:

Currently the path to kfp component specifications files in the python workflow file is hard coded.
This causes an error when compiling super pipeline v2 reported in issue #909
To avoid this. the path is saved in an env variable with a default value.

#909


roytman (commented):
Q. should we use `../../../../kfp/kfp_ray_components/` as a shared constant?

revit13:
> `../../../../kfp/kfp_ray_components/`

Good Idea, I can add a new directory named `compile_utils` under `shared_workflow_support`

revit13:
> Q. should we use `../../../../kfp/kfp_ray_components/` as a shared constant?

Done. Thanks

roytman (approved):
LGTM

PR #945, opened 2025-01-15
Cleanup documentation for 1.0.0

touma-I:

Cleanup documentation following switch over to 1.0 using the refactored transform structure and simplified API


touma-I:
@swith005 Shalisha, it wold be great if you can comment on this PR. Primarily the README.md and the doc/quick-start/contribute-your-own-transform.md. key question for you to answer: If you had this available to you before you start, would that have helped ? what do you think it is still missing ? Thanks a lot.

swith005:
@touma-I thanks for adding a read me doc for contributing transforms. this is very helpful! please find below some comments after reviewing: 
1.
**Typos: there are several sentences with typos:** 
- The user defined function for is tutorial is a simple annotator that uses the python hashlib package:
- We will described the contents of the file in 2 sections:
- TransformUtils provides a number of shortcuts commonly used by transfdorms for data conversions, table manipulations, etc.
- This file implements a serie of user defined validation steps that must be executed prior to invoking the transform.
- PythonTransformLauncher, PythonTransformRuntimeConfiguration Implemented by the framework for orchestration the execution of the transform
- Some of the key classes indetified in this import list include:
- This file implements the necessary API for integrating the digest transfromw with the ray orchestrator
- Create a virtual environment and run the transform against the input data to produce the expected output data. This will be used by the CI/CD to verify that the logic of the transform always produces the same output for a give input.
- Developers have some freedom in designing their unit tests. in this section we should how developers can use the test fixture defined in the framework to rapidly create unit tests
- The repo implements a rich set of functionality for setting up the environment, running unit tests, publish the transforms to pypi, building the transforms as part of a docker image and running it with Kubeflow. For the prupose of this section, we will explore only a portion of the capabilities for support this initial phase of the implementation
- The Makefile has a number of predefined target that will be useful for testing and publish the trnasform. To get a list of target, run the following command from the digest folder:
- edit transforms/pyproject.toml and add the requirements.txt for the module and the name of the module and it package location.

2.
**Step 3: Implement TransformConfiguration**
This section was somewhat confusing to understand, and I would've been lost had it not been for prior experience with inner. It would be more helpful to provide more detail for what runtime does in relation to cli arguments and usage. 

everything else I believe is covered efficiently~

shahrokhDaijavad (approved):
LGTM.

PR #946, opened 2025-01-16
Update README.md in html2parquet transforms

Sri-Krishna-V:
Change the notebook link properly

The changes are needed so that, the link can lead directly to the notebook and for no further confusion


touma-I:
Thank you @Sri-Krishna-V for reporting this. We had another PR that was merged this morning addressing the exact same issue. Please update your branch from dev and let me know if I should close this PR. Thanks again for your contribution.

PR #959, opened 2025-01-22
 Updated the score to a rank

AnLiGentile:
The returned similarity score did not make much sense from a usability perspective.
Results should be interpreted as a ranked list, wher the first one is the most relevant.


AnLiGentile:
still didn't manage to fix my signing off properly

PR #966, opened 2025-01-24
refactored profiler transform

matouma:

Refactored like all other transforms


shahrokhDaijavad (approved):
I have done various fixes to the README file and checked the two notebook files that look good to me.

Issue #972, opened 2025-01-26
[Feature] data preprocessing code for finetuning


PoojaHolkar:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Other

### Feature

To fine-tune using DPK and make the process seamless, we currently lack a domain-specific notebook or reference code for implementation.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

shahrokhDaijavad:
PR has been merged.

PR #973, opened 2025-01-26
DPK processing of text data for finetuning 

PoojaHolkar:

This code demonstrates how to run a domain-specific use case using DPK, process the data accordingly, and fine-tune it to achieve improved domain-specific results.

data preprocessing code for finetuning #972


shahrokhDaijavad:
@PoojaHolkar In the latest notebook (not in the PR yet), in the first cell, when you are explaining the steps, please change the chunking and dedup sequence, to be consistent with the sequence that the notebook implements later. In the PR notebook, to run on Google Colab, you are downloading all 3 files from the repo correctly, but in the latest notebook, you are downloading only one wrong file (Invoice.pdf) and not all 3 files. 

Here is an important change I would like you to make if you want to keep the option of running in Google Colab in the notebook. We have decided to switch to a native Python command for downloading files and not use wget anymore (wget usually requires installation, though the Google Colab environment has it).  You can see an example of how we do this here: https://github.com/IBM/data-prep-kit/blob/dev/examples/notebooks/Run_your_first_transform_colab.ipynb in which we use urllib and shutil libs to do this. Can you please download the 3 files you need for Google Colab this way?

PoojaHolkar:
@shahrokhDaijavad made the required changes for the transform sequence and to pull data files without using wget. Please verify.

shahrokhDaijavad:
Thanks for making the changes, @PoojaHolkar. I have verified that the latest Notebook runs successfully in my local environment. However, as I reviewed the notebook with @touma-I, he asked for a few more changes, before we can merge:
1) In cell one, you need only one pip install, i.e., installing the language **includes** pii_redactor, so one pip install, as follows:
`!pip install 'data-prep-toolkit-transforms[language]'`  should install everything. We now have a major release 1.0.0 and the version number does not need to be specified either. Can you please test this after the change? 
2) Please fully get rid of cell 2 "Figure out Runtime Environment" which tests whether the environment is Google Colab or not. We don't use the result anywhere later. 
3) Cell 3 "Download Data if running on Google Colab," should be a cell in which we download the 3 files we need, BOTH for running in the local environment and in the Google Colab environment.  So, the title of the cell should be "Download Data". We have now created a permanent place where we keep test files for various use cases under the examples subdirectory, including the three that you need. So, the URL that points to these files in Cell 3 will change to that permanent place in the repo. 
For example, the place for your one of your data files is: 
https://github.com/IBM/data-prep-kit/blob/dev/examples/data-files/fine-tuning/language/MSA-DPK-1.pdf

This way, you can get rid of the input-data folder in your PR, and the files will always be downloaded from the repo at runtime. 
4) Now that you have this notebook that uses both HAP and PII, is there any reason that we cannot get rid of the two older notebooks for HAP and PII that you and @AishaDarga developed? Those notebooks under examples use the old APIs and are confusing for our repo users. If there is no good reason, can you please delete both those folders in this PR?

PoojaHolkar:
@shahrokhDaijavad I made the required changes. The data is now getting pulled from the new path for input files. On point 4  - the idea was if the client wants only to use a specific transform like PII or HAP they can make use of the direct examples. This code is domain-specific and demonstrates how multiple transforms can be put to use. Both serve distinct purposes.
@agoyal26 please confirm

agoyal26:
@PoojaHolkar I agree with Shahrokh that we should update HAP and PII notebooks to also use newer simplified API structure in case we want to keep them. 
@shahrokhDaijavad I was thinking of structuring our example note books with these tags to be linked from main readme ..
Do think we can keep PII and HAP specifically calling out that we are using customer service data - where these are more relevant
Screenshot 2025-01-29 at 5 47 50 PM

PoojaHolkar:
@agoyal26 updating the notebooks with latest apis can be done.

shahrokhDaijavad:
Thank you, @PoojaHolkar for making all the changes. Looks good! I am approving the PR. 
On point 4, you may know that we already have PII and HAP notebooks with the **new APIs** in their corresponding transform folders. However, the ones that you and Aisha developed create a better story with the CSV and PDF files that you start from and are nicely documented. We will keep them if they are updated using the new APIs. Please remember that your  data files should be placed in the new data-files subdirectory under examples and new subdirectories HAP and PII in there.
 
@agoyal26 I like your screenshot, but as we discussed today, this should not be in the main README file. The main README file should have a single sentence with a link to a new README file in the examples folder having a structure like what you are showing.

shahrokhDaijavad (approved):
LGTM.

PoojaHolkar:
sure @shahrokhDaijavad thank you. We will work on updating those notebooks.

examples/notebooks/fine tuning/language/fine-tune-language.ipynb
touma-I, line 1: cell 5:
`os.makedirs(input_folder, exist_ok=True)`
is not needed since the same is done in cell 3:
`shutil.os.makedirs("input-data", exist_ok=True)`

touma-I, line 1: In cell 4, I don't think you need those lines:
```Python
output_folder = os.path.join( "output")
local_conf = {
    "input_folder": input_folder,
    "output_folder": output_folder,
}
```

touma-I (changes requested):

shahrokhDaijavad:
Thanks, @PoojaHolkar I just tested the latest version and everything worked. 
@touma-I The cleanup you asked for has been done. I approve again and ask you to merge.

shahrokhDaijavad (approved):
LGTM

touma-I (approved):

PR #992, opened 2025-01-28
added Optional step for enabiling kfp

touma-I:

Enhance the tutorial to add the steps that the developer has to do in order to enable KFP pipelines for this transform.


shahrokhDaijavad (approved):
There are a couple of typos that I will fix, after the merge, by doing a PR. Instead of fixing now in a rush, as I copy this text in the tutorial tomorrow, I can do a better job.

Issue #989, opened 2025-01-28
[Bug] FDedup Fails on Windows


touma-I:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/universal/fdedup

### What happened + What you expected to happen

When running FDedup on windows, I get an error:
13:50:52 ERROR - Exception during execution 'processing_time': None
13:50:52 INFO - Completed execution in 0.0 min, execution result 1
13:50:52 ERROR - ClusterAnalysis failed with status 1, aborting ...

A complete trace can be found here:

https://github.com/IBM/data-prep-kit/blob/v1.0.0.post1/transforms/transforms-1.0-lang-Windows.ipynb

### Reproduction script

Run the following notebook on Windows:

https://github.com/IBM/data-prep-kit/blob/v1.0.0.post1/transforms/transforms-1.0-lang-Windows.ipynb

### OS

Windows WSL

### Python

Other

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

cmadam:
@touma-I  commit https://github.com/IBM/data-prep-kit/pull/1003/commits/86bbdedc66d089b6f37136dc2f8759fff5d023cd is the first try to fix windows problem. Can you please test it using the local file system and see if it works?  I still think it will not work for s3 storage; is there a way, inside a transform, to determine whether it uses the local file system or S3

touma-I:
@cmadam with the lastest PR, I get the following error below. It seems the filename issue was solved but now another error pops up on windows:
```
17:40:40 INFO - Starting SignatureCalculation step
17:40:40 INFO - Got parameters for SignatureCalculation
17:40:40 INFO - minhash parameters are : {'document_id_column': 'int_id_column', 'contents_column': 'contents', 'seed': 42, 'num_permutations': 112, 'jaccard_similarity_threshold': 0.75, 'word_shingle_size': 5, 'num_bands': 14, 'num_minhashes_per_band': 8, 'num_segments': 1, 'shingle_option': 'word'}
17:40:40 INFO - data factory scdata_ is using local configuration without input/output path
17:40:40 INFO - data factory scdata_ max_files -1, n_sample -1
17:40:40 INFO - data factory scdata_ Not using data sets, checkpointing False, max files -1, random samples -1, files to use ['.parquet'], files to checkpoint ['.parquet']
17:40:40 INFO - pipeline id pipeline_id
[...]

Issue #982, opened 2025-01-27
[Bug] FDedup failing with latest release mmh3==5.1.0


touma-I:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/universal/fdedup

### What happened + What you expected to happen

When building the transform with current requirements.txt, we are pulling in the latest release of mmh3 that was published to pypi on Jan 25 and causing the UT of fdedup to fail.

### Reproduction script

To produce the error, run test-src with mmh3>=4.1.0

To prevent the error, modify transforms/universal/fdedup/requirements.txt to add:
mmh3>=4.1.0, <=5.0.1

### Anything else

Also investigate if relaxing polars can be achieved. It is currently set to polars==1.9.0 and the preferred approach would be to use polars>=1.9.0

### OS

Ubuntu

### Python

3.11.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

touma-I:
Stop gap in current dev branch:  mmh3>=4.1.0, <=5.0.1

Long term, investigate how we can make it work for latest release of mmh3

cmadam:
I have fixed the code so that it works with any version of `mmh3`

cmadam:
The code  worked with `polars` releases up to, and including, `1.9.0`.  Then, it stopped working with releases `1.10.0`, `1.11.0`, and `1.12.0`.  The problem went away and the tests passed successfully in all the subsequent releases of `polars` (`1.13.0`, `1.13.1`, `1.14.0`, `1.15.0`, `1.16.0`, `1.17.0`, `1.18.0`, `1.19.0`, `1.20.0`, and the latest, `1.21.0`).  So I changed the `requirements.txt` file with:
```
polars>=1.13.0
````
to eliminate all the problematic versions (`1.10.0 - 1.12.0`)

PR #1003, opened 2025-01-30
Fdedup package versioning and windows fixes

cmadam:
* Fix incompatibilities between fuzzy dedup and releases of `mmh3` and `polars` packages
* Fix failing run on windows

#982 
#989


transforms/universal/fdedup/requirements.txt
```diff
@@ -1,9 +1,9 @@
 pyyaml>=6.0.2
 boto3>=1.34.69
 kubernetes>=30.1.0
-polars==1.9.0
+polars>=1.13.0
```
touma-I: Please consider relaxing this to polars>=1.9.0. I have other modules that depend on polars and I want to make sure that all packages work together and meet the minimum requirements. Would your code break with 1.9.0 ?

cmadam: As I mentioned in issue #982, fuzzy dedup does not work with the following `polars` releases:
* 1.10.0
* 1.11.0
* 1.12.0
It works with any release before, and including `1.9.0` and after, or including 1.13.1.  As long as you can exclude these three particular versions from polars, we are good.

touma-I (approved):

PR #1016, opened 2025-02-05
Cargo fix

swith005:
changes made to :

 - install cargo on init (build for user's architecture) 
 - removed unoptimized dedup functions
 - fix bug reporting duplicate documents
 - use dpk_utils for logging 
 - change default num_threads used
 - update readme to remove manual cargo install


transforms/universal/rep_removal/dpk_rep_removal/dedup_pq_level.py
touma-I, line 1: No longer being referenced: line 27
```Python
import dpk_rep_removal.utils
```
Where/how this is being used ? line 31
```Python
run_in_OCP = True
```

No longer needed!? lines 84-85
```Python
pre_sep = b"\xff\xff"
post_sep = b""
```

touma-I, line 1: Can we delete the various print statements instead of commenting them out ?

touma-I, line 1: Consider deleting: lines 210-215
 ```Python
 # print ('############# Number of duplicate made from two subsequent documents: ', count_between_docs)
    # print ('############# Number of duplicate made from two subsequent documents: ', duplicate_between_docs)

    # df_dict = pd.DataFrame(remove_ex)
    # print(remove_ex)
    # return remove_ex```

transforms/universal/rep_removal/README.md
touma-I, line 1: THANK YOU! This is so much easier now to work with now!

transforms/universal/rep_removal/dpk_rep_removal/runtime.py
```diff
@@ -85,7 +85,7 @@ def add_input_params(self, parser: ArgumentParser) -> None:
             "--rep_removal_num_threads",
             type=str,
             required=False,
-            default="4",
+            default=str(cpu_count(logical=False)),
```
touma-I: Can we also change the line in the **notebook** example to show rep_removal_num_threads=cpu_count() or simply remove it that setting from the **notebook** ?

Folks will be tempted to copy/paste from the notebook...

transforms/universal/rep_removal/dpk_rep_removal/transform.py
```diff
@@ -9,9 +9,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ################################################################################
-import logging
+
 import os
+import subprocess
 import tempfile
+import datetime
```
touma-I: Not being used. Consider deleting
```suggestion
```

transforms/universal/rep_removal/dpk_rep_removal/runtime.py
```diff
@@ -85,7 +85,7 @@ def add_input_params(self, parser: ArgumentParser) -> None:
             "--rep_removal_num_threads",
             type=str,
```
touma-I: Would it make sense to make this int ?
```suggestion
            type=int,
```

transforms/universal/rep_removal/dpk_rep_removal/runtime.py
touma-I: got for type int and get rid of str()
```suggestion
            default=cpu_count(logical=False),
```

transforms/universal/rep_removal/dpk_rep_removal/runtime.py
touma-I, line 1: Would it be more intuitive to make all those of type=int (instead of str) ?
            "--rep_removal_length_thresh",
            "--rep_removal_frequency_threshold",
            "--rep_removal_num_threads",
            "--rep_removal_num_cpus",

touma-I (changes requested):
thanks @swith005 . Please do a pass on the UT files in test folder, I think they are still using rep_removal_num_threads='1' (for consistency, this should use an int such as rep_removal_num_threads=1)

touma-I (approved):

touma-I:
need to run a few more tests to understand the effect of using cpu_count(). Other than that, this looks ready to merge. Thanks

Issue #985, opened 2025-01-28
[Bug] The S3 secret name is hardcoded in the KFP library


roytman:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Library/kfp

### What happened + What you expected to happen

Due to https://github.com/kubeflow/pipelines/issues/10914, we cannot provide secret names as pipeline arguments, so we agreed to a workaround that the secret name will be hard-coded.
The secret names should be in pipeline, so users will be able to overwrite them. Currently it is in the [library](https://github.com/IBM/data-prep-kit/blob/5f86475dd3521003f06b609363ec216aaa89a071/kfp/kfp_support_lib/kfp_v2_workflow_support/src/workflow_support/compile_utils/component.py#L134), so users cannot change it.

### Reproduction script

.

### OS

Red Hat Enterprise Linux (RHEL), Ubuntu

### Python

3.11.x

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

PR #1030, opened 2025-02-10
[KFP v2] Fix the S3 secret name is hardcoded in the KFP library.

revit13:

FIX - Due to https://github.com/kubeflow/pipelines/issues/10914, we cannot provide secret names as pipeline arguments, so we agreed to a workaround that the secret name will be hard-coded.
The secret names should be in pipeline, so users will be able to overwrite them. Currently it is in the [library](https://github.com/IBM/data-prep-kit/blob/5f86475dd3521003f06b609363ec216aaa89a071/kfp/kfp_support_lib/kfp_v2_workflow_support/src/workflow_support/compile_utils/component.py#L134), so users cannot change it.

#985


kfp/pipeline_generator/single-pipeline/templates/simple_pipeline.py
```diff
@@ -221,7 +221,15 @@ def {{ pipeline_name }}(
...
         )
         ComponentUtils.add_settings_to_component(execute_job, ONE_WEEK_SEC)
-        ComponentUtils.set_s3_env_vars_to_component(execute_job, data_s3_access_secret)
+        if os.getenv("KFPv2", "0") == "1":     
+            from kfp import kubernetes
+            
+            # FIXME: Due to kubeflow/pipelines#10914, secret names cannot be provided as pipeline arguments.
+            # As a workaround, the secret name is hard coded.
+            env2key = ComponentUtils.set_secret_key_to_env()
+            kubernetes.use_secret_as_env(task=execute_job, secret_name="s3-secret", secret_key_to_env=env2key)
```
roytman: I suggest defining a constant at the start of the file and using it here. It will simplify for users to update it.

roytman (approved):

touma-I:
@revit13 Can you merge with the latest from dev and let me know if this fixes the problem with the kfp1 test for lang_id ? I had to add a read-only secret for hugging face, exposed that in the workflow as an env. variable and then use it in the lang_id_wf.py (more details here: https://github.com/IBM/data-prep-kit/pull/1068). I am looking at code_quality now.

roytman (commented):
I don't think we need to update the examples/kfp-pipleines/superworkflow/ray/kfpv1

revit13:
> I don't think we need to update the examples/kfp-pipleines/superworkflow/ray/kfpv1

@roytman Fixed Thanks

Issue #1009, opened 2025-02-04
Tokenizing parquet files to arrow tables


shahrokhDaijavad:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

In the GneissWeb recipe, after the rep-removal transform and before using the extreme-tokenizer transform, we need to convert parquet files to arrow tables. It seems this work was one internally as a transform by @santoshborse called tokenization2arrow. 
@santoshborse Is it possible for you to bring your transform to the open repo? 

cc: @touma-I @Hajar-Emami 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

shahrokhDaijavad:
Completed with merging of PR #1033

PR #1033, opened 2025-02-10
Tokenization2Arrow - New Transform to tokenize data and generate .arrow and metadata files

santoshborse:

Tokenization2Arrow - New Transform to tokenize data and generate .arrow and metadata files

https://github.com/IBM/data-prep-kit/issues/1009


shahrokhDaijavad:
Thank you very much, @santoshborse. I tested the transform by running `make run-cli-sample-python` successfully. 
There are some things missing like a `test` directory that we use for CI/CD testing and I will make the README more consistent with the other READMEs, but for now, one small change in README:

`make run-cli-sample` => `make run-cli-sample-python` 
and another urgent request for a simple notebook that @Hajar-Emami  can use to mimic in her GneissWeb recipe notebook. 
An example of such a minimum notebook is this one: 
https://github.com/IBM/data-prep-kit/blob/dev/transforms/universal/tokenization/tokenization.ipynb
i.e.,
1) Show the CLI option table that you have in README
2) from dpk_tokenization2arrow.transform_python import Tokenization2Arrow
3) Tokenization2Arrow(input_folder= "test-data/ds02/input",
output_folder= "output", 
.... other CLI arguments).transform()
4) import glob
glob.glob("output/*")

cmadam (commented):
One problem with this transform is that it has no tests:
```
$ make test-src
...
source venv/bin/activate;       \
export PYTHONPATH=../src:../: ;  \
cd test; pytest -s .
/bin/bash: line 3: cd: test: No such file or directory
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.11.11, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/cma/de/data-prep-kit/transforms
configfile: pyproject.toml
plugins: cov-6.0.0, anyio-4.8.0
collected 0 items                                                                                                                                                                                                                           

=========================================================================================================== no tests ran in 0.01s ===========================================================================================================
make: *** [../../../.make.defaults:442: .defaults.test-src] Error 5
```
Please add some tests to the transform.

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform.py
```diff
@@ -0,0 +1,167 @@
...
+        in the output table, along with the document id and token count stored respectively in `document_id` and `token_count` column.
+        """
+        self.data_access: DataAccess = config.get("data_access")
+        super().__init__(config)
+        # we will tokenize the content only if language column value is in lang_filters
+        self.output_folder = self.data_access.get_output_folder()
+        self.output_folder = self.output_folder if self.output_folder.endswith("/") else f"{self.output_folder}/"
+        self.logger.info(f"Tokenizer {config['tokenizer'] = } loaded.")
+
+    def transform_to_arrow(
```
cmadam: I would suggest that you define this function as a private function (`_transform_to_arrow`), because it is not being used anywhere else outside this class, and because it is not part of the `AbstractTAbleTransform`, or the `TokenizationTransform` classes.

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_python.py
```diff
@@ -0,0 +1,53 @@
...
+        self.params = {}
+        for key in kwargs:
+            self.params[key] = kwargs[key]
+        # if input_folder and output_folder are specified, then assume it is represent data_local_config
+        try:
+            local_conf = {k: self.params[k] for k in ("input_folder", "output_folder")}
+            self.params["data_local_config"] = ParamsUtils.convert_to_ast(local_conf)
+            del self.params["input_folder"]
+            del self.params["output_folder"]
+        except:
```
cmadam: Not sure why `data_local_config` is set manually in here. `data_local_config`, or `data_s3_config` should be setup automatically by the dpk framework, there is no reason to duplicate the code here.

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_ray.py
```diff
@@ -0,0 +1,62 @@
...
+        except:
+            pass
+        try:
+            worker_options = {k: self.params[k] for k in ("num_cpus", "memory")}
+            self.params["runtime_worker_options"] = ParamsUtils.convert_to_ast(worker_options)
+            del self.params["num_cpus"]
+            del self.params["memory"]
+        except:
+            pass
+
```
cmadam: Same observation as above.  `data_local_config`, `runtime_worker_options` should all be parsed and setup by the dpk framework, no need to reprocess here.

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform.py
santoshborse: ok, make sense

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_python.py
santoshborse: I think it is there to support cases where instead of passing json `data_local_config`, it just pass directly `input_folder` and `output_folder`, I am following `Tokenization` transform code.

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_ray.py
santoshborse: Same as above , this is created to handle case where user directly pass cpu and memory param.

transforms/universal/tokenization2arrow/Dockerfile.python
touma-I, line 1: Can you use the content from here instead? https://github.com/IBM/data-prep-kit/blob/tokenization2arrow_transform/transforms/Dockerfile.python.template

transforms/universal/tokenization2arrow/Dockerfile.ray
touma-I, line 1: Can you use the content from here ? https://github.com/IBM/data-prep-kit/blob/tokenization2arrow_transform/transforms/Dockerfile.ray.template

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_ray.py
touma-I, line 1: For consistency, it may be better to use ray.transform. This will make it easier for maintaining the module

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform.py
```diff
@@ -0,0 +1,168 @@
...
+        with pa.BufferOutputStream() as bos:
+            with pa.ipc.new_file(bos, schema=schema) as writer:
+                for token_line in tokens_column:
+                    writer.write(pa.record_batch([pa.array(token_line)], schema=schema))
+        # self.data_access.save_file(arrow_file_path, bos.getvalue().to_pybytes())
+
+        # TODO: check if we should add anything to tokenization_metadata
+        return [(bos.getvalue().to_pybytes(), ".arrow")], tokenization_metadata
+
+    def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
```
touma-I, line 101: Not clear why this is implementing transform_binary() and not transform(). I think the code structure will be easier to understand/maitain if you implement transform() and then call super.transform() before calling transforms_to_arrow() .

touma-I (changes requested):
Please provide one or more Unit Test in the test folder.

transforms/universal/tokenization2arrow/requirements.txt
touma-I, line 1: Let's discuss how we can redo this one. Maybe 2 requirements.txt, one that is used as part of the packaging and one that is used for pulling the dependency on the tokenization. Also, have you considered making this module as part of the tokenization module ? Would it be easier for inheritance for this module to be an extension on the tokenization rather than its own ?

touma-I (commented):
added the module to pyproject.toml for when building the wheel.

transforms/universal/tokenization2arrow/Dockerfile.python
santoshborse: the docker file look exact same to me, the template seems to be using TRANSFORMER_NAME which I feel is not needed

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform.py
santoshborse: `transform_binary` returns - `tuple[list[tuple[bytes, str]], dict[str, Any]]:`
`transform` returns - `tuple[list[pa.Table], dict[str, Any]]:`

I am using `transform_binary` so that I can return data in `bytes` ( so that f/w can write .arrow files )

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_ray.py
santoshborse: I think because we have `transform_python` which we use as,
`from dpk_tokenization2arrow.transform_python import Tokenization2Arrow`

I makes more sense to have `transform_ray`, but if you insist I will change that

shahrokhDaijavad:
Thanks for adding the notebook, @santoshborse!

transforms/universal/tokenization2arrow/dpk_tokenization2arrow/transform_ray.py
touma-I: @santoshborse The latest convention is to use dpk_\<transform\>.runtime (instead of dpk_<transform>.transform_python) and dpk_\<transform\>.ray.runtime . It will be nice if you can stay with the convention. It makes suppport easier and there does not seem to be a good reason not to.

transforms/universal/tokenization2arrow/Makefile
touma-I, line 1: @santoshborse This file is missing the following:
TRANSFORM_PYTHON_SRC=
TRANSFORM_RAY_SRC=

see https://github.com/IBM/data-prep-kit/blob/dev/transforms/Makefile.transform.template for example. This is also related to the comment above on transform_python and transform_ray. Let's follow the convention since there is no really good reason not to.

touma-I (changes requested):

touma-I (changes requested):
@santoshborse The two additional comments I added should address the failed CI/CD. Please reach out if anything is not clear.

transforms/universal/tokenization2arrow/Makefile
santoshborse: ok, I will make changes, I followed https://github.com/IBM/data-prep-kit/blob/dev/transforms/universal/tokenization/Makefile

santoshborse: hi @touma-I I have updated the module names and Makefile as you asked.

touma-I: @santoshborse Thank you! This looks good. Sorry about the confusion. I am hoping in the next iteration we will simplify things further and get rid of a few constraints. Please stay tuned. I might also reach out to bounce off a few ideas.

touma-I (approved):

touma-I:
@shahrokhDaijavad Can you review the readme.md and  notebooks before I merge? Thanks

shahrokhDaijavad (changes requested):
@santoshborse In the two notebooks, can you please change
from dpk_tokenization2arrow.transform_python import Tokenization2Arrow
to 
from dpk_tokenization2arrow.runtime import Tokenization2Arrow

and 
from dpk_tokenization2arrow.transform_ray import Tokenization2Arrow
to
from dpk_tokenization2arrow.ray.runtime import Tokenization2Arrow



Also,
In the makefile, don't we need to make the same change for the 2 targets:
dpk_$(TRANSFORM_NAME).transform_python => dpk_$(TRANSFORM_NAME).runtime 
and 
dpk_$(TRANSFORM_NAME).transform_ray  => dpk_$(TRANSFORM_NAME).ray.runtime

After these changes, if you don't mind, I will make a few small changes in the README file myself.

santoshborse:
@shahrokhDaijavad updated the notebook and makefile, also tested locally to make sure notebook and both commands in makefile works.

transforms/universal/tokenization2arrow/Makefile
```diff
@@ -0,0 +1,35 @@
...
+# a given transform's directory tree,  so must use compatible syntax.
+#
+################################################################################
+# This defines the name of the transform and is used to match against
+# expected files and is used to define the transform's image name. 
+TRANSFORM_NAME=$(shell basename `pwd`)
+
+################################################################################
+
+TRANSFORM_PYTHON_SRC=-m dpk_$(TRANSFORM_NAME).runtime
```
touma-I: This needs double quotes:
```suggestion
TRANSFORM_PYTHON_SRC="-m dpk_$(TRANSFORM_NAME).runtime"
```

transforms/universal/tokenization2arrow/Makefile
```diff
@@ -0,0 +1,35 @@
...
+TRANSFORM_RAY_SRC=-m dpk_$(TRANSFORM_NAME).ray.runtime
```
touma-I: This needs double quotes:
```suggestion
TRANSFORM_RAY_SRC="-m dpk_$(TRANSFORM_NAME).ray.runtime"
```

touma-I (changes requested):

touma-I (approved):

shahrokhDaijavad (approved):
Thanks, @santoshborse. I made a few changes to the README file.

Issue #1048, opened 2025-02-13
[Bug] HAP transform crashes if "contents" is empty


burn2l:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### What happened + What you expected to happen

If a sample has an empty "contents" field HAP fails with:

**_Exception executing transform max() arg is an empty sequence_**

### Reproduction script

Append a 51st sample to test-data/input/test1.parquet such as:
**{"doc_id": 51, "contents": " "}**
and run pytest test/test_hap.py

### Anything else

Can be fixed by changing line 72 in _transforms/universal/hap/dpk_hap/transform.py_ to 
            doc_scores.append(max(temp, **default=0.0**))

### OS

Red Hat Enterprise Linux (RHEL)

### Python

3.11.x

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

shahrokhDaijavad:
@ian-cho Can you please make this fix too and submit a PR? Thanks.

ian-cho:
> Can be fixed by changing line 72 in transforms/universal/hap/dpk_hap/transform.py to doc_scores.append(max(temp, default=0.0))

Thanks. How about using -1 instead of 0 to indicate the content is empty? 0 has the specific meaning (NON-HAP)

Hi @burn2l, I did not quite get why you need to append {"doc_id": 51, "contents": " "} in the test file. However, from my understanding, when processing a large number of parquet files, there is a possibility that the `contents` column may be empty. Is it the same issue you were referring to?

burn2l:
I thought the hap_score range was 0.0 to 1.0 so an empty file with no HAP content should be 0.  I don't see much value in using a value outside this range to indicate that the content is empty since that fact can be easily determined by inspecting the contents.
I added the 51st line to the test data as a way to demonstrate the problem of an empty contents, and to test the fix. 
Note that #1052 is also a case of empty contents ... i.e. no tokens.

ian-cho:
@burn2l Hi, thanks for raising this issue and the problem definitely need a fix. From my perspective, two points:
- If we use 0, users processing a large number of Parquet files may see many instances with a score of 0 and mistakenly assume their files contain no HAP terms. However, this could simply be due to the existence of empty documents. This change would remove the warning functionality for users.
- The current transform behaves the same whether using 0 or -1, as we take the maximum value as the final hap score.

[...]

PR #1058, opened 2025-02-19
Update transform.py

ian-cho:
fixed error when `contents` column is empty

To handle cases where the contents field may be empty when processing a large number of Parquet files using HAP transform

#1048

### DIFF
--- a/transforms/universal/hap/dpk_hap/transform.py
+++ b/transforms/universal/hap/dpk_hap/transform.py
@@ -69,7 +69,7 @@ class HAPTransform(AbstractTableTransform):
         doc_scores = []
         for i in range(nb_doc):
             temp = [score for idx, score in zip(sent_ids, sent_scores) if i == idx]
-            doc_scores.append(max(temp))
+            doc_scores.append(max(temp, default=-1))
         return doc_scores
 
     def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Table], dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


shahrokhDaijavad (approved):
LGTM, after all the good discussion in #1048

shahrokhDaijavad:
@touma-I Please merge.

touma-I (changes requested):
Can you please add some comments explaining the logic?
1- If I am reading this correctly: We could have documents that do not have a hap score (i.e. doc_id  in the range of [0..nb_doc] but sent_ids does not contain doc_id. correct?
2- Also, are sent_ids always sequential in the range of 0..nb_doc ?

ian-cho:
@touma-I sure. You mean the comments in the code or in this issue here?

ian-cho:
@touma-I added comment for that logic.

transforms/universal/hap/dpk_hap/transform.py
```diff
@@ -69,7 +69,7 @@ def _apply_aggregate(self, nb_doc: int, sent_scores: list[float], sent_ids: list
         doc_scores = []
         for i in range(nb_doc):
             temp = [score for idx, score in zip(sent_ids, sent_scores) if i == idx]
```
touma-I, line 72:
```suggestion
           # Compile a list of hap score for each sentence in the document. IF the document is empty then zero out the    final score.
            temp = [score for idx, score in zip(sent_ids, sent_scores) if i == idx]
```

touma-I (changes requested):
Please enhance/accept the writeup explaining the line of code

touma-I (approved):

PR #1066, opened 2025-02-26
Update super pipeline kfp v2.

revit13:

This PR resolves issue encountered while testing the superpipeline in KFP v2 - it updates `PYTHONPATH` when a nested pipeline imports module from subdirectory, as seen in the ededup, fdedup and profile case.
This PR also includes code refactoring for the superpipeline, making it easier to extend with additional nested pipelines if needed.


examples/kfp-pipelines/superworkflows/ray/kfp_v2/README.md
```diff
@@ -28,6 +30,9 @@ Another useful feature of the KFP v2 is the `Json` editor for the `dict` type in
 cd examples/kfp/superworkflows/ray/kfp_v2/
 make clean
 export KFPv2=1
-export PYTHONPATH=../../../../../transforms
 make workflow-build
 ```
+
+Please be aware that in KFP v2, there is an (issue)[default pipeline input params are missing from the GUI](https://github.com/kubeflow/pipelines/issues/11515]] where default values for pipeline input parameters do not appear in the GUI,
```
roytman: I think it is relevant only for booleans with default `False` and empty strings.
BTW we can easily overcome the boolean case, just use the inverted value ;-)

examples/kfp-pipelines/superworkflows/ray/kfp_v2/superpipeline_noop_docId_ededup_v2_wf.py
```diff
@@ -16,27 +16,53 @@
...
 
 noop_image = "quay.io/dataprep1/data-prep-kit/noop-ray:latest"
 doc_id_image = "quay.io/dataprep1/data-prep-kit/doc_id-ray:latest"
+ededup_image = "quay.io/dataprep1/data-prep-kit/ededup-ray:latest"
 
+# A list of the transform names, arranged in the order of their execution within the superpipeline.
+ordered_transforms = ["noop", "doc_id", "ededup"]
 
-def _remove_unused_params(d: dict[str, Any]) -> None:
+def _remove_unused_params(d: dict[str, Any], remove_params: list = None) -> None:
```
roytman, line 32: do we really need this function, I've started to think about the separation of the input parameters use an additional prefix. Let's talk

roytman (approved):
LGTM

Issue #1038, opened 2025-02-11
[Bug] Error running lang_id and code_quality kfp pipelines


revit13:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

KFP V1 workflows 

### What happened + What you expected to happen

running `make workflow-test` under `language/lang_id` gives the following error:
```
(orchestrate pid=273, ip=10.244.2.16) 10:35:49 INFO - Cluster resources: {'cpus': 4, 'gpus': 0, 'memory': 12.0, 'object_store': 3.1541931135579944}
(orchestrate pid=273, ip=10.244.2.16) 10:35:49 INFO - Number of workers - 2 with {'num_cpus': 0.8, 'max_restarts': -1} each
(RayTransformFileProcessor pid=273, ip=10.244.1.19) 10:35:50 ERROR - Exception creating transform  401 Client Error. (Request ID: Root=1-67aa4706-032e880645c8d1a22105e459;60b4b4c8-3d63-4ea1-97bb-42a7c98c2028)
(RayTransformFileProcessor pid=273, ip=10.244.1.19)
(RayTransformFileProcessor pid=273, ip=10.244.1.19) Repository Not Found for url: https://huggingface.co/facebook/fasttext-language-identification/resolve/main/model.bin.
(RayTransformFileProcessor pid=273, ip=10.244.1.19) Please make sure you specified the correct `repo_id` and `repo_type`.
(RayTransformFileProcessor pid=273, ip=10.244.1.19) If you are trying to access a private or gated repo, make sure you are authenticated.
(RayTransformFileProcessor pid=273, ip=10.244.1.19) Invalid credentials in Authorization header
(RayTransformFileProcessor pid=273, ip=10.244.1.19) Exception raised in creation task: The actor died because of an error raised in its creation task, ray::RayTransformFileProcessor.__init__() (pid=273, ip=10.244.1.19, actor_id=777db29ec2b2a2ac5d1202fd02000000, repr=<data_processing_ray.runtime.ray.transform_file_processor.RayTransformFileProcessor object at 0x7f45b4d5af20>)
(RayTransformFileProcessor pid=273, ip=10.244.1.19)   File "/home/ray/anaconda3/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
(RayTransformFileProcessor pid=273, ip=10.244.1.19)     raise HTTPError(http_error_msg, response=self)
[...]

PR #1083, opened 2025-03-01
use None value rather than None string

matouma:

Default HF token value should be set to None for V1. Currently it uses  "Null" string

https://github.com/IBM/data-prep-kit/issues/1038


revit13 (approved):

transforms/code/code_quality/kfp_ray/code_quality_wf.py
```diff
@@ -27,7 +27,7 @@
 PREFIX: str = ""
 
 task_image = "quay.io/dataprep1/data-prep-kit/code_quality-ray:latest"
-
+HF_TOKEN= "None" if os.getenv("KFPv2", "0") == "1" else None
```
revit13, line 33: Consider adding a comment about the reason it is needed.

PR #1084, opened 2025-03-02
Share secret securrely across fork

matouma:

Modify KFP workflow so we can securely share secret across forks- Initially this is being tested with kfp workflows that need the HF Token but will be extended to other workflows as well until the code get cleaned up over time.

The intent is to use pull_request_target to trigger the workflow. This will allow us to expose the secrets and then do a context switch to the new branch. However, before running the workflow, we verify that the user that is triggering the workflow is authorized to do so.


roytman (approved):
LGTM

Issue #1139, opened 2025-03-15
[Bug] Filter transform should not return an empty output table after filtering


klwuibm:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

transforms/filter

### What happened + What you expected to happen

In our recent data operations, we notice that after the filtering pipeline, there are a large number of empty parquet files produced in the output folder. Because the number of empty parquet files in that particular use case is really large, in the order of millions, it caused very challenging errors on the downstream operations.  It took several days to realize that the large number of empty parquet files was the problem.

### Reproduction script

Looking at the code in filter transform, this can be fixed by simply doing a check of the size of the `filtered_table_cols_dropped` before return [filtered_table_cols_dropped], metadata. If it is filtered_table_cols_dropped.num_rows == 0, then do a `return [], metadata`. 

### Anything else

I will assign this bug to myself and fix it along with #1044 . 

### OS

Other

### Python

Other

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

klwuibm:
@touma-I Please assign this bug to me, as part of the pull request to be created for #1044 .

agoyal26:
PR merged

Issue #1044, opened 2025-02-11
[Feature] Filter both the parquet and arrow files and update the metadata simultaneously


Hajar-Emami:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

Filter the corresponding arrow files simultaneously in the parquet-level filtering pipeline, so the tokenization step won’t be necessary after filtering.
It also streamlines the preparation of the ablation data by eliminating the need to re-tokenize, if the data is already tokenized. 

How:
Filter both the parquet and arrow files and update the metadata simultaneously. Please find the link to the code that can filter both the parquet and arrow files and update the metadata simultaneously:

[https://github.ibm.com/ai-models-data/data-prep-kit-inner/blob/hajar_readability_sco[…]forms/language/Readabilty%20Scores/Parquet_Arrow_Filtering.py](https://github.ibm.com/ai-models-data/data-prep-kit-inner/blob/hajar_readability_scores/transforms/language/Readabilty%20Scores/Parquet_Arrow_Filtering.py)](https://github.ibm.com/ai-models-data/data-prep-kit-inner/blob/hajar_readability_scores/transforms/language/Readabilty%20Scores/Parquet_Arrow_Filtering.py)
I include only the filtering parts (the original code also includes the annotation).
Main functions are:
write_to_arrow_doc_files writing the arrow files and meta files
filter_lowQual_Rscore_pq filters from parquet files
filter_Rscore_arw: filters from arrow files

Please let me know if you have any questions. The main part of the code is straightforward. I’d be happy to schedule a call to explain the details.

cc: @shahrokhDaijavad @touma-I 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

shahrokhDaijavad:
Thanks, @Hajar-Emami.

@swith005 Can you please help with this? @Hajar-Emami has already created the code for doing this internally, but an open  DPK transform for this will be a great contribution to both GneissWeb and DPK. Please discuss this with @Hajar-Emami. Thank you.

klwuibm:
[...]

PR #1147, opened 2025-03-19
Filter parquet arrow - [DRAFT] adding support to also filter arrow and meta files

klwuibm:
This pull request addresses two issues: #1044 and #1139
#1044 asks for filtering the .arrow and meta files, including .docs and .docs.ids files, when they are available.  This is to save the extra tokenization pipeline if the parquet folder to be filtered has been tokenized. This pull request added the necessary code before the return of the filtered table by the `FilterTransform.transform()`.
#1139 To address this bug, additional check for the size of the filtered table before returning. If the size is zero, return [], metadata. This way no empty parquet output file will be written to storage.

#1044 and #1139


klwuibm:
CC: @touma-I @shahrokhDaijavad @santoshborse @cmadam @Hajar-Emami

transforms/universal/filter/dpk_filter/transform.py
```diff
@@ -179,6 +310,28 @@ def add_input_params(self, parser: argparse.ArgumentParser) -> None:
             choices=["AND", "OR"],
             help="logical operator (AND or OR) that joins filter criteria",
         )
+        parser.add_argument(
+            f"--{filter_input_arrow_folder_cli_param}",
+            type=str,
+            required=False,
+            default=None,
+            help={f"the input path to the .arrow files"}
```
touma-I:
```suggestion
            help="the input path to the .arrow files"
```

transforms/universal/filter/dpk_filter/transform.py
```diff
@@ -179,6 +310,28 @@ def add_input_params(self, parser: argparse.ArgumentParser) -> None:
...
+        )
+        parser.add_argument(
+            f"--{filter_output_arrow_folder_cli_param}",
+            type=str,
+            required=False,
+            default=None,
+            help={f"the output path to the .arrow files"}
```
touma-I:
```suggestion
            help="the output path to the .arrow files"
```

transforms/universal/filter/dpk_filter/transform.py
```diff
@@ -179,6 +310,28 @@ def add_input_params(self, parser: argparse.ArgumentParser) -> None:
...
+        )
+        parser.add_argument(
+            f"--{filter_doc_id_column_name_cli_param}",
+            type=str,
+            required=False,
+            default="id",
+            help={f"the unique doc_id column name"}
```
touma-I:
```suggestion
            help="the unique doc_id column name"
```

touma-I (changes requested):
@klwuibm If you were to accept my changes, I think this will fix the issue with:
python -m dpk_filter.runtime --help

that was failing CI/CD

touma-I:
@revit13 Could you please provide guidance on the error we are seeing with KFPv2 ? @klwuibm code introduces new variable of type string that have default values of None. I seem to have seen that before? Anyway, do you have a way to confirm if this is indeed the case and how to fix it? thanks a lot

touma-I (approved):
LGTM- Thank you for the quick turnaround

shahrokhDaijavad:
@klwuibm I hope you don't mind that I pushed some small changes to the comment cells of the two notebooks. I had tested running the notebooks without any problems, but the parameter list was missing (in the python case) and was wrong (in the ray case). So, I just added links to the README file. Otherwise, everything looked good. Thank you.

shahrokhDaijavad (approved):
LGTM.

PR #1149, opened 2025-03-20
remove duplicate declaration for runtime entry points

matouma:

Cleanup Makefile template as it repeats the same entry points


touma-I (approved):
LGTM

PR #1150, opened 2025-03-20
new transform to remove commented code

ykalathiya:
This transform help to detect commented code from various known programming languages.

84


Param-S:
@ykalathiya What is the source of this model https://github.com/data-prep-kit/data-prep-kit/blob/ec6797f9ca72f7a3b4b4ac5919e0bc22c8302672/transforms/code/comment_cleanser/model/clf_tfid_char.joblib  you took it from huggingface or other source?

ykalathiya:
I trained it on the local system using open source code and text data. It is multinomial naive bayes with TF-IDF vector in pipeline.

scripts/check-workflows.sh
```diff
@@ -17,7 +17,7 @@ if [ ! -d transforms ]; then
     echo Please run this script from the top of the repository
     exit 1
 fi
-KFP_BLACK_LIST="doc_chunk pdf2parquet pii_redactor text_encoder license_select repo_level_ordering header_cleanser fdedup hap rep_removal gneissweb_classification"
+KFP_BLACK_LIST="doc_chunk pdf2parquet pii_redactor text_encoder license_select repo_level_ordering comment_cleanser header_cleanser fdedup hap rep_removal gneissweb_classification"
```
Param-S: @shivdeep-singh-ibm @matouma This is a new transform. I don't have understanding of this black list. when would we need to a transform to black list?

Param-S (commented):
@ykalathiya can you update this new transform as per new simplified structure.  For example:  proglang_select transform is updated with new code structure here: https://github.com/data-prep-kit/data-prep-kit/tree/dev/transforms/code/proglang_select, use this as reference

PR #1159, opened 2025-03-24
Restore workflow with pull_request_target

matouma:

Restore workflow with pull_request_target


roytman (approved):
LGTM

Issue #1162, opened 2025-03-25
[Bug] Filter invokes duckdb in non thread-safe manner


revit13:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

transforms/filter

### What happened + What you expected to happen

The current filter implementation invokes DuckDB in a manner that is not thread-safe, which could lead to deadlocks when the transform is called from multiple threads. 
https://github.com/data-prep-kit/data-prep-kit/blob/98269a8a064b6cb3d6aa953c649fadc138770d7e/transforms/universal/filter/dpk_filter/transform.py#L97

### Reproduction script

Execute the following script to show deadlock:

```bash
import duckdb
import threading
import pyarrow as pa

data= pa.Table.from_pydict({
    'id': [1, 2, 3, 4],
    'value': [10, 20, 30, 40]
})

def op(thread_id):
    table = duckdb.execute("SELECT id, value FROM data WHERE value > 20").arrow()
    print(table)

if __name__ == "__main__":

    threads = []
    for i in range(5):
        thread = threading.Thread(target=op, args=(i,))
        threads.append(thread)
        thread.start()

    # Wait for all threads to complete
    for thread in threads:
        thread.join()
```

### Anything else

The issue arises because multiple threads share the same DuckDB connection (see the [default Connection](https://duckdb.org/docs/stable/clients/python/dbapi.html#in-memory-connection)). When executing `duckdb.execute(criterion_sql).arrow()`, a temporary table is created within the connection as mentioned [here](https://github.com/duckdb/duckdb/issues/11018#issuecomment-1981207989).

A solution would be to create a connection per- thread as describe in https://duckdb.org/docs/stable/guides/python/multiple_threads.html

### OS

Ubuntu

### Python

3.11

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

swith005:
added in PR #1186

PR #1168, opened 2025-03-26
Filter: Access duckdb in thread-safe manner.

revit13:

The current filter implementation invokes DuckDB in a manner that is not thread-safe, which could lead to deadlocks when the transform is called from multiple threads.
Solution is based on https://duckdb.org/docs/stable/guides/python/multiple_threads.html

#1162


revit13:
Opened PR 1186 instead

Issue #908, opened 2025-01-02
[Feature] Html2ParquetTransform support output_format_value json


1337stn:
### Search before asking

- [X] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Transforms/Other

### Feature

Running through "RAG with Data Prep Kit" focusing on Step-2:

https://github.com/IBM/data-prep-kit/blob/6a06d8763ede388a93e956635af530df3494a9c8/examples/notebooks/rag/rag_1A_dpk_process_python.ipynb

```
Step-2: Process Input Documents (RAG stage 1, 2 & 3)

This code uses DPK to

    Extract text from PDFs (RAG stage-1)
    Performs de-dupes (RAG stage-1)
    split the documents into chunks (RAG stage-2)
    vectorize the chunks (RAG stage-3)
```

In **Extract text from PDFs (RAG stage-1)** when calling `pdf2parquet_transform_python` there are three options for output `pdf2parquet_contents_types: markdown, text, json`

https://github.com/IBM/data-prep-kit/blob/6a06d8763ede388a93e956635af530df3494a9c8/transforms/language/pdf2parquet/dpk_pdf2parquet/transform.py#L67

```
class pdf2parquet_contents_types(str, enum.Enum):
    MARKDOWN = "text/markdown"
    TEXT = "text/plain"
    JSON = "application/json"
```

The next step **split the documents into chunks (RAG stage-2)** when calling `doc_chunk` there are three options for `chunking_type: dl_json, li_markdown, li_token_text` with default = `dl_json`

https://github.com/IBM/data-prep-kit/blob/6a06d8763ede388a93e956635af530df3494a9c8/transforms/language/doc_chunk/README.md?plain=1#L66

Thus we see the example code use **Stage-1 output = json** and **Stage-2 type = json**.

When attempting to change **Stage-1** to  **Extract text from HTML** using `html2parquet_transform_python` there are two options for `html2parquet_output_format: markdown, txt`

https://github.com/IBM/data-prep-kit/blob/6a06d8763ede388a93e956635af530df3494a9c8/transforms/language/html2parquet/dpk_html2parquet/transform.py#L168

```
class html2parquet_output_format(str, enum.Enum):
    MARKDOWN = "markdown"
    TEXT = "txt"
```

However `html2parquet_transform_python` reports to use Trafilatura where Trafilatura also supports JSON output:

[...]

PR #1187, opened 2025-04-06
[Feature] Html2ParquetTransform support output_format_value json #908

1337stn:

dpk_html2parquet.transform_python could allow html2parquet_output_format: json which would pass-through to Trafilatura which already supports JSON. This would allow the flow of Stage-2 and beyond in the RAG example(s) to be maintained since they default to JSON.

Adding JSON support in html2parquet_output_format to align with pdf2parquet_output_format options along with the underlying Trafilatura supported options.

Supporting JSON as an additional output format for the html2parquet transform and making it consistent with the pdf2parquet transform output formats, since Trafilatura already supports this.

None

### DIFF
--- a/transforms/language/html2parquet/README.md
+++ b/transforms/language/html2parquet/README.md
@@ -51,7 +51,7 @@ The table below provides the parameters that users can adjust to control the beh
 
 | Parameter         | Default    | Description                                                                 |
 |-------------------|------------|-----------------------------------------------------------------------------|
-| `output_format`    | `markdown` | Specifies the format of the extracted content. Options: `markdown`, `txt`.  |
+| `output_format`    | `markdown` | Specifies the format of the extracted content. Options: `markdown`, `txt`, `json`.  |
 | `favor_precision`  | `True`     | Prefers less content but more accurate extraction. Options: `True`, `False`. |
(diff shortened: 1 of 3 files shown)


shahrokhDaijavad:
Hi, @1337stn. Thank you for your contribution. Now that you have added the JSON output option, can I ask you one (actually two!) favor(s) for the sake of the completeness of your work, before we merge it?
1) Please modify the README file, to say that in addition to the default markdown and the second txt output options, there is now the 3rd option of JSON as output. 
2) To show that this option works, please add a cell to the notebook file (https://github.com/data-prep-kit/data-prep-kit/blob/dev/transforms/language/html2parquet/html2parquet.ipynb) that exercises this output json format, after exercising the default markdown. It would be fantastic if you could also add another cell that would exercise the txt option too (the original developers should have done this!). 
Thank you again.

1337stn:
@shahrokhDaijavad I have completed the requested changes.  Sorry for delay, but fell down a rabbit hole at work.

1) README file now includes json as an option
2) notebook has cells added to demo txt and json output options
3) notebook had typo plus it seemed to require lxml_html_clean now as my system complained without it

PR #1209, opened 2025-04-16
Fix pip intall typos in Windows install instructions

darshanmalagimani:

These changes fix a typo in the Windows installation instructions for the data-prep-kit. Specifically, it corrects the misspelling of the pip install command (which was incorrectly written as pip intall) to ensure that new users can correctly follow the installation steps without confusion.


touma-I (approved):
LGTM. Thank you @darshanmalagimani

PR #1211, opened 2025-04-17
GneissWebClassifiation: Do not require credential on Command line

matouma:

Allow users to pass credentials as an environment variable and not always as a command line parameter.

### DIFF
--- a/transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
+++ b/transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ################################################################################
-
+import os
 from argparse import ArgumentParser, Namespace
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, config: dict[str, Any]):
         # of ClassificationTransformConfiguration class
         super().__init__(config)
         
-        self.model_credential = config.get(model_credential_cli_param)
+        self.model_credential = config.get(model_credential_cli_param, os.environ.get('HF_READ_ACCESS_TOKEN', "PUT YOUR OWN HUGGINGFACE CREDENTIAL"))
```
swith005: Why is "PUT YOUR OWN HUGGINGFACE CREDENTIAL" an option here? is the workflow not expecting `HF_READ_ACCESS_TOKEN`?

swith005 (commented):
looks good, but left comment on `transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py` regarding setting mode_credential variable

transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
touma-I: @swith005 yes, it is expecting something and it will fail if it does not get a valid token. Originally, this string was meant for developers  who may come up with other alternatives. what do you suggest? We can let it default to None!?!?

swith005: Given that it will fail anyway if not set, I would just let it default to none and have it fail at that stage, or a check immediately after assignment to fail immediately.

touma-I:
```suggestion
        self.model_credential = config.get(model_credential_cli_param, os.environ.get('HF_READ_ACCESS_TOKEN', None))
```

swith005 (approved):
lgtm!

touma-I (approved):

Issue #1228, opened 2025-04-24
[Feature] Refactor the repo_level_order transform to follow the new transform style


ian-cho:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Other

### Feature

Refactor the codebase for repo_level_order using the new format by following [Data Prep Kit- Developer Tutorial](https://github.com/data-prep-kit/data-prep-kit/blob/dev/doc/quick-start/contribute-your-own-transform.md)

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

agoyal26:
ok to close this? as PR is merged

ian-cho:
Hi @agoyal26 I think this can be closed since the PR was merged. Thanks (@touma-I )

PR #1229, opened 2025-04-24
Refactor repo_level_order using new transform style

ian-cho:
The current repo_order_level transform uses old style. Refactor it into new style.

#1228


ian-cho:
Hi @touma-I @shahrokhDaijavad I refactored the old codebase into new one and have several questions that need your answer (in the upcoming meeting): 

1. The previous codebase was named repo_level_ordering, but many subdirectories and scripts use the name repo_level_order, leading to inconsistency. To unify naming, I would like to standardize everything to repo_level_order. I’ve already updated some scripts and YAML files, but there might still be places I’ve missed. Could you help check if anything else needs to be updated?
2. The original codebase did not include a proper Python runtime setup, which makes notebook creation different. Could you help me in setting that up correctly?
3. The output folder was not generated as expected. If you know what might be causing this issue, I would appreciate your guidance. 

Thank you!
(cc @issei-ibm)

shahrokhDaijavad:
Hi, @ian-cho I think unifying with the name repo_level_order is a good idea. 

For questions 2 and 3, let us ask @shivdeep-singh-ibm for help, as a first resource. He either developed this transform himself or knows who did it.

shivdeep-singh-ibm:
@ian-cho @shahrokhDaijavad 

1) As of point 1) raised above, we can use a unified name.
2) Yes, this version has only ray, since it makes use of runtime and additional ray actor as a store. 
     you can skip adding python launcher, I will add it.
3) I will check what is causing the problem. Thanks
    you need to add RepoLeveLOrderRuntime in RayTransformConfiguration. like https://github.com/data-prep-kit/data-prep-kit/pull/1229/files#r2063065463

transforms/code/repo_level_order/dpk_repo_level_order/ray/runtime.py
```diff
@@ -0,0 +1,55 @@
...
+from data_processing_ray.runtime.ray.runtime_configuration import (
+    RayTransformRuntimeConfiguration,
+)
+from data_processing.utils import ParamsUtils
+import sys
+
+class RepoLevelOrderRayTransformConfiguration(RayTransformRuntimeConfiguration):
+    def __init__(self):
+        super().__init__(
+            transform_config=RepoLevelOrderTransformConfiguration()
```
shivdeep-singh-ibm: *You need to add RepoLevelOrderRuntime in super.init() like:*

from dpk_repo_level_order.transform import RepoLevelOrderTransformConfiguration, RepoLevelOrderRuntime

class RepoLevelOrderRayTransformConfiguration(RayTransformRuntimeConfiguration):
     def __init__(self):
         super().__init__(
      transform_config=RepoLevelOrderTransformConfiguration(),
            runtime_class=RepoLevelOrderRuntime,
         )

touma-I (approved):
LGTM

transforms/code/repo_level_order/dpk_repo_level_order/ray/runtime.py
touma-I: Re-use existing implementation of the same class.

shivdeep-singh-ibm (approved):
lgtm

shahrokhDaijavad:
@touma-I I removed the notebook, because it was both incomplete and broken. If we want to add it, since it was a Ray runtime notebook, the parameters need to be carefully added.

PR #1231, opened 2025-04-25
remove print statement and replace with logger.info

matouma:
minor cleanup of print statements


touma-I (approved):
LGTM

Issue #1042, opened 2025-02-11
[Bug] Pdf2parquet inbuilt ocr error


ShiroYasha18:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

Documentation

### What happened + What you expected to happen

I was trying to do pdf2parquet on google colab using the inbuilt ocr- easy ocr to be specific. Apparently it is not working as in throwing the error when the parameters are are set according to the documentation

### Reproduction script

![Image](https://github.com/user-attachments/assets/9ca64b7d-1728-47a4-99ee-846a21d114c3)

![Image](https://github.com/user-attachments/assets/b1fb2122-e68a-4559-a569-4caae762880c)

I have attached the images of the errors and the documentation which was about the parameters of the pdf2parquet

### Anything else

If I remove the do_ocr and the ocr_engine parameter it will work just fine soo rest of the things are working fine and all dependencies are installed properly

### OS

Other

### Python

3.10.x

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

agoyal26:
@dolfim-ibm Please see above. Is this related to Docling functionality?

dolfim-ibm:
I think you have to prepend the arguments with `pdf2parquet`. Whether it is needed or not, it depends if the transform is used via the DPK launcher or standalone (like all other transforms).

Example at https://github.com/IBM/data-prep-kit/blob/dev/transforms/language/pdf2parquet/dpk_pdf2parquet/local_python.py#L44

ShiroYasha18:
UPDATE:
ok prepending `pdf2parquet` worked for `do_ocr` param but I cannot still change the ocr_engine from easyocr to tesseract

however it works fine if the params are : 
` "pdf2parquet_do_ocr": True,`
    `"pdf2parquet_ocr_engine": "tesseract_cli",`



the ocr engine is working fine for easyocr and tesseract_cli

dolfim-ibm:
Do you get an error message or something else?

ShiroYasha18:
Exception creating transform  tesserocr is not correctly installed. Please install it via `pip install tesserocr` to use this OCR engine. Note that tesserocr might have to be manually compiled for working withyour Tesseract installation. The Docling documentation provides examples for it. Alternatively, Docling has support for other OCR engines. See the documentation.
Traceback (most recent call last):

[...]

PR #1235, opened 2025-04-26
Fixed docling2parquet documentation + bug fix in code2parquet + pdf2parquet optimization

ShiroYasha18:
 I checked inbuilt ocr feature for both local as well as google colab. I found out that for colab the ocr engine and the ocr features wont work unless ``` pdf2parquet_``` is prepended to ```do_ocr``` and ```ocr_engine```. After prepending it works in google colab. Locally it works both with and without prepending I believe , hence I think it was not outlined before the raised issue.

Another commit is a small typo error in the code2parquet where the this fix ensures that the snapshot configuration parameter is correctly utilized, allowing users to specify different values for both the domain and the snapshot. As a result, the metadata in the output Parquet files accurately reflects the intended configuration. Without this fix, the snapshot value would always default to the domain value, which could lead to confusion during later analysis of the processed data
 
 Finally the ```transform.py``` is to address the optimization of pdf2parquet(now docling2parquet). Even when I was working with the ```pdf2parquet``` I thought that it is kind of slow which pretty much affects the user experience as pointed out by @sujee  in the issue #573 . The main culprit over here are 2 things :
 
 1. PDF backend : in dpk currently we are using``` DLPARSE_V2``` which is fab for complex layouts but has a significant sacrifice on the speed. I have switched that to PyPDFium which will increase the speed in parsing in theory 4x to 5x   faster. Yes its not as powerful as compared to ```DLPARSE_V2``` but sure not all pdfs have exremely complex layouts by default . If needed we can possibly router the pdfs with complex things to ```DLPARSE_V2``` but normally lets use PyPDFium due to its fast nature and speed sure should be priority. As much as I have used PyPDFium I think its also is a pretty solid pdf backend and speed is insanely fast. 
 
 2. Default OCR: Now I saw that the pdf2parquet used the OCR of EasyOCR in the code by default. Well it sure would be another big reason for such delay . Its kind of scary to turn off the default ocr but there is was a dynamic usage which was in action as much as I understood using the bitmap area threshold. however again that was around 0.05 . I changed that to 0.10 that means the threshold for using OCR directly would be that it should cover atleast 10% of the page. This will leave out the ocr on super small images (images which are less than 10% of the page) but again if u see most of the pdfs processed has a proper diagram or image for ocr it wont be that small I believe so we save a bit more time on the uncessary ocr reduction 
 
 and the last commit is in ```flush_binary``` 
- Fixed critical issue in the``` flush_binary``` method
- Ensures proper buffer handling during multiple document processing
- Prevents potential errors in distributed processing environments
  This fix ensures that if flush_binary is called multiple times (which can happen in distributed processing environments or during error recovery), the second call won't raise a TypeError when trying to check len(self.buffer) on a None value.
 
 EDIT 29/04/2025 :
here is the link to the video of a screen recording which I made after docling2parquet was merged by @shahrokhDaijavad  and even locally the do_ocr was to be prepended to work . 
https://drive.google.com/file/d/1FMtiwgUXkCVuNsTHKF0oHgupSiPyLWnz/view?usp=drive_link

Please have a look !

Referring to issue #1042 for the ocr issue
Referring to issue #573  for th transform optimization

### DIFF
--- a/data-connector-lib/pyproject.toml
+++ b/data-connector-lib/pyproject.toml
@@ -32,7 +32,7 @@ Issues = "https://github.com/data-prep-kit/data-prep-kit/issues"
 Documentation = "https://data-prep-kit.github.io/data-prep-kit/"
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0,<=8.3.0"]
 build-backend = "setuptools.build_meta"
 
 [project.optional-dependencies]
--- a/data-processing-lib/pyproject.toml
+++ b/data-processing-lib/pyproject.toml
@@ -18,7 +18,7 @@ Issues = "https://github.com/data-prep-kit/data-prep-kit/issues"
 Documentation = "https://data-prep-kit.github.io/data-prep-kit/doc"
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0,<=8.3.0"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.dynamic.dependencies]
--- a/kfp/kfp_support_lib/kfp_v1_workflow_support/pyproject.toml
+++ b/kfp/kfp_support_lib/kfp_v1_workflow_support/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
 ]
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0,<=8.3.0"]
 build-backend = "setuptools.build_meta"
 
 [project.optional-dependencies]
--- a/kfp/kfp_support_lib/kfp_v2_workflow_support/pyproject.toml
+++ b/kfp/kfp_support_lib/kfp_v2_workflow_support/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
 ]
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0,<=8.3.0"]
 build-backend = "setuptools.build_meta"
 
 [project.optional-dependencies]
--- a/kfp/kfp_support_lib/shared_workflow_support/pyproject.toml
+++ b/kfp/kfp_support_lib/shared_workflow_support/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
 ]
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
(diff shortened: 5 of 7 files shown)


agoyal26:
@dolfim-ibm Please take a look. Thanks

shahrokhDaijavad:
Thanks, @ShiroYasha18. Yes, the video clearly shows that the README change is needed. As for the default OCR backend and the default threshold values that you propose to change for performance reasons, I need @dolfim-ibm to comment.

transforms/language/docling2parquet/README.md
```diff
@@ -66,8 +66,8 @@ The transform can be initialized with the following parameters.
 | `artifacts_path`             | <unset> | Path where to Docling models artifacts are located, if unset they will be downloaded and fetched from the [HF_HUB_CACHE](https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache) folder. |
 | `contents_type`         | `text/markdown`        | The output type for the `contents` column. Valid types are `text/markdown`, `text/plain` and `application/json`. |
 | `do_table_structure`         | `True`        | If true, detected tables will be processed with the table structure model. |
-| `do_ocr`                     | `True`        | If true, optical character recognition (OCR) will be used to read the content of bitmap parts of the document. |
-| `ocr_engine`                 | `easyocr`     | The OCR engine to use. Valid values are `easyocr`, `tesseract`, `tesseract_cli`. |
+| `docling2parquet_do_ocr`                     | `True`        | If true, optical character recognition (OCR) will be used to read the content of bitmap parts of the document. |
```
dolfim-ibm: The fact of prepending `docling2parquet_` to the parameters depend on how the transform is used. In some use cases you don't need it, in others you need it.

I don't think we want to update the description of the parameters here. (in case, it must be for *all* of them)
Just a few lines below, it is described that in the CLI usage the user must prepend the parameters with `docling2parquet_`.

We should describe better all the usages when it is needed since it is not only the CLI.
@shahrokhDaijavad @touma-I this may apply to many other transforms.

transforms/language/docling2parquet/dpk_docling2parquet/transform.py
```diff
@@ -95,9 +95,9 @@ def __str__(self):
 docling2parquet_contents_type_default = docling2parquet_contents_types.MARKDOWN
 docling2parquet_do_table_structure_default = True
 docling2parquet_do_ocr_default = True
-docling2parquet_bitmap_area_threshold_default = 0.05
+docling2parquet_bitmap_area_threshold_default = 0.10
```
dolfim-ibm: this is ok with me.

transforms/language/docling2parquet/dpk_docling2parquet/transform.py
```diff
@@ -95,9 +95,9 @@ def __str__(self):
...
 docling2parquet_ocr_engine_default = docling2parquet_ocr_engine.EASYOCR
-docling2parquet_pdf_backend_default = docling2parquet_pdf_backend.DLPARSE_V2
+docling2parquet_pdf_backend_default = docling2parquet_pdf_backend.PYPDFIUM2
```
dolfim-ibm: Here I'm not so sure. We also have the DLPARSE_V4 now, so I would like to first see a comparison of the latest Docling version and parser. You might want to consider an issue on Docling about this, specially if you can provide some example documents.

I also think it is better to keep the same default values as in Docling.

dolfim-ibm (changes requested):

transforms/language/docling2parquet/dpk_docling2parquet/transform.py
ShiroYasha18: Oh I didnt know that we have DLPARSE_V4 too? I mean pypdfium still would be faster but DLPARSE_V4 would be a worthy competitor especially with the accuracy around tables etc. So what are the next steps ? I need to open an issue at the docling repo ?

PR #1241, opened 2025-05-02
Malware refactored

matouma:

Refactored malware transform


shahrokhDaijavad:
@touma-I I cleaned up the README file, however, since there is no notebook example, it would be good to create one makefile target like `run-cli-sample` that shows the user how to run the transform from the command line. The README file would then add this example.

shivdeep-singh-ibm (approved):
lgtm

PR #1255, opened 2025-05-08
candidate release testing: 1.1.1.dev1

matouma:

Allow incremental merges with feature/pre-release to accelerate testing


touma-I (approved):
Approved for testing on feature branch only

touma-I:
Meging with feature branch as it will receive incremental updates until all problems are fixed

PR #1268, opened 2025-05-09
Re-write simplified release process

touma-I:

New documentation of release process


shahrokhDaijavad:
@touma-I A couple of differences between the simplified version that you have written and the existing document:

1) You are not using `scripts/release-branch.sh`, which has simplified things! 
2) Obviously, you don't use these notes to publish to pypi, so I guess the part about publishing wheels is not needed here.

touma-I:
> @touma-I A couple of differences between the simplified version that you have written and the existing document:
> 
> 1. You are not using `scripts/release-branch.sh`, which has simplified things!
> 2. Obviously, you don't use these notes to publish to pypi, so I guess the part about publishing wheels is not needed here.

Yes @shahrokhDaijavad . We should have a separate document for publishing wheels and docker images. The release process for me is just a snapshot in time. The wheels and docker images is more of an ongoing things that we do on release but also for non-official releases.

shahrokhDaijavad (approved):
LGTM

RELEASE.md
```diff
@@ -2,112 +2,55 @@
...
-git checkout pending-releases/vX.Y.Z 
-make show-version
+git checkout dev
+git pull
+git checkout -b pending-release/x.x.x     ## Replace x.x.x with the proper release tag
 ```
-This will print for example, 1.2.3. 
-
-To run the script from the top of the repo:
+From the main folder, edit the `.make.versions` and remove the suffix (e.g., dev0, dev1, ... ) for all the components being released.
```
swith005: > From the main folder, edit the .make.versions and remove the suffix (e.g., dev0, dev1, ... ) for all the components being released. 
would be helpful to add a link to this file

RELEASE.md
```diff
@@ -2,112 +2,55 @@
...
 
-```shell
-scripts/release-branch.sh
+```
+make set-versions
 ```
 
-After running the script, you should
+Edit the release notes `releases-notes.md`.
```
swith005: > Edit the release notes releases-notes.md.
What to edit? The following line showing `git add .` also doesn’t connect

swith005 (changes requested):
left comments for clarification

RELEASE.md
touma-I: Added some additional text to explain what to add

swith005 (approved):
lgtm - thanks for resolving comments

PR #1275, opened 2025-05-13
Consistent names for runtime files

touma-I:

Provide consistent naming of files for all transforms runtime;

rename transform_python and ray/transform.py to runtime.py and ray.runtime.py


transforms/language/html2parquet/dpk_html2parquet/ray/transform.py
swith005, line 1: confirming this is replaced by ray/runtime.py?

swith005 (commented):
Since this PR isn't in draft, want to confirm whether this is the finalized list of transform updates for this PR?

touma-I:
> Since this PR isn't in draft, want to confirm whether this is the finalized list of transform updates for this PR?

Hi @swith005 I think we should keep it in draft mode for a while. Releasing this now can potentially break many of the notebooks. Let's discuss on our call today on when we should release this.

shahrokhDaijavad (changes requested):
Postpone merging until most notebook recipes have been upgraded in using the consistent runtimes.

PR #1276, opened 2025-05-13
Introduction of recipe folder and its first notebook

shahrokhDaijavad:
We are introducing a recipe folder with sample notebooks, in which notebook output cells have been removed after testing. 

We still need to get rid of the file_utils.py file.


shahrokhDaijavad:
@touma-I I took out the Google Colab stuff, since we don't need the complication of adding an extra conda environment for this recipe, which is intended to show "sequencing" of DPK transforms. I created issue #1277 for getting rid of `files_util.py`. In this PR, I have already taken out the part of this file that did `download_file`, since we don't need it anymore.

touma-I:
@shahrokhDaijavad let's review when you are back. thanks.

touma-I (approved):

Issue #1260, opened 2025-05-08
A feature that would allow classifier transforms to download models from either S3, COS or HF


shahrokhDaijavad:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

transdforms/Other

### Feature

At the moment, classifier transform either download small test models locally of from HuggingFace. We need to extend this to allow downloading from COS or S3. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

swith005:
thanks for creating @shahrokhDaijavad ; 'll work on this.

swith005:
Merged in PR #1316

PR #1281, opened 2025-05-16
Download util rebase

swith005:
support for transforms to download utilized models/tokenizers from cos, huggingface, or locally, rather than each transform having their own method.

#1260


touma-I:
Replace with https://github.com/data-prep-kit/data-prep-kit/pull/1282

PR #1282, opened 2025-05-16
Now utility/service for retrieving models from HF, S3 or local

touma-I:

Work contributions from Shalisha and tested with Land_id

Squashed by MT


swith005 (approved):
lgtm! thanks for adding my PR to comply with verified signature

PR #1285, opened 2025-05-19
removed runtime_code_location parameter from lang_id and execution co…

swith005:
removed runtime configuration as a parameter from runtime execution config to set dynamically from docker environment. starting with lang_id for latest changes (corresponds with changes in data-ops)


data-processing-lib/python/src/data_processing/runtime/execution_configuration.py
```diff
@@ -48,18 +48,6 @@ def add_input_params(self, parser: argparse.ArgumentParser) -> None:
         parser.add_argument(f"--{runtime_cli_prefix}pipeline_id", type=str, default="pipeline_id", help="pipeline id")
         parser.add_argument(f"--{runtime_cli_prefix}job_id", type=str, default="job_id", help="job id")
 
-        help_example_dict = {
```
touma-I: @swith005 I think removing this will cause all the transforms that are passing this as a parameter to fail. I suggest first modify all the _wf.py and all the UT that are passing this parameter to remove and then remove these lines from her. Unless you want to do it in the same PR

touma-I (changes requested):

transforms/language/lang_id/Dockerfile.python
touma-I, line 1: Can you please make the same changes to transforms/Dockerfile.python.template

transforms/language/lang_id/Dockerfile.ray
touma-I, line 1: @swith005 Can you add the same 4 lines to the transforms/Dockerfile.ray.template ?

touma-I (changes requested):

.make.defaults
touma-I, line 1: @swith005 We need to add the following to the .defaults.test-src:: (line 445)
         export PYTHONPATH=../src:../: ;  \
	export BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') ; \
	export TRANSFORM_PATH=$(shell git rev-parse --show-prefix) ;  \
	export GIT_URL=$(shell git config --get remote.upstream.url) ; \
	export GIT_COMMIT=$(shell git log -1 --format=%h) ; \
	cd test; $(PYTEST) .

touma-I (changes requested):
a few comments/request for changes to address failed CI/CD tests

data-processing-lib/python/src/data_processing/runtime/execution_configuration.py
swith005: you might be right; another option is to update the configuration codes to use `args, _ = parser.parse_known_args()` instead of `args = parser.parse_args()`, which should be the preferred method anyway.

touma-I (approved):

PR #1306, opened 2025-05-30
Adding example to run the gopher_repetition_annotator with all cli params and adding main in ray runtime

santoshborse:


swith005 (approved):
lgtm; thanks

shahrokhDaijavad:
@santoshborse Isn't it strange that the kfp-v1 test is now failing, even though it was passing before, and the changes you made have nothing to do with this test?

swith005:
> @santoshborse Isn't it strange that the kfp-v1 test is now failing, even though it was passing before, and the changes you made have nothing to do with this test?

@shahrokhDaijavad I caught this too, which is why I wasn't against approving his updates... but also curious why this was triggered and failing...Not enough information in the logs unfortunately other than: 
` WARNING - Failure to get pipeline. Number of pipelines with name gopher_repetition_annotator_wf is 0`
 Maroun plans to walk me through setting up kind cluster for running these test-kfp runs when he returns, so don't quite have an idea myself as of yet.

santoshborse:
@shahrokhDaijavad yes it is strange, not sure what to do

shahrokhDaijavad (changes requested):
@santoshborse One thing that I had not noticed when the original PR was merged is that the title of the https://github.com/santoshborse/data-prep-kit/blob/annotator_gopher_repetition_1/transforms/universal/gopher_repetition_annotator/kfp_ray/README.md file should be changed from "Programming Language Select" to "Gopher Repetition Annotator". Please make this change while the KFP failure is being investigated.

touma-I:
> @shahrokhDaijavad yes it is strange, not sure what to do

@santoshborse In your workflow, you are using `gopher_repetition_` prefix while your code is now setup for `gra_` prefix. You need to update the list of parameters in the _wf.py to use gra_

touma-I (changes requested):
Update the prefix in the _wf.py (should be gra_ not gopher_repetition_ )

transforms/universal/gopher_repetition_annotator/kfp_ray/gopher_repetition_annotator_wf.py
```diff
@@ -48,21 +48,21 @@ def compute_exec_params_func(
...
+    gra_top_2_grams_cname: str,
+    gra_top_3_grams_cname: str,
+    gra_top_4_grams_cname: str,
+    gra_dup_5_grams_cname: str,
+    gra_dup_6_grams_cname: str,
+    gra_dup_7_grams_cname: str,
+    gra_dup_8_grams_cname: str,
+    gra_dup_9_grams_cname: str,
+    gra_dup_10_grams_cname: str,
+    gra_normalize_string: bool,
```
touma-I: @santoshborse I don't see gra_normalized_string in the list of parameters defined by the Configuration class. The test is failing with:
`runtime.py: error: unrecognized arguments: --gra_normalize_string=False
`

touma-I (changes requested):
`gra_normalize_string: bool` is not defined

transforms/universal/gopher_repetition_annotator/kfp_ray/gopher_repetition_annotator_wf.py
santoshborse: removed

shahrokhDaijavad (approved):

touma-I (approved):
LGTM

PR #1316, opened 2025-06-05
updated model_loader util and all transforms that utilize model loadi…

swith005:
…ng to utitlize util

Models currently can only be passed through hugging face url; a new model loader utility was added to allow loading not only from hugging face, but also s3 and locally. consequently, code_quality, hap, gneissweb_classification, and tokenization have been updated to utilize this utility for broader model access.


shahrokhDaijavad (approved):
LGTM. Existing issue with Spark transforms (not relevant to this PR) are ignored.

PR #1321, opened 2025-06-09
Added prefect in-memory workflow orchestration 

swith005:
New directory for creating in-memory workflows utilizing Prefect orchestration library.


prefect_workflow/README.md
```diff
@@ -0,0 +1,210 @@
...
+python orchestrator/abstract_orchestrator.py /path/to/your/flow.yaml
+```
+
+Supports `.yaml`, `.yml`, or `.json`.
+
+---
+
+## 📦 Output
+
+Results (e.g., transformed data, logs) will be stored in the specified `output_folder` defined under `global_config`.
```
revit13, line 209: It would be helpful to include a complete, ready-to-use JSON input that allows running the example out of the box. This way, we can directly test the implementation using input files and observe the resulting output files.

prefect_workflow/orchestrator/abstract_orchestrator.py
```diff
@@ -0,0 +1,193 @@
...
+
+        try:
+            while True:
+                batch_files = da.get_batches_to_process()
+                if len(batch_files) == 0:
+                    break
+
+                # generate prefect code
+                prefect_code = "@flow(timeout_seconds=3600,retries=3, retry_delay_seconds=60, log_prints=True)"
+                prefect_code += "\nasync def dpk_pipeline(orch_obj, name, op_dag, in_memory, data_access, batch_files):"
```
revit13, line 128: it seems the function flow is generated for each batch.  Would it be possible to generate it once and reuse it for each batch instead?

prefect_workflow/requirements.txt
```diff
@@ -0,0 +1,2 @@
+prefect
```
revit13, line 1: it will be safer I think to use a specific version.

prefect_workflow/README.md
```diff
@@ -0,0 +1,210 @@
...
+#### `global_config`
+Defines settings that apply to the entire flow:
+- `data_storage_type`: Currently supports `"local"` and `"s3"` (can be extended).
+- `data_access`: Contains configuration for data IO and processing.
+  - `s3_config`: Input/output folder paths on S3.
+  - `s3_credentials`: Credentials for accessing the S3 bucket.
+  - `batch_size`: Number of files/records per operation batch.
+  - `files_to_use`: List of file extensions to process (e.g., `.parquet`, `.csv`).
+
+#### `sequence`
```
roytman, line 34: I suggest renaming it to `dag`

prefect_workflow/README.md
```diff
@@ -0,0 +1,210 @@
...
+A list of sequential or dependent transforms in the DAG-style workflow.
+
+Each node includes:
+- `name`: Unique name of the node (used as reference).
+- `transform`: The transformation function to apply (e.g., `lang_id`, `doc_quality`).
+- `params`: Dictionary of parameters required by the transform.
```
roytman, line 40: Should it be `config` ?

prefect_workflow/README.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+> 💡 If you get a `SyntaxError: invalid decimal literal`, it may be due to a conflict with the `uuid` package (which is built into Python and **should not be installed** as a separate package). To resolve:
+>
+> ```bash
+> pip uninstall uuid
+> ```
+
+
+### 3. Install Prefect Dependencies
```
roytman, line 185: Install Third-party Dependencies (?)

prefect_workflow/orchestrator/abstract_orchestrator.py
```diff
@@ -0,0 +1,193 @@
...
+                            prefect_code += f"\n\tdata_access_{index + 1} = __execute_step.submit(orch_obj, \"{name}_{i}\", op_dag[{i}],  data_access_{prev_index + 1}.result(), batch_files)"
+                        i = i + 1
+
+                prefect_code += f"\n\tsave_in_memory_to_file(batch_files, data_access_{index + 1}.result(), data_access)"
+
+                self.logger.info("Generated Code:\n")
+                self.logger.info(prefect_code)
+
+                local_env = {}
+                exec(prefect_code, globals(), local_env)
```
roytman, line 161: @swith005 , I think your code is based on our old Prefect version, which created a string-based Prefect code and then executed it. Currently, we use only Python code, which is more robust (IMHO).

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -155,6 +159,31 @@ def _get_files_to_process_internal(self) -> tuple[list[str], dict[str, float], i
             )
         return path_list, profile, retries
 
+    def get_batches_to_process(self, batch_size=None):
```
roytman, line 162: If you plan to call it from different threads, should it be protected?

prefect_workflow/README.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+Supports `.yaml`, `.yml`, or `.json`.
+
+---
+
+## 📦 Output
+
+Results (e.g., transformed data, logs) will be stored in the specified `output_folder` defined under `global_config`.
+
```
roytman, line 210: Maybe it makes sense to explain that the flow can be observed later at the Prefect dashboard, like:
image

swith005: good suggestion; will add that

prefect_workflow/orchestrator/abstract_orchestrator.py
swith005: yes, @revit13 mentioned in the meeting. I also was making changes for this string-based execution, so in agreement as well .

prefect_workflow/README.md
swith005: sure, I will show json in addition to yaml example.

prefect_workflow/orchestrator/abstract_orchestrator.py
swith005: good point; yes, I'm updated for one flow, and many tasks, as opposed to many flows per batch.

prefect_workflow/README.md
swith005: I thought `params` was more easier to understand than `config`, but to align with the actual argument name in the transform, I can change to config.

prefect_workflow/README.md
swith005: I saw the name was changed to sequence in the original prefect code, but if it was changed back to dag, no problem changing as well. However, I think sequence is a more straightforward name.

data-processing-lib/python/src/data_processing/data_access/data_access.py
swith005: I believe that should be the orchestrator's responsibility; not at the data access level; it just produces batches.

swith005:
closing pending changes; utilizing transform chain for sequential transform runs from pr #1329

PR #1323, opened 2025-06-10
Fix imcompatible versions for PySpark

touma-I:

The latest release of PySpark 4.0.0 on May 22, 2025 is not compatible with the current release of Java used by the workflow. Address this issue by using PySpark 3.5.5
Keep issue open as the final solution will be upgraded the workflow to use a newer version of Java
#1324 1324


shahrokhDaijavad (approved):
LGTM. Good job in finding the source of the problem!

PR #1329, opened 2025-06-17
new service chain  for one or more transforms running in a sequence on micro-batch

swith005:
native python method for running transforms in sequence without intermediate output files.


.github/workflows/test-workflow-dpk_transform_chain.yml
```diff
@@ -0,0 +1,53 @@
+#
+# This file was built manually as it does not follow the same structure as other transforms function
+#
+name: Test - @TARGET_TRANSFORM_DIR@
```
touma-I:
```suggestion
name: Test - Workflow chain composition using in-memory data access
```

touma-I:
@swit005 To fix this error https://github.com/data-prep-kit/data-prep-kit/actions/runs/15787085148/job/44505744989?pr=1329
I suggest we modify line 44 of check-workflows.sh 
https://github.com/data-prep-kit/data-prep-kit/blob/c608465053cd89a252e101c255d5568e70b4c942/scripts/check-workflows.sh#L44 as follow:
`	if [ ! -e $workflow ] && [ -e transforms/$category/$transform/Makefile ]; then
`

This way if the folder does not have a Makefile, the we do not require a workflow (since the workflow relies on having a Makefile anyway)

touma-I (changes requested):
One more change to fix Ci/CD error..

swith005:
> One more change to fix Ci/CD error..

@touma-I updated line and test seems to pass; re-requested review

touma-I (approved):

PR #1344, opened 2025-06-27
trigger flush of buffered data before process a new subfolder

matouma:

ttps://github.com/data-prep-kit/data-prep-kit/issues/1343


swith005 (approved):
thanks for walking me through - looks good to me, and flushing flag solution makes sense.

touma-I (approved):

PR #1364, opened 2025-07-09
Fixing the logo in README.md 

Ibrahim2595:
The logo was stretched because both height and width were set to 60%. I've set the width only to be 50%. Height will adjust accordingly.


shahrokhDaijavad (approved):
Thank you, @Ibrahim2595. It looks good. I don't know at what point the logo got stretched like this, because it looked good before. In any case, your change fixes the stretch. @touma-I Please merge.

PR #1371, opened 2025-07-10
get rid of removed_column

matouma:

Get rid of removed_column from parquet
added removed column to metadata json

https://github.com/data-prep-kit/data-prep-kit/issues/1370


klwuibm (approved):
The code changes look good to me.  Two KFP tests still running, but they probably should be independent of the transform code changes.

swith005 (approved):
lgtm and makes more sense as metadata content

Issue #1380, opened 2025-07-14
DPK's Docling2parquet CSV ingestion


ShiroYasha18:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Python Runtime

### What happened + What you expected to happen

I was testing the CSV conversion to parquet using ```docling2parquet```. I realised that csv are not identified in the mime type inspite of docling having a backend code and support for that. I saw a similar kind of  error in one of the issues raised by @shahrokhDaijavad where the problem I think was with html ingestion I think and gets probably solved by updating mime type. I am also updating the mime logic and this :```examples/notebooks/hap/generate_hap_score_csv.ipynb``` in which for csv conversion pandas is being used .

### Reproduction script

 .

### OS

MacOS

### Python

3.10

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

shahrokhDaijavad:
@ShiroYasha18 Despite the fact that I added the mime type fix in #1381 and tested that it works correctly, we will probably not merge #1381, for two reasons:

1) Support for csv to parquet is a couple of lines of code using pandas
```
import pandas as pd
df = pd.read_csv("input.csv")
df.to_parquet("output.parquet")
```
2) We are not saying anything about supporting csv files in the README file for docling2parquet in DPK. 

So, unless docling is doing something special with csv files, this is not a bug.

ShiroYasha18:
helllo @shahrokhDaijavad  sir, Thank you soo much for replying.


> 1. Support for csv to parquet is a couple of lines of code using pandas

Yes, that’s absolutely true. However, while exploring Docling’s code, I noticed that it has a separate backend for CSV ingestion:
https://github.com/docling-project/docling/blob/main/docling/backend/csv_backend.py

From what I understand, this backend provides more structured control compared to plain pandas — especially around metadata handling, MIME typing, and uniform parsing through DoclingDocument. This kind of control and consistency seems aligned with Docling’s goal of supporting multiple formats through a common structure.

That made me curious — if Docling is already supporting CSVs at that level, would DPK follow the same path and treat CSVs as a first-class format, using the same abstraction?

> 2. We are not saying anything about supporting csv files in the README file for docling2parquet in DPK.

[...]

PR #1381, opened 2025-07-14
docling suport for csv files

shahrokhDaijavad:
Addressing issue #1380


shahrokhDaijavad:
@touma-I: When I made the numpy version change to 1.26.4 in the data-processing-lib/python/requirements.txt file, I got the error:

error: data-prep-toolkit 1.0.1.dev0 requires numpy<=1.26.4, but you have numpy 2.2.6 which is incompatible.

So, I left that file as it was: numpy < 2.0.0 and added the same line ( numpy < 2.0.0) to the requirements.txt file for docling2parquet, and now it passes, so if we don't restrict numpy, docling installs 2.2.6, which is problematic. 

We don't need to merge this PR, but it serves as a learning tool for the numpy version requirement. 

cc: @swith005

touma-I:
> @touma-I: When I made the numpy version change to 1.26.4 in the data-processing-lib/python/requirements.txt file, I got the error:
> 
> error: data-prep-toolkit 1.0.1.dev0 requires numpy<=1.26.4, but you have numpy 2.2.6 which is incompatible.
> 
> So, I left that file as it was: numpy < 2.0.0 and added the same line ( numpy < 2.0.0) to the requirements.txt file for docling2parquet, and now it passes, so if we don't restrict numpy, docling installs 2.2.6, which is problematic.
> 
> We don't need to merge this PR, but it serves as a learning tool for the numpy version requirement.
> 
> cc: @swith005

@shahrokhDaijavad Please close this PR and raise an issue. It seems the error yo are running into is NOT related to your code but can be reproduce by anyone using the latest dev. Please do instruct your students NOT to raise a PR unless requested to do so by the maintainers of the repo.

@touma-I Done. Issue #1384 was raised.

PR #1383, opened 2025-07-15
add the ABC class as a base class to all transforms

roytman:
Add the Pythonic mechanism to define abstract classes and methods. 

https://github.com/data-prep-kit/data-prep-kit/issues/1382

### DIFF
--- a/data-processing-lib/python/src/data_processing/data_access/data_access.py
+++ b/data-processing-lib/python/src/data_processing/data_access/data_access.py
@@ -344,7 +344,7 @@ class DataAccess(ABC):
                  Retries are performed on operation failures and are typically due to the resource overload.
 
         """
-    	pass
+        pass
 
     def get_folder_files(
             self, path: str, extensions: list[str] = None, return_data: bool = True
--- a/data-processing-lib/python/src/data_processing/data_access/data_access.py
+++ b/data-processing-lib/python/src/data_processing/data_access/data_access.py
@@ -11,13 +11,14 @@
 # limitations under the License.
 ################################################################################
 
+from abc import ABC, abstractmethod
 import random
 from typing import Any
 import pyarrow as pa
 from data_processing.utils import KB, MB, GB, TransformUtils, get_logger
 
 
-class DataAccess:
+class DataAccess(ABC):
     """
     Base class for data access (interface), defining all the methods
     """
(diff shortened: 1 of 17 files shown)


data-processing-lib/python/src/data_processing/transform/binary_transform.py
```diff
@@ -39,17 +41,17 @@ def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tupl
                 to metadata.  Each element of the return list, is a tuple of the transformed bytes and a string
                 holding the extension to be used when writing out the new bytes.
         """
-        raise NotImplemented()
+        pass
```
swith005, line 44: @roytman why pass instead of raising `NotImplementedError()` to make it more explicit?

roytman: It will raise `TypeError:`
```
a = AbstractBinaryTransform()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class AbstractBinaryTransform without an implementation for abstract method 'transform_binary'
```
I understood that it is a Python convention

swith005: I see; I usually encounter the explicit NotImplementedError, but TypeError from @abstractmethod is fine since the message is pretty descriptive.

swith005 (approved):

touma-I (changes requested):
Thanks @roytman . For consistency, should we also annotate the abstract methods in folder_transform and table_transform ? Otherwise, this is good. Thanks

roytman:
> Thanks @roytman . For consistency, should we also annotate the abstract methods in folder_transform and table_transform ? Otherwise, this is good. Thanks

Yes, good point, I missed their methods.

swith005 (approved):

swith005 (changes requested):
@roytman tests are now failing due to indent levels for `pass` statements in code.

roytman:
@touma-I  @swith005  All checks have passed.

data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -52,19 +53,21 @@ def __init__(
...
         self.logger = get_logger(__name__)
 
+    @abstractmethod
     def get_output_folder(self) -> str:
         """
         Get output folder as a string
         :return: output_folder
         """
-        raise NotImplementedError("Subclasses should implement this!")
+    pass
```
swith005: @roytman the indent level still seems not to line up on this file.

swith005 (changes requested):

data-processing-lib/python/src/data_processing/data_access/data_access.py
touma-I:
```suggestion
        pass
```

touma-I (commented):
@swith005: I think I managed to fix the last indentation.

swith005 (approved):
looks good now - thanks

touma-I (approved):
Thanks Alexey!

PR #1385, opened 2025-07-15
patch v1.1.2 to address bug in filter code and relax requirements for testing with later releases of pydantic

touma-I:

Patch filter code causing "NoneType has no object strip" using empty string for folder name instead of None
Patch requirements.txt for pit to allow testing with Prefect

When invoked directly without the user of a runtime (e.g. Filter({}).Transform(table), the call fails with the following error:
```
---> 97 if bool(self.input_arrow_folder.strip()):
     98     self.input_arrow_folder = self.input_arrow_folder if self.input_arrow_folder.endswith("[/](http://localhost:8888/)") else f"{self.input_arrow_folder}[/](http://localhost:8888/)"
     99 self.output_arrow_folder = config.get(filter_output_arrow_folder_key, None)

AttributeError: 'NoneType' object has no attribute 'strip'

```

### DIFF
--- a/transforms/language/pii_redactor/requirements.txt
+++ b/transforms/language/pii_redactor/requirements.txt
@@ -1,9 +1,10 @@
 # Later versions of numpy is ausing error with:
 # AttributeError: module 'numpy' has no attribute 'AxisError'
-numpy < 1.29.0
+# there is no such a thing as 1.29.0. Last 1.x release is 1.26.4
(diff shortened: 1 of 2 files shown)


klwuibm (approved):
LGTM.

roytman (approved):

swith005 (approved):
lgtm

PR #1388, opened 2025-07-16
Updated the Gneissweb notebook with the latest release and new API for the filter

shahrokhDaijavad:
We want the GneissWeb example as a "supported" recipe that we maintain. As such, I copied it to the recipe folder and updated it with the latest PyPi release, and tested it. During the test, I realized the changed API for the filter transform, so I changed that too. The notebook was tested successfully, but the output cells were removed. 

Once this notebook is merged, we can remove the old one from the example folder, but before doing that, we need to inform Bhatta and the GneissWeb team, since they have a link in their blog to this notebook.


swith005 (approved):
Thanks for updating and testing!

PR #1390, opened 2025-07-22
finalize numpy<=1.26.4 across all requirements.txt files

matouma:

when install multiple transforms, the dependencies across requirements.txt is NOT enforced by pip and we can run into situations where pip install numpy >=2.0 even though the core library requires <=1.26.4. To address this, all transforms that need numpy must specify <=1.26.4

Also removed uuid as it is no longer required for python 3 

https://github.com/data-prep-kit/data-prep-kit/issues/1384


swith005 (approved):
lgtm

PR #1395, opened 2025-07-25
Relax requirements for xxhash for work Haifa is doing

matouma:

data-prep-toolkit-transforms[language] 1.1.2.post1 depends on xxhash==3.4.1; extra == "language"
Other tools that can use DPK depend on xxhash>=3.5.0


roytman (approved):
LGTM

PR #1396, opened 2025-07-25
Parse metadata.json at end of the run and flag for exceptions

matouma:

Flag runs as "failed" if they have one or more exception at the end of the run

https://github.com/data-prep-kit/data-prep-kit/issues/1392


data-processing-lib/python/src/data_processing/data_access/data_access.py
```diff
@@ -191,7 +191,6 @@ def get_batches_to_process(self, batch_size=None):
         for i in range(0, len(files), bs):
             yield files[i:i + bs]
 
-    @abstractmethod
```
swith005, line 194: why is this removed as an abstract method?

swith005 (commented):
please see comment regarding data_access

data-processing-lib/python/src/data_processing/data_access/data_access.py
touma-I: @swith005 This class does not make sense for non-hierarchical data stores (such as lakehouse)

swith005 (approved):
lgtm

Issue #1391, opened 2025-07-22
[Feature] Docling2parquet needs to handle PDF files with mathematical formulas


shahrokhDaijavad:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

transforms/docling2parquet

### Feature

When testing the native docling package with the pdf file attached and using the enrichment features: https://docling-project.github.io/docling/usage/enrichments/, we found that it didn't handle the equations properly. We reported this to the docling team and they fixed it today. 
We need to bring the `--enrich-formula` to docling2parquet! 

[Newton2.pdf](https://github.com/user-attachments/files/21370171/Newton2.pdf)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

shahrokhDaijavad:
@ShiroYasha18 Is this something you can handle? We now have a real use case, in which we need this feature of docling in DPK.

ShiroYasha18:
Sure @shahrokhDaijavad , thanks for tagging me — I’ll look into this issue and see how best it can be handled.

ShiroYasha18:
Hello soo quick update :



I tested out things with ```--enrich-formula ``` in docling and tried to bridge it to dpk. The bridge is done via ```PdfPipelineOptions```. However one of the issues I have observed that these for these pdfs the formulas are not really getting decoded. So I tried to test```--enrich-formula ```  from docling too and the results were same   For example : 

[Newton2 (2).pdf](https://github.com/user-attachments/files/21475271/Newton2.2.pdf) is the original pdf

[Newton2_enhanced_formulas.docx](https://github.com/user-attachments/files/21475719/Newton2_enhanced_formulas.docx) this is after using the ```--enrich-formula ```  clearly all the formulas are left out / not recognised. 

to confirm this I tested it with some more pdfs .

[2501.12948v1.pdf](https://github.com/user-attachments/files/21475292/2501.12948v1.pdf)

[deepseek.docx](https://github.com/user-attachments/files/21475293/deepseek.docx)

In this they do get the formulas but its not organised properly .

cc : @shahrokhDaijavad

shahrokhDaijavad:
> Hello soo quick update :
> 
[...]

Issue #1347, opened 2025-06-30
[Bug] Unable to implement GPU in docling2parquet


azka2001:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Ray Runtime

### What happened + What you expected to happen

We are trying to enable GPU support during the first step — converting PDF files to Parquet using Docling. However, we have not found any documentation or backend configuration options that allow us to utilize GPUs during this process.

Our environment details:

- Running inside a Docker container with the full Data Prep Kit environment set up

- Hosted on an EC2 instance with 4 available GPUs

- The cluster detects and lists all 4 GPUs correctly

Despite this, the docling2parquet step does not seem to utilize any GPU resources, and there’s no clear way to enable or configure this support.

Could you please confirm whether GPU acceleration is supported for this step, and if so, provide guidance or configuration steps to enable it?

### Reproduction script

%%time

from dpk_docling2parquet.ray.transform import Docling2Parquet
from dpk_docling2parquet.transform import docling2parquet_contents_types

STAGE = 1
print (f"🏃🏼 STAGE-{STAGE}: Processing input='{input_dir}' --> output='{output_docling2pq_dir}'\n", flush=True)

result =  Docling2Parquet(input_folder= input_dir,
                    output_folder= output_docling2pq_dir,
                    data_files_to_use=['.pdf'],
                    docling2parquet_contents_type=docling2parquet_contents_types.MARKDOWN,   # markdown
                    docling2parquet_do_ocr=True,
                
                          
                    # runtime config
                    run_locally= True,
                    num_cpus= CONFIG_RAY_NUM_CPUS,
                    #num_gpus= CONFIG_RAY_NUM_GPUS,
                    #use_gpu=True,
                    memory= CONFIG_RAY_MEMORY,
                    runtime_num_workers = CONFIG_RAY_RUNTIME_WORKERS,

                    ).transform()

if result == 0:
    print (f"✅ Stage:{STAGE} completed successfully")
else:
    raise Exception (f"❌ Stage:{STAGE}  failed")

### Anything else

It is able to pick it up in the cluster resources: 
(orchestrate pid=35155) 10:48:27 INFO - Cluster resources: {'cpus': 48, 'gpus': 4, 'memory': 120.64055500086397, 'object_store': 55.69447746220976}
[...]

Issue #1145, opened 2025-03-19
[Feature] IBM's SmolDocling integration for ocr ai as an another option for ocr in pdf2parquet


ShiroYasha18:
### Search before asking

- [x] I searched the [issues](https://github.com/IBM/data-prep-lab/issues) and found no similar issues.

### Component

transforms/pdf2parquet

### Feature

Hi I was recently been working with pdf2parquet for couple of months  and I have tested the pdf2parquet on multiple documents for my internship project at IBM. I have seen that traditional ocr fails on handwritten documents and documents with other issues like multicolor and different size of fonts which is very classical for the traditional ocrs. For my project I used a vlm but recently IBM's docling team released smoldocling which sure is pretty impressive. I would like to contribute to this project by integrating that feature in the pdf2parquet pipeline! This will enable the support for pdfs and maybe images too in future there comes a image2parquet around. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

touma-I:
@ShiroYasha18 Thank you. I would like to tag @dolfim-ibm for his thoughts on this. Also, Do you happen to have specific example of sample data that shows where the gap is in today's release of the code and how the proposed enhancement will address those gaps?

dolfim-ibm:
You basically ehave to upgrade the Docling version and reproduce this PR which exposes the option in the Docling CLI: https://github.com/docling-project/docling/pull/1199/files#diff-bab084dbf6f5d7e3159fc059293894cf3cf58bf0fa70bd154382e03d9ba0184b.

In practice:
1. Upgrade Docling
2. Define the new `pipeline` option
3. Modify the pdf2parquet code setting the `pipeline_options` similar to the PR linked above.
4. This might also require updating the test results.

agoyal26:
@ShiroYasha18 Did the above steps help ?

ShiroYasha18:
Hello sorry for the late reply ! 

@touma-I  Thank you for assigning this issue to me. I will do the PR for the same. The major sample data which I have tested this for is handwritten answer sheets which are present in the educational institutions . Basically the basic ocr doesnot perform well for such data so it was expected that the the text wont be extracted . here is a sample image I am talking about [

![Image](https://github.com/user-attachments/assets/6e7033a6-1bc3-4e98-bd95-8eee44351f7b)



[...]

PR #1397, opened 2025-07-29
Formula patch bridge from docling + Rapidocr and ocrmac additions + GPU accelerator patch + VLM backend

ShiroYasha18:
Bridges the ```-- enrich_formula``` from docling for the formula patch 
I also added the 2 new ocr engines which was added to docling - ocrmac and the rapidocr as it was through the same ```pipeline_options```

Update :
Added the GPU accelerator patch as it was being asked by couple of folks in discussions and issues . Imported necessary pipeline options from Docling and wrote a test for the same . 

Update part 2 :
Added the VLM backend functionality with the necessary pipeline options . Added all the VLM models .

In reference to the issue #1391 
GPU Accelerator issue : #1347 
VLM backend : #1145


shahrokhDaijavad:
Thanks, @ShiroYasha18.

A few comments:

1) Adding GPU accelerator option is fine. Thanks.
2) There are unverified commits in this PR. If you do not want to use the squash option, you need to close this PR and open a new one with all commits being verified. 
3) The added test python file is missing license information. 
4) The reason for failing test-src is that all expected files need to be regenerated after changing Docling transform files. Please see how this is done, using `make generate-expected` as a Makefile target.

transforms/code/code2parquet/dpk_code2parquet/transform.py
touma-I, line 1: Why is this in here?

touma-I (changes requested):

transforms/code/code2parquet/dpk_code2parquet/transform.py
ShiroYasha18: it was a part of my older PR #1235 

It is a small typo error in the code2parquet where the this fix ensures that the snapshot configuration parameter is correctly utilized, allowing users to specify different values for both the domain and the snapshot. As a result, the metadata in the output Parquet files accurately reflects the intended configuration. Without this fix, the snapshot value would always default to the domain value, which could lead to confusion during later analysis of the processed data

If not required I will remove this

PR #1403, opened 2025-08-05
latest release of setuptools is breaking the build

touma-I:

The latest release of setup tools is breaking the build with the following error. 

```
LookupError: setuptools-scm was unable to detect version for /Users/touma/data-prep-kit-outer/transforms.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj

Alternatively, set the version with the environment variable SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME} as described in https://setuptools-scm.readthedocs.io/en/latest/config/

ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel

```

Need to adapt out toml files. For now, stay with earlier working release

### DIFF
--- a/data-connector-lib/pyproject.toml
+++ b/data-connector-lib/pyproject.toml
@@ -32,7 +32,7 @@ Issues = "https://github.com/data-prep-kit/data-prep-kit/issues"
 Documentation = "https://data-prep-kit.github.io/data-prep-kit/"
 
 [build-system]
-requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=68.0.0", "wheel", "setuptools_scm[toml]>=7.1.0,<=8.3.0"]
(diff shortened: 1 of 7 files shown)


swith005 (commented):
@touma-I thanks for pushing this. 

one concern is are the test-src tests failing with these fix being added?

touma-I:
> @touma-I thanks for pushing this.
> 
> one concern is are the test-src tests failing with these fix being added?

Yes. I had missed a few of the toml files. Now it passes. Please approve and merge as you see fit.

swith005 (approved):
lgtm

Issue #1413, opened 2025-08-12
[Feature] allow input/output_folder in local config to be blank


swith005:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Other

### Feature

Currently, validate config for data_access_local requires `input_folder` and `output_folder` to be set; would be better to allow for either to be blank in case only input or output is desired for the data access. In either case, if not present in the config, defaulting to current directory would suffice. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

swith005:
merged in latest release

PR #1414, opened 2025-08-12
updated valid config for io for data_access_local

swith005:
make missing input_folder and output_folder valid config for data_local_access when initiated via factory - defaulting to current dir if not set. 

ref: issue #1413


touma-I (approved):
LGTM

Issue #1416, opened 2025-08-14
[Feature] Enhance the base transform classes to provide information about the transformation.


roytman:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Other

### Feature

Base transform classes `AbstractTransform`, `AbstractBinaryTransform`, and `AbstractTableTransform` do not define any methods that can provide information about the transform processing. For example, which pyArrow tables are added by the class, which are deleted and which are required. 
This metadata can facilitate the generation and validation of automatic workflows. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

PR #1422, opened 2025-08-18
Add metadata functions to the Base Transforms. 

roytman:
We need it for automatic workflow generation and creation

Related issue #1416

### DIFF
--- a/data-processing-lib/python/src/data_processing/transform/abstract_transform.py
+++ b/data-processing-lib/python/src/data_processing/transform/abstract_transform.py
@@ -12,7 +12,7 @@
 ################################################################################
 
 from abc import ABC, abstractmethod
-from pydantic import BaseModel, Field
+from pydantic import BaseModel
 from typing import Any, ClassVar
 
 
@@ -23,6 +23,7 @@ class AbstractTransform(BaseModel, ABC):
 
     NAME: ClassVar[str] = "name"
     ID: ClassVar[str] = "id"
+    DESCRIPTION: ClassVar[str] = "description"
     JOB_ID: ClassVar[str] = "job_id"
     JOB_RUN_ID: ClassVar[str] = "job_run_id"
     CONTEXT_ID: ClassVar[str] = "context_id"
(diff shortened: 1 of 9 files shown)


data-processing-lib/python/src/data_processing/transform/abstract_transform.py
```diff
@@ -12,12 +12,79 @@
...
+        This simply stores the given instance in this instance for later use.
+        """
+        if config is None:
+            raise ValueError("config cannot be None")
+
+        self.config = config.copy()  # optional: copy to avoid mutating external dict
+        self.name = config.get(TransformConstants.NAME)
+        self.id = config.get(TransformConstants.ID)
+        self.job_id = config.get(TransformConstants.JOB_ID)
+        self.job_run_id = config.get(TransformConstants.JOB_RUN_ID)
```
revit13, line 42: Are these fields mandatory?

data-processing-lib/python/src/data_processing/transform/abstract_transform.py
```diff
@@ -12,12 +12,79 @@
...
+
+    def get_required_features(self):
+        # The concrete subclasses will retrieve the required features.
+        return []
+
+    def get_metadata(self) -> dict:
+        """
+        Return the transform matadata
+        """
+        pass
```
revit13: should it return {} or it has @abstractmethod decorator?

data-processing-lib/python/src/data_processing/transform/abstract_transform.py
```diff
@@ -12,12 +12,79 @@
...
+
+        self.config = config.copy()  # optional: copy to avoid mutating external dict
+        self.name = config.get(TransformConstants.NAME)
+        self.id = config.get(TransformConstants.ID)
+        self.job_id = config.get(TransformConstants.JOB_ID)
+        self.job_run_id = config.get(TransformConstants.JOB_RUN_ID)
+        self.context_id = config.get(TransformConstants.CONTEXT_ID, self.job_id)
+        self.output_features_to_drop = config.get(TransformConstants.OUTPUT_FEATURES_TO_DROP, [])
+
+    def validate(self, **kwargs) -> None:
```
revit13, line 46: should it have @abstractmethod decorator?

roytman: Ideally, it should be, but it will require updating all transforms, so for now, I have not set it intentionally.

data-processing-lib/python/src/data_processing/transform/abstract_transform.py
roytman: Correct, I added the return statement. And I have not set the decorator intentionally, to prevent updating all transforms (for now)

roytman:
We discussed yesterday the possibility of separating entries defined by `TransformCategory` or `ExecutionStatus`. Unfortunately, Python doesn't allow inheritance of Enumerations. 
```python
class TransformCategory(str, Enum):
    Extract = "Extract"

class TransformCategoryEx(TransformCategory):    # <-- This doesn't work
    Ingest = "Ingest"
    Functional = "Functional"
```

Another option is to use plain classes with String values (we can prevent accidental modification of the values).
However, in this case, the statement `category: TransformCategory = TransformCategory.Extract` will work but will produce a type checking warning `Expected type 'TransformCategory', got 'str' instead`

PR #1428, opened 2025-08-19
Define input/output path for in-memory DataAccess.

Mohammad-nassar10:

This PR aligns DataAccessMemory with other DataAccess types by implementing the missing get_output_folder() and get_input_folder() methods.

By implementing these methods, we enable a standard way for transforms to access the results of preceding transforms. For example, in a Transform A -> Transform B pipeline, Transform B can now use DataAccessMemory's get_output_folder() method to retrieve the in-memory path or key to the saved output of Transform A. This ensures consistency and compatibility across all DataAccess classes, including DataAccessLocal.

### DIFF
--- a/data-processing-lib/python/src/data_processing/data_access/data_access_memory.py
+++ b/data-processing-lib/python/src/data_processing/data_access/data_access_memory.py
@@ -59,8 +59,13 @@ class DataAccessMemory(DataAccess):
             self.input_folder = None
             self.output_folder = None
         else:
-            self.input_folder = os.path.abspath(config.get("input_folder", None))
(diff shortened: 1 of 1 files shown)


swith005 (commented):
Thanks for submitting this PR. However, I don't think there's a logical connection of "input/output" path for in-memory; the abstractions were created before for this data access was included, and will perhaps be reevaluated to eliminate enforcement;

Mohammad-nassar10:
Thanks @swith005, I'll close the PR as I found an alternative solution to use the existing implementation of the DataAccessMemory class.
But anyways maybe it is worth it to have a standard "api"/functions for all the types of DataAccess classes in order to work with them transparently without handling each DataAccess type in a different way.

PR #1431, opened 2025-08-26
fix additional secrets

roytman:
Fixes the issue when we provide several additional secrets to KFP


kfp/doc/simple_transform_pipeline.md
```diff
@@ -371,3 +371,26 @@ type: Opaque
...
+Another way to export secrets is by using the `other_secrets` KFP pipeline parameter, as illustrated in the code_quality workflow example below.
+With this approach, secrets values can be modified at runtime.
+Note that the secrets in `other_secrets` are applied to both the head and worker pods.
+
+```bash
+def code_quality(
+    # Ray cluster
+    ray_name: str = "code_quality-kfp-ray",  # name of Ray cluster
+    ...
+    other_secrets: dict = {HF_SECRET: {HF_READ_ACCESS_TOKEN: HF_SECRET_KEY}},
```
roytman: can you provide an example with several secrets?

do you provide the `secret key` here? If yes, it could be the reason why we added the `other_secrets` option. Should we prevent this option?

kfp/doc/simple_transform_pipeline.md
```diff
@@ -371,3 +371,26 @@ type: Opaque
...
+    ...
+```
+
+If an environment variable is defined in both `other_secrets` and `ray_head_options` or `ray_worker_options`, the latter will take precedence.
```
roytman: when you say `an environment variable is defined in both` do you mean `from_ref` or `HF_SECRET` ?

transforms/language/lang_id/kfp_ray/lang_id_wf.py
```diff
@@ -138,6 +138,7 @@ def lang_id(
     # data access
     data_s3_config: str = "{'input_folder': 'test/lang_id/input/', 'output_folder': 'test/lang_id/output/'}",
     data_s3_access_secret: str = S3_SECRET,
+    other_secrets: dict = {},
```
roytman: why do we need this line?

kfp/kfp_ray_components/src/create_ray_cluster.py
```diff
@@ -10,13 +10,15 @@
...
 ################################################################################
+import copy
 import json
 import os
 import sys
 
 from runtime_utils import KFPUtils, RayRemoteJobs
 
 
+values_from_key = "valuesFrom"
```
roytman: It looks like a const, should it be capitalized?

kfp/kfp_ray_components/src/create_ray_cluster.py
```diff
@@ -34,6 +36,18 @@ def start_ray_cluster(
...
+        """
+        Combine environment variables specified in the other_secrets_dict and ray_pod_options_dict dictionaries,
+        with values in ray_pod_options_dict taking precedence.
+        :param other_secrets_dict: other_secret dict
+        :param ray_pod_options_dict: head_options or worker_options dict
+        :return: the merged dict
+        """
+        values_from_other_secrets = other_secrets_dict.get(values_from_key, {})
+        ray_pod_options = ray_pod_options_dict.get(values_from_key, {})
+        return copy.deepcopy(values_from_other_secrets | ray_pod_options)
```
roytman, line 49: why do we need `deepcopy`?

kfp/doc/simple_transform_pipeline.md
revit13: I added explanation. hope its clearer. Thanks

transforms/language/lang_id/kfp_ray/lang_id_wf.py
revit13: Its a mistake Thanks

kfp/kfp_ray_components/src/create_ray_cluster.py
revit13: Doen. Thanks

swith005 (approved):
tested and works - thanks!

Issue #1432, opened 2025-08-26
[Bug] Tokenization fails if the content is a list of strings


touma-I:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

transforms/tokenization

### What happened + What you expected to happen

Tokenization transform only works on string types. when tokenizing a payload that consists of a list of string, the process fails with the following:
```
(RayTransformFileProcessor pid=398, ip=172.17.59.38)   File "/home/ray/anaconda3/lib/python3.12/site-packages/data_processing/runtime/transform_file_processor.py", line 92, in process_file
(RayTransformFileProcessor pid=398, ip=172.17.59.38)     out_files, stats = self.transform.transform_binary(file_name=f_name, byte_array=filedata)
(RayTransformFileProcessor pid=398, ip=172.17.59.38)                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(RayTransformFileProcessor pid=398, ip=172.17.59.38)   File "/home/ray/dpk_tokenization2arrow/transform.py", line 131, in transform_binary
(RayTransformFileProcessor pid=398, ip=172.17.59.38)     return self._transform_to_arrow(table=out_tables[0], file_name=file_name, tokenization_metadata=metadata)
(RayTransformFileProcessor pid=398, ip=172.17.59.38)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(RayTransformFileProcessor pid=398, ip=172.17.59.38)   File "/home/ray/dpk_tokenization2arrow/transform.py", line 96, in _transform_to_arrow
(RayTransformFileProcessor pid=398, ip=172.17.59.38)     writer.write(pa.record_batch([pa.array(token_line)], schema=schema))
(RayTransformFileProcessor pid=398, ip=172.17.59.38)                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(RayTransformFileProcessor pid=398, ip=172.17.59.38)   File "pyarrow/table.pxi", line 5828, in pyarrow.lib.record_batch
[...]

PR #1433, opened 2025-08-26
[bug] Extend tokenization to process list of strings

matouma:

when presented with a payload that consists of a list of strings, all the strings are concatenated using double new lines and then tokenized as a single string.
'\n\n'.join(list)

#1432


swith005 (approved):
changes make sense and lgtm

transforms/universal/tokenization2arrow/tokenization2arrow-rayjob.yaml
```diff
@@ -0,0 +1,118 @@
...
+                    cpu: 1
+                    memory: 16Gi
+                  limits:
+                    cpu: 1
+                    memory: 16Gi
+  # Job details
+  # The command to execute the Python script
+  entrypoint: >
+    python -m dpk_tokenization2arrow.ray.runtime 
+    --tkn_tokenizer s3://cos-optimal-llm-pile/bluepile-processing/models/granite-4.0-tiktoken-r250219a/
```
swith005: maybe not good idea to show cos-path in yaml file.

transforms/universal/tokenization2arrow/tokenization2arrow-rayjob.yaml
swith005, line 1: I feel this yaml file would best belong in the ray folder rather than the root

transforms/universal/tokenization/dpk_tokenization/transform.py
swith005, line 1: why is this substitute something user can define? Why not have blank?

swith005 (changes requested):
please review comments

transforms/universal/tokenization2arrow/tokenization2arrow-rayjob.yaml
touma-I: @swith005 We should be able to use with the python image as well.

transforms/universal/tokenization2arrow/tokenization2arrow-rayjob.yaml
touma-I: Good catch. Fixed using  environment variable to hide actual value.

transforms/universal/tokenization/dpk_tokenization/transform.py
touma-I: I just don't know what @Param-S would want so I made it configurable so we can merge the code and not wait for final decision

transforms/universal/tokenization2arrow/tokenization2arrow-rayjob.yaml
swith005: @touma-I but doesn't the entry point specifically state 
`python -m dpk_tokenization2arrow.ray.runtime` for entrypoint, a well as raycluster specs and the name of file? I like the idea of including the yaml for running on kind, but without context, wouldn't know what this was just placed in the root.

swith005 (approved):
agree with changes per follow-up conversation.

PR #1434, opened 2025-08-26
updated filter transform to return empty table with original schema r…

swith005:
…ather than empty list


transforms/universal/filter/dpk_filter/transform.py
```diff
@@ -262,6 +260,38 @@ def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Tab
         
             return [filtered_table_cols_dropped], metadata
 
+    def transform_binary(self, file_name: str, byte_array: bytes) -> tuple[list[tuple[bytes, str]], dict[str, Any]]:
```
touma-I: @swith005: Why did you choose to overwride the method rather than modify the one in the base class?

touma-I (changes requested):
Can we do it differently than overwriting the transform_binary method? I think this will make it harder to maintain (2 methods that do "almost" the same thing). down the road we want to make sure any transform can return an empty table (not only an empty list)

transforms/universal/filter/dpk_filter/transform.py
swith005: If we overwrite the base class, we'd have to tests all other transforms. Hence, since this PR is only concerned with filter, better to only update filter's `transform_binary`. However, according to datasift team, they don't even use the transform_binary for filter, so perhaps we can even just remove the override altogether? What's your thoughts on this @touma-I ?

transforms/universal/filter/dpk_filter/transform.py
```diff
@@ -197,7 +195,7 @@ def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Tab
         if file_name is not None:
             if not file_name.endswith(".parquet"):
                 self.logger.error(f"Error: input_file name doesn't end with '.parquet': {file_name}")
-                return [], {"wrong file tpye": 1}
+                return [table.schema.empty_table()], {"wrong file type": 1}
```
touma-I: Check if we need to continue to rreturn [] if the input is not valid...

touma-I (approved):
Subject to final check on file_name

PR #1435, opened 2025-08-27
Soham Internship Pipeline

SohamDasgupta91:
End to end pipeline that prompts a granite model to get article titles, passes them through the Wikimedia API and runs RAG evaluation.


swith005 (changes requested):
@SohamDasgupta91 thanks for submitting, but there's a lot of information that should be removed from checking in - such as username/passwords/tokens, etc. Please have these removed. 
Would be better to include a markdown cell detailing where/how to retrieve this info for replicating the example

swith005 (changes requested):
As discussed, please update PR with the following changes: 

- Add requirements.txt file for libraries necessary to work 
- Add markdown for prerequisites before running (i.e. getting wikimiedia account + token, hugging face token, setting up venv and installing specific version of dpk + requirements, etc. )
- Add markdown for introduction and sections explaining what this notebook does, and the metrics/results
- Ensure LLMs come from public repository (i.e. hugging face)

PR #1437, opened 2025-09-02
Drop lower bound on boto3 dependency.

revit13:
Drop lower bound on boto3 dependency


touma-I (changes requested):
@revit13 Please run make 'set-version' to update all requirements/pyprojects that are dependent on new suffix. thanks

swith005 (approved):
looks good, but also reporting fdedup failure; will need to merge after polars issue resolved

touma-I (approved):

PR #1444, opened 2025-09-16
Change in the governance personnel

shahrokhDaijavad:
Replacing Sophie with Adrian as an LF TSC member


swith005 (approved):
thanks for the update

PR #1456, opened 2025-09-25
added options for vlm granite docling in docling2parquet

swith005:
## 

adding support for granite-docling conversion after model released on huggingface


touma-I (approved):
LGTM

PR #1458, opened 2025-09-26
Gneissweb: use new convention for runtime file names

touma-I:

rename files to use new convention and update test notebook accordingly.


swith005 (approved):
lgtm! thanks

shahrokhDaijavad (approved):
LGTM too.

PR #1460, opened 2025-09-26
Tekton pipelines running on K8s

touma-I:


tekton/tasks/task-docling-2.yaml
```diff
@@ -0,0 +1,40 @@
...
+      script: |
+        #!/bin/sh
+        JOBFILE=/workspace/jobs/data-prep-kit/transforms/language/docling2parquet/docling2parquet-pythonjob.yaml
+        ##############################################################################
+        ## Most jobs follow the same pattern, so we can templatize this in the future
+        JOBNAME=`yq '.metadata.name' $JOBFILE`
+        TMPFILE=$(mktemp)
+        envsubst < "$JOBFILE" > "$TMPFILE"
+        cat "$TMPFILE" && kubectl apply -f "$TMPFILE"
+        echo "Watiing for Kubernetes rayjob/$JOBNAME to finish or timeout after 24h..."
```
klwuibm: To be consistent, I think rayjob/$JOBNAME should be changed to pythonjob/$JOBNAME on line 36.

klwuibm (approved):
Added a nitpick on Line 36 on tekton/tasks/task-docling-2.yaml.

tekton/tasks/task-docling-2.yaml
touma-I:
```suggestion
        echo "Watiing for Kubernetes pythonjob/$JOBNAME to finish or timeout after 24h..."
```

tekton/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Pipeline expects configuration parameters such Bucket name, Inuput folder and output folder for each transform 
+```
+spec:
+  workspaces:
+    - name: dpk-pipeline-ws
+  params:
+    - name: S3_BUCKET_PATH
+      description: Bucket name or bucket name and path for the root S3 folder 
+      type: string
+    - name: HTML_ZIP
```
swith005: curious about the name for the docling input i.e. `why HTML_ZIP` instead of `DOCLING_INPUT`?

tekton/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+tasks:
+    - name: execute-repo-task
+      taskRef:
+        name: repo
+      workspaces:
+    - name: execute-docling-task
+      taskRef:
+        name: docling
+      runAfter:
```
swith005, line 73: since tasks can runAfter is a previous task, is it possible for the input output of each subsequent task to be taken from the previous with explicitly being listed?

tekton/README.md
touma-I: @swith005 Just like any other name... no reason other than it contains all HTML files that are zipped :-)

tekton/README.md
touma-I: @swith005 I haven't figured out how to do that but maybe we can do it in another PR as we gain experience with those.

tekton/README.md
swith005: Thanks for clarifying @touma-I . so will the input always be zipped html files? if that's the case, then the name makes sense; otherwise, I would suggest renaming to `DOCLING_INPUT`

tekton/README.md
swith005: agreed; not necessary for now.

swith005 (approved):
lgtm - thanks for updating

PR #1468, opened 2025-10-01
restrict version of matplotlib in requirements.txt for flair in pii_r…

swith005:
…edactor

fixing bug installing pii_redactor with docling2parquet in venv due to matplotlib installations in flair library


touma-I (approved):
LGTM

PR #1471, opened 2025-10-01
adding dev1 release for regression testing

swith005:

### DIFF
--- a/.make.versions
+++ b/.make.versions
@@ -18,14 +18,14 @@ REQUIRED_PYTHON_VERSIONS=">=3.10,<3.13"
 # for integration testing or other
 
 DPK_NEXT_RELEASE=1.0.3
-DPK_VERSION_SUFFIX=.dev0
+DPK_VERSION_SUFFIX=.dev1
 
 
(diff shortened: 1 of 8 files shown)


.make.versions
```diff
@@ -18,14 +18,14 @@ REQUIRED_PYTHON_VERSIONS=">=3.10,<3.13"
 # for integration testing or other
 
 DPK_NEXT_RELEASE=1.0.3
-DPK_VERSION_SUFFIX=.dev0
+DPK_VERSION_SUFFIX=.dev1
 
 
 DPK_CONNECTOR_NEXT_RELEASE=0.2.4
-DPK_CONNECTOR_SUFFIX=.dev0
+DPK_CONNECTOR_SUFFIX=.dev1
```
touma-I, line 25: I don't think we are releasing the connector... Nothing had changed in this module. right ?
```suggestion
DPK_CONNECTOR_SUFFIX=.dev0
```

touma-I (approved):
Prior to release, need to restore suffix for dpk connector since this is not being released

PR #1478, opened 2025-10-03
updated rag_pdf_example with version 1.1.4 and docling2parquet

swith005:


shahrokhDaijavad (approved):
@swith005 I made two changes:
1) I went back to using the .env file for the replicate token in rag_4... notebook, because in the README file, we talk about .env file. The way this works is that we copy env.sample to .env and then change the .env file by replacing xyz with the actual token.
2) I updated the dpk_process_ray file  that you had not touched. I tested after the change and it works correctly. Functionally, it does exactly what the python notebook does, so you don't need to use it in a demo of the workflow from 1-4. 

Everything has been tested in this version, and it works correctly.

PR #1479, opened 2025-10-03
Restructured folder separating tasks from pipelines

touma-I:

Restructured folders
Use script template in all of the tasks


klwuibm (approved):
LGTM

PR #1490, opened 2025-10-13
Update doc_id-ray.ipynb with correct int column name

swith005:
updated parameter for int_column in ray notebook


shahrokhDaijavad (approved):
LGTM. Thank you, @roytman for catching this and thank you, @swith005 for fixing it so fast!

PR #1492, opened 2025-10-14
Enable Tekton pipeline to use python runtime

touma-I:

Provide examples and Dockerfile for using python runtime images in addition to ray runtime images


transforms/language/docling2parquet/Dockerfile.python
klwuibm, line 1: I wonder if we still need to copy the transform code to the container and installing the data-prep-kit lib. I see in my Dockerfile.python in the text_encoder transform these: 
```
# Copy and install data processing libraries 
# These are expected to be placed in the docker context before this is run (see the make image).
COPY --chown=dpk:root data-processing-dist/ data-processing-dist/
RUN  pip install data-processing-dist/${DPK_WHEEL_FILE_NAME}

# END OF STEPS destined for a data-prep-kit base image 

COPY --chown=dpk:root dpk_text_encoder/ dpk_text_encoder/
COPY --chown=dpk:root requirements.txt requirements.txt
RUN pip install ${PIP_INSTALL_EXTRA_ARGS} --no-cache-dir -r requirements.txt

```

klwuibm: I am trying to use this as an example and create a similar `lance-commit-pythonjob.yaml`.  I understand the command and args.  But, I have questions on the `Dockerfile.python`.

PR #1497, opened 2025-10-15
updated blocklist to accept blocked_domain_list locally without daf

swith005:


transforms/universal/blocklist/dpk_blocklist/transform.py
```diff
@@ -47,6 +46,23 @@ def _get_domain_list(domain_list_url: str, data_access: DataAccess) -> set[str]:
...
     return domain_list
 
+def _get_domain_list_local(domain_list_path: str) -> set[str]:
+    from data_processing.data_access import DataAccessLocal
+
+    domain_list = set()
+    da_config = {
+        "config": {
+            "input_folder": f"{domain_list_path}",
+            "output_folder": f"{domain_list_path}",
```
touma-I, line 56: Is this needed? do we have to specify output_folder even if it is not being used?

transforms/universal/blocklist/Makefile
touma-I, line 1: This is still using old naming conventions for the files. Maybe a good time to make the transition from:
**transform_python.py** to r**untime.py** 
and from **ray/transfrom.py** to **ray/runtime.py**

touma-I (commented):
Please consider using new file names. Maybe a perfect time to make the transition. if you do so, make sure you also modify the workflow.

touma-I (approved):

PR #1498, opened 2025-10-15
DocID: Create and test new Python Dockerfile and Python job 

touma-I:

New Python job for docId
New Python Dockerfile that works on OCP
Enforce runtime.py filename 
New Dockerfile.python template file


transforms/universal/doc_id/dpk_doc_id/ray/transform.py
```diff
@@ -10,133 +10,12 @@
...
-from dpk_doc_id.transform import (
-    DocIDTransformBase,
-    DocIDTransformConfigurationBase,
-    IDGenerator,
-    id_generator_key,
-    start_id_key,
-)
-from ray.actor import ActorHandle
+## Depricated code
+## Maintained for backwards compatibility with existing workflows
```
swith005, line 17:
```suggestion
import warnings

warnings.warn(f"This module is deprecated and will be removed in a future version. Use .runtime in the future to avoid disruption in the future.", DeprecationWarning, stacklevel=2)
```

swith005 (commented):
Added suggestion for deprecation in ray runtime

swith005 (approved):
LGTM!

PR #1501, opened 2025-10-16
DocId: Tested to show depricated warning

touma-I:

had to use logger.warning for the warning message to show in my tests


swith005 (approved):
same desired effect - thanks

PR #1502, opened 2025-10-16
added easyocr to requirements sincen no longer included in latest ver…

swith005:
…sion of docling


touma-I (approved):
LGTM

roytman:
@swith005 , @touma-I , should we harden the required docling versions? It appears they release a new version every several days.

PR #1513, opened 2025-11-05
Fix Doc id to be able to process lists content

touma-I:

Current digest calculation only works for string. SFT content has alwso multi-turn messages presented as lists

Fix diget calculation for lists.

### DIFF
--- a/transforms/universal/doc_id/dpk_doc_id/transform.py
+++ b/transforms/universal/doc_id/dpk_doc_id/transform.py
@@ -94,14 +94,25 @@ class DocIDTransformBase(AbstractTableTransform):
         This implementation makes no modifications so effectively implements a copy of the
         input parquet to the output folder, without modification.
         """
-        TransformUtils.validate_columns(table=table, required=[self.doc_column])
-
         if self.hash_column is not None:
+            import hashlib, json
+            TransformUtils.validate_columns(table=table, required=[self.doc_column])
             # add doc id column
             docs = table[self.doc_column]
             doc_ids = [""] * table.num_rows
             for n in range(table.num_rows):
-                doc_ids[n] = TransformUtils.str_to_hash(docs[n].as_py())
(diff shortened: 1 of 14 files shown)


transforms/universal/doc_id/dpk_doc_id/transform.py
```diff
@@ -94,14 +94,25 @@ def transform(self, table: pa.Table, file_name: str = None) -> tuple[list[pa.Tab
...
-                doc_ids[n] = TransformUtils.str_to_hash(docs[n].as_py())
+                try:
+                      doc_ids[n] = TransformUtils.str_to_hash(docs[n].as_py())
+                except AttributeError as e:                    
+                    ### Raised exception if a list type is encountered
+                    doc_ids[n] = \
+                        hashlib.sha256(json.dumps(docs[n].as_py(), sort_keys=True, separators=(",", ":")).encode("utf-8")).hexdigest()[:16]
+
+            prev_col_name=f"{self.hash_column}.original" 
+            if prev_col_name not in table.column_names:
```
swith005, line 112: can you explain these new columns? this seems to occur regardless if list type is encountered

touma-I: Yes. that is correct. The idea here is that, regardless of the use case, if the user specifies a name for the column (e.g. uuid) that already exists in the parquet file, the existing column will be renamed with .original (e.g. uuid.original) before adding a new column with the same name uuid.

shahrokhDaijavad:
@touma-I I have now tested the `doc_id_jsonl ` notebook by specifying the `--data_files_to_use` once as `.parquet` and once as` .jsonl` (numinamath case). Is setting this parameter mandatory or optional, with the default being one of the 2 formats? If the latter, which one is the default?

swith005:
rebased as issue #1520

PR #1519, opened 2025-11-10
add a Rich-based Log handler

### LINKED ISSUE #1518: [Feature] The latest log handler provides output in JSON format, which complicates exceptions debugging
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

Other

### Feature

The latest log handler provides output in JSON format, which complicates exception debugging. We need an option to choose between JSON and developer-friendly log handlers.

- [x] Yes I am willing to submit a PR!

roytman:
The latest log handler provides output in JSON format, which complicates exception debugging. We need an option to choose between JSON and developer-friendly log handlers.

/closes https://github.com/data-prep-kit/data-prep-kit/issues/1518.

This PR adds a possibility to choose between 2 log handlers: JSON and a Rich-based one.
The output of the Rich log handler is very similar to the Prefect log handler.

image

In order to work with JSON handler, the `DPK_LOG_HANDLER` env variable should be set to "JSON" (case insensitive), in all other cases the Rich handler will be used.

The file handler always uses the JSON formatting.


README.md
```diff
@@ -118,6 +118,17 @@ The matrix below shows the the combination of modules and supported runtimes. Al
...
 </details>
 
+## Logging configuration
+DPK uses a unified logger - `dpk`. It can be configured, by setting the following environment variables
+
+| Variable name | Default value | Description                                                                                                         |
+|--|-----------|---------------------------------------------------------------------------------------------------------------------|
+|DPK_LOG_LEVEL | INFO      | The loggger level                                                                                                   |
+|DPK_LOG_FILE | None      | The path to the log file, if set the log message will be stored in the file                                         |
+| DPK_LOG_HANDLER | ""        | If set to `JSON` (case insensitive) the logs will be in JSON format                                                 |
```
swith005: since JSON is the only option taken into account, why not just have this as a flag?

swith005 (changes requested):
@roytman thanks. please review comment regarding the flag option. Also, would be good to provide test cases to confirm JSON handler requests works vs the default rich option, as well as logging to file.

README.md
roytman: thanks, done

swith005 (approved):
thanks for addressing requests - lgtm

data-processing-lib/python/src/data_processing/utils/log.py
```diff
@@ -12,39 +12,137 @@
...
+
+theme = Theme({
+    "debug": "white",
+    "info": "cyan dim",
+    "warning": "yellow",
+    "error": "red",
+    "critical": "red",
+    "time": "white",
+    "logger": "dim",
+    "message": "white",
```
touma-I, line 42: @roytman I noticed the white was difficult to read on white background. I tried to change it to blue and or yellow but I don't seem to be able to. Do you have a pointer that you can provide on how we can control the color theme? Thanks

touma-I (approved):
LGTM. Would be great if we can provide pointers on if and how to change the color scheme.

PR #1533, opened 2025-11-19
Some pip tag corrections and a README fix

shahrokhDaijavad:
1) The 3 image transforms should use `>=` instead of `==` for the packages they pip install. 
2) The pii-redactor needed a change in the version of cryptography, as I was getting the error below:
#pyopenssl 25.3.0 requires cryptography<47,>=45.0.7, but you have cryptography 44.0.3 which is incompatible
cryptography<47,>=45.0.7
3) README for the Collapse transform had a wrong title.


shahrokhDaijavad:
@swith005 @touma-I Do you know why test-image for pii_redactor is failing? I have modified its requirements.txt file so that the two notebooks succeed with the latest release (1.1.6), which is pip installed in the notebooks.

swith005:
> @swith005 @touma-I Do you know why test-image for pii_redactor is failing? I have modified its requirements.txt file so that the two notebooks succeed with the latest release (1.1.6), which is pip installed in the notebooks.

It's odd that test-src passed while test-img failed... Looking at the stack trace, it seems to be some circular import issue: 
`ImportError: cannot import name 'dispatch_model' from partially initialized module 'accelerate.big_modeling' (most likely due to a circular import)`

it comes from transformers library, but I'd think both test-src and test-img would fail for this same reason. need to look further into

PR #1534, opened 2025-11-19
Doc id fix

touma-I:

Fix issue with DocIDTransform being in the wrong file
Added target for verifying notebooks operations after changes to the transforms


.secrets.baseline
swith005, line 1: what are these .secrets.baseline changes related to, and how to confirm validation?

transforms/.make.cicd.targets
swith005, line 1: should this be part of this PR for doc id fix? this seems applicable for all other transforms, no?

touma-I: Yes, It is applicable to all transforms but since this is the first transform that uses it and having it in this PR should allow me to use it for all the subsequent PRs that I am doing for all the other transforms. @swith005

.secrets.baseline
touma-I: There are three different scenarios. @swith005 
1- The first one, is actually NOT a secret for the opensearch CI/CD but for some reason the tool things it is.. 
2- The second one is a secret but is actually a temporary secret used by CI/CD to access minio
3- The third one is a template for the opensearch

swith005: ok, thanks for clarifying

transforms/.make.cicd.targets
swith005: alright, I'm ok with introducing here for doc_id using it

swith005 (approved):
changes makes sense; thanks

PR #1553, opened 2025-12-04
Pii image notebook 

shahrokhDaijavad:
This notebook recipe was developed as we received a request to combine our PII transform capability with face-blurring capabilities of our multimedia transforms, for maximum privacy concerns.


swith005 (approved):
thanks for adding and testing - looks good!

PR #1555, opened 2025-12-11
Small fixes in the logger formatter

roytman:
We get log level env variable, and use it in several places. However, we transform the input string to upper case only once, so in other places, input like "Debug" breaks the run. 
This PR fixes it. 
Additionally, there was a bug in the definition of a place where a log message was generated. 

And finally, Rich incorrectly detects terminal width when it runs in PyCharm; this PR fixes it. 

The logger outputs are:
When we are working with `INFO` or higher level:
image

when we are working with `DEBUG` level:
image

### DIFF
--- a/data-processing-lib/python/src/data_processing/utils/log.py
+++ b/data-processing-lib/python/src/data_processing/utils/log.py
@@ -13,6 +13,7 @@
 
 import logging
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


swith005 (approved):
this looks good and makes sense to me. thanks for addressing

PR #1559, opened 2025-12-16
fix type mispatmch str/int in ededup when using int_doc_id

touma-I:

Address issue reported when running ededup on int_doc_id. 

When compiling the full list of document, the code is using the native data type (int) for the document id but when deciding if the document is duplicate or not, the code was using str(id)


swith005 (approved):
thanks for going over this with me - looks good!

PR #1560, opened 2025-12-16
Show some selected output cells

shahrokhDaijavad:
I performed a minor cleanup of the notebook and am now displaying some of the interesting output cells.


recipes/PII-image/PII_Image_redactor.ipynb
swith005, line 1: Hi @shahrokhDaijavad with granite_docling2parquet notebook, you had submitted a PR to hide the image output results as it made things unnecessarily large; wouldn't this be the same result?

shahrokhDaijavad: Thank you, @swith005. You are very observant! Two reasons:

1) In the granite_docling2parquet case, the large notebook size was due to printing a huge parquet output file that had no useful effect. In this case, showing the "blurred" face is useful and visually proves the effectiveness of our image transform. 
2) The size of the notebook in that case, with the output cells shown, was more than 4MB, which is not insignificant. Here, the size of the notebook with the image displayed is less than 150kB.

swith005: Thanks for clarifying. OK, makes sense to me :)

swith005 (approved):
change makes sense - thanks

Issue #1563, opened 2025-12-22
[Bug] dpk_text_encoder fails when GPU available


revit13:
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

transforms/text_encoder

### What happened + What you expected to happen

Running `python3.11 local_python.py` (after commenting out `embeddings_in_parquet_cli_param` in the file)

10:06:59 [INFO] transform_orchestrator.py:90 - Number of files is 1, source profile {'max_file_size': 0.04261493682861328, 'min_file_size': 0.04261493682861328, 'total_file_size':
0.04261493682861328}
10:07:01 [INFO] transform.py:121 - GPU is available!
10:07:01 [ERROR] /home/eres/data-prep-lab/data-processing-lib/python/src/data_processing/runtime/pure_python/transform_file_processor.py:57 - Exception creating transform
'TextEncoderTransform' object has no attribute 'model'
Traceback (most recent call last):
  File "/home/eres/data-prep-lab/data-processing-lib/python/src/data_processing/runtime/pure_python/transform_file_processor.py", line 55, in __init__
    self.transform = transform_class(self.transform_params)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eres/data-prep-lab/transforms/language/text_encoder/dpk_text_encoder/transform.py", line 122, in __init__
    self.model.half()
    ^^^^^^^^^^
AttributeError: 'TextEncoderTransform' object has no attribute 'model'

During handling of the above exception, another exception occurred:

### Reproduction script

python3.11 local_python.py

### Anything else

it happens when GPU is available

### OS

Ubuntu

### Python

3.11

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

klwuibm:
Fixed it.

klwuibm:
@swith005 Please close this issue, if you can.  Thanks.

PR #1564, opened 2025-12-29
Fix a bug in transform.py when GPU is available

klwuibm:
In the `init()` of `transform()` method, `self.model.half()` was referenced before` self.model `is assigned, triggering a bug. In this pull request, it is fixed.  Also removed a nonexistent parameter in `local_python.py`. 

#1563


klwuibm:
@touma-I Please review.

touma-I (approved):

swith005 (approved):
makes sense - thanks

PR #1565, opened 2026-01-05
add an option to set dpw log handlers

roytman:
Generally, DPK is a library and should avoid setting log handlers or formatters, leaving that responsibility to the application. 

However, because the repository includes numerous Python examples and Jupyter notebooks, this PR seeks a balance: it avoids the need to update every existing script while still allowing users to configure their own custom log handlers


swith005:
hi @roytman . Same verified signature problem that was reported from Revital's PR. Please close this PR and reopen with verified signatures when merging.

roytman:
I created another PR #1568 instead of this one

PR #1568, opened 2026-01-07
add an option to set dpw log handlers

roytman:
Generally, DPK is a library and should avoid setting log handlers or formatters, leaving that responsibility to the application.

However, because the repository includes numerous Python examples and Jupyter notebooks, this PR seeks a balance: it avoids the need to update every existing script while still allowing users to configure their own custom log handlers

### DIFF
--- a/data-processing-lib/python/src/data_processing/utils/log.py
+++ b/data-processing-lib/python/src/data_processing/utils/log.py
@@ -212,32 +212,3 @@ def get_dpk_logger(
         logger.addHandler(create_file_handler(log_file, log_level))
 
     return logger
-
-
-# Test logging
-# logger = get_dpk_logger()
-# logger.info("Hello, JSON world!", extra={"transaction_ID": "TRANSACTION999", "user_id": "USER999"})
-#
-# logger2 = get_dpk_logger()
-# logger2.debug("debug message")
-# logger2.info("info message")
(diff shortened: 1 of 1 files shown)


data-processing-lib/python/src/data_processing/utils/log.py
```diff
@@ -167,4 +228,16 @@ def add_handler_once(handler, tag_name):
...
+# logger3.warning("3 warning message")
+# custom_logger = get_dpk_logger(
+#     handlers=[
+#         logging.StreamHandler(),
+#     ],
+#     replace_handlers=True,
+# )
+# custom_logger.warning("4 warning message")
+# logger4 = get_dpk_logger()
+# logger4.warning("custom warning message")
```
swith005: What's with these new commented code? better to just delete

swith005 (changes requested):
request to remove commented code

data-processing-lib/python/src/data_processing/utils/log.py
roytman: done, thanks.

swith005 (approved):
thanks - lgtm

swith005:
merging while understanding that the` test-src` scripts for transforms/chain, transforms/code/malware, transforms/universal/opensearch are failing for known reasons.

PR #1570, opened 2026-01-08
update boto3 version in data-proccessing-lib requirements

swith005:


data-processing-lib/python/requirements.txt
```diff
@@ -6,7 +6,7 @@ numpy < 2.0.0
 pyarrow<=17.0.0
 # Later version 1.38.7 is breaking authentication. 
 # check compatibility with other applications that depend on botocore(DMF, DSift, etc.)
-boto3<=1.38.18
+boto3==1.42.22
```
roytman: I'd suggest providing more flexibility e.g. `boto3 ~= 1.42` or `boto3>=1.41`

swith005: agreed and updated

touma-I (approved):

swith005:
merging while understanding that the test-src scripts for transforms/chain, transforms/code/malware, transforms/universal/opensearch are failing for known reasons.

PR #1571, opened 2026-01-09
added line in template to skip certain tests-src from transforms

swith005:
adding line to .github workflow templates that lists specific transforms to skip when running `test-src`


.github/workflows/test-chain-dpk_transform_chain.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
 
     test-src:
-
+        if: false
         needs: [check_if_allowed]
```
swith005, line 64: see if needs can be run before if statement

PR #1587, opened 2026-05-11
added new malware transform with YARA rules

swith005:
Implemented a new malware scanning method based off YARA rules, separate from DPK's existing ClamAV method.


shahrokhDaijavad (approved):
LGTM, @swith005.

PR #1589, opened 2026-06-16
updated ray version for all python versions and removed pyarrow versi…

swith005:
…on constraint

systems like granite.build require a higher ray number, so testing upgraded ray is important; also pyarrow constraint no longer necessary after fastext-predict.


shahrokhDaijavad (approved):
LGTM. Nice job, @swith005!

PR #1594, opened 2026-07-13
Add context to S3 secret regarding local testing

revit13:

This PR adds inline comments to the secret file and updates the README to explicitly clarify that the MinIO S3 credentials are strictly for local testing and must not be used in production environments.


swith005 (approved):
lgtm - thanks for quick turnaround

PR #1595, opened 2026-07-14
fix(gneissweb): accumulate stats across all configured classifiers

### LINKED ISSUE #1453: [Bug] Metadata created by the Gneissweb transform shows only the stats for the last classifier when multiple classifiers are used
### Search before asking

- [x] I searched the [issues](https://github.com/data-prep-kit/data-prep-kit/issues) and found no similar issues.

### Component

transforms/gneissweb_classification

I ran the Gneissweb classifier (docker python), by specifying 4 topic classifiers as input parameters and the metadata.json file has the statistics associated with only the last classifier (science topic in this case). The metadata file is attached. 

[metadata-4topics-docling.json](https://github.com/user-attachments/files/22503421/metadata-4topics-docling.json)

### Reproduction script

See above. 

### Anything else

_No response_

### OS

MacOS

### Python

3.12

Ys876:
## What

Fixes #1453. When `ClassificationTransform` is configured with more
than one classifier, the output metadata only contained stats for the
last configured classifier, earlier classifiers' counts were silently
dropped.

## Root cause

`transform()` runs one loop iteration per configured classifier, and
each iteration does `table, stats = get_label_ds_pa(...)`. `stats` is
reassigned every iteration rather than accumulated, so after the loop
only the last iteration's dict survives to the `return` statement.
`table` doesn't have this problem, since each iteration correctly adds
a column to the same table object, it's specifically the stats
tracking that's broken.

This wasn't caught by the transform's existing test because that test
only configures a single classifier, and with one loop iteration,
"overwrite" and "accumulate" are indistinguishable.

## Fix

Accumulate into a running `all_stats` dict via `.update()` instead of
reassigning `stats` directly. Minimal logical diff: one new variable, one
`.update()` call, one return value changed.

## Testing

- Added `test/test_transform_stats.py`, a lightweight unit test that
  mocks the classifier call (`get_label_ds_pa`) so it runs in
  milliseconds with no network access or model download required.
- Confirmed the new multi-classifier test fails against the
  unmodified code (reproducing the exact reported symptom: only the
  last classifier's stats present) and passes after the fix.
- Added a second test confirming the single-classifier case (what the
  existing integration test already covers) is unaffected by this
  change.

## Scope note

Left two pre-existing, unrelated unused imports in this file
(`FastTextModel`, `get_label_ds_pa_parallel`) untouched, since they
predate this change and aren't part of the reported bug.

This repo's required `pre-commit` hooks (`black`/`isort`) reformatted
the whole `transform.py` file on commit, since the file was not
previously formatted to this repo's style. That reformat is purely
cosmetic (whitespace, quote style, line wrapping, import order) with no
logic changes beyond the 4-line fix described above; I did not run
`black` manually or expand scope myself, this is `pre-commit`'s
documented, required behavior for a file this PR touches. Happy to
split it into a separate formatting-only commit/PR if maintainers would
prefer that instead.

### DIFF
--- a/transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
+++ b/transforms/language/gneissweb_classification/dpk_gneissweb_classification/transform.py
@@ -10,20 +10,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ################################################################################
+import ast
 import os
 from argparse import ArgumentParser, Namespace
 from typing import Any
 
 import pyarrow as pa
-
-import ast
-
 from data_processing.transform import AbstractTableTransform, TransformConfiguration
 from data_processing.utils import CLIArgumentProvider, TransformUtils, load_model
-from dpk_gneissweb_classification.classification_models import FastTextModel, ClassificationModel
+from dpk_gneissweb_classification.classification_models import (
+    ClassificationModel,
+    FastTextModel,
+)
 from dpk_gneissweb_classification.nlp import get_label_ds_pa
 from dpk_gneissweb_classification.nlp_parallel import get_label_ds_pa_parallel
-
 
 
 short_name = "gcls"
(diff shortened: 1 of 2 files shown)


shahrokhDaijavad:
Thanks for your contribution, @Ys876! I had known about this problem, but never got around to fixing it. I think your fix is spot-on! @swith005 I am ready to approve this. What do you think?

swith005 (approved):
lgtm and all tests passed - thanks!

PR #1603, opened 2026-08-26
fix(pii_redactor): handle empty/whitespace-only text in _redact_pii

### LINKED ISSUE #1599: [Bug] pii_redactor crashes the whole file when a row is empty or whitespace-only

`_redact_pii` (`transforms/language/pii_redactor/dpk_pii_redactor/transform.py:78-84`) has no `else` branch:

```python
def _redact_pii(self, text: str):
    text = text.strip()
    if text:
        analyze_results, entity_types = self._analyze_pii(text)
        anonymized_results = self.anonymizer.anonymize_text(text, analyze_results)
        return anonymized_results.text, entity_types
```

For an empty or whitespace-only row it falls through and returns `None`. `transform()` then does `zip(*table[pii_contents_column].to_pandas().apply(self._redact_pii))` over the whole column, and a single `None` in that Series makes `zip(*...)` raise `TypeError: 'NoneType' object is not iterable`, so one blank row crashes the transform for the entire file, not just that row. Blank or empty documents are an ordinary feature of scraped/ingested corpora (truncated downloads, boilerplate-only pages, upstream encoding failures), so this is easy to hit in real data.

Confirmed at current `dev` (`c2bf5dc6203410fe4b942c1d2e917b89de5bc7e0`), file unchanged.

Expected: rows with empty/blank content pass through the transform (unredacted, zero detected entities) like any other transform in this repo handles empty text.

Actual: `TypeError: 'NoneType' object is not iterable` from `zip(*...)`, and the whole file's transform aborts.

### Reproduction script

```python
import pandas as pd

# same control flow as _redact_pii, transform.py:78-84
def _redact_pii(text):
    text = text.strip()
    if text:
        return ("REDACTED:" + text, ["PERSON"])

series = pd.Series(["Alice called Bob", "", "Second doc with text"])
redacted_texts, entity_types_list = zip(*series.apply(_redact_pii))
```

```
TypeError: 'NoneType' object is not iterable
```

### Anything else

Suggested fix (not built, per this repo's own norm of getting an issue assigned before sending a PR): give `_redact_pii` an `else` branch returning `(text, [])` so blank rows pass through unredacted instead of returning `None`.

Happy to send a PR for this once it's assigned, if useful.

AmirF194:
## Root cause

`_redact_pii` (`transform.py:79-84`) strips the input text and only returns a value
inside the `if text:` branch, so an empty or whitespace-only row falls off the end of
the function with an implicit `None`. `transform()` unpacks every row's result with
`zip(*table[pii_contents_column].to_pandas().apply(self._redact_pii))`, so a single
blank row raises `TypeError: 'NoneType' object is not iterable` and drops the whole
file, not just that row. Blank/whitespace-only documents are an ordinary artifact of
real scraped/ingested corpora (truncated downloads, boilerplate-only pages), so this is
a realistic crash, not a contrived edge case.

## Fix

Add an `else` branch to `_redact_pii` that returns `(text, [])`: the stripped text
passes through unredacted with zero detected entities, matching the tuple shape the
function already returns for non-empty text. `_redact_pii` has exactly one caller
(`transform()`), so this is the only path affected.

## Verification

- Added a whitespace-only row to `test/test_data.py` (shared by both
  `test_pii_redactor_transform.py` and `test_pii_redactor_redact_anonamize.py`), fails
  on `dev` HEAD (`TypeError: 'NoneType' object is not iterable`) and passes on this
  branch, run both ways in a clean container.
- `black --config .black.toml` and `isort` report both changed files unchanged.
- I ran the two transform-level test classes above with the Presidio/Flair analyzer
  and anonymizer stubbed out (this environment can't install those model
  dependencies), keyed on the exact sentences in `test_data.py` so the existing rows'
  expected output is unchanged. I did not independently run the full
  `make test-src`/`test-image` targets that exercise the real NLP stack; the new row
  never reaches the analyzer (it returns before calling it), so it doesn't depend on
  that stack either way.

Fixes #1599

### DIFF
--- a/transforms/language/pii_redactor/dpk_pii_redactor/transform.py
+++ b/transforms/language/pii_redactor/dpk_pii_redactor/transform.py
@@ -82,6 +82,9 @@ class PIIRedactorTransform(AbstractTableTransform):
             analyze_results, entity_types = self._analyze_pii(text)
             anonymized_results = self.anonymizer.anonymize_text(text, analyze_results)
             return anonymized_results.text, entity_types
+        # Empty/whitespace-only text has nothing to analyze; without this branch the
+        # implicit None return breaks zip() in transform() for the whole file.
+        return text, []
 
     def transform(self, table: pa.Table, file_name: Optional[str] = None) -> tuple[list[pa.Table], dict[str, Any]]:
         """
--- a/transforms/language/pii_redactor/test/test_data.py
+++ b/transforms/language/pii_redactor/test/test_data.py
@@ -20,9 +20,10 @@ table = pa.Table.from_pydict(
                 "My name is Tom chandler. Captain of the ship",
                 "I work at Apple and I like to eat apples",
                 "My email is tom@chadler.com and dob is 31.05.1987",
+                "   ",
             ]
         ),
-        "doc_id": pa.array(["doc1", "doc2", "doc3"]),
+        "doc_id": pa.array(["doc1", "doc2", "doc3", "doc4"]),
     }
 )
 expected_table = table.add_column(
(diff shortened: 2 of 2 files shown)


AmirF194:
The two test workflows here are still sitting on the fork-approval gate (action_required), so CI hasn't run yet. Could someone approve the run when there's a moment?

PR #1605, opened 2026-09-03
fix(ci): use pull_request instead of pull_request_target in 4 workflows

swith005:
actions/checkout now refuses to check out fork PR code from a pull_request_target workflow, since fetching and running a fork's code with the base repository's GITHUB_TOKEN, secrets, and runner access is the classic "pwn request" vulnerability. This blocks CI on any fork PR (e.g. #1596) with:

    Error: Refusing to check out fork pull request code from a
    'pull_request_target' workflow.

These 4 workflows were the only ones in .github/workflows/ still using pull_request_target. The other ~56 test workflows -- and test-transform.template, which three of these files declare themselves generated from -- already use pull_request. Switching the trigger restores that consistency and runs fork code in the correct untrusted context, with a read-only token and no secrets.

Changed files:
    .github/workflows/test-chain-dpk_transform_chain.yml
    .github/workflows/test-code-code_quality.yml
    .github/workflows/test-language-gneissweb_classification.yml
    .github/workflows/test-language-lang_id.yml

Scope notes:

- Trigger line only. The check_if_allowed job and the per-job secret plumbing ("Expose secret to trusted users working in a fork" steps and the `ref: github.event.pull_request.head.sha` checkout overrides) are intentionally left in place to keep this diff minimal and reviewable. Secrets such as HF_READ_ACCESS_TOKEN and DOCKER_REGISTRY_* are withheld by GitHub on fork PRs under pull_request regardless, so tests needing them may fail for external contributors and can be re-run by a maintainer.

- Edited by hand rather than regenerated with `make transform-tests`. That target does `rm -rf test-*-*.yml` and rebuilds from the template, which would have rewritten all ~60 workflow files and dropped local customizations in these four -- the "Test2 -" name prefix and the "staging-*" branch entry on lang_id and gneissweb_classification. Both are preserved here.

- Follow-up (not addressed here): check_if_allowed assigns `restricted_acccess` (three c's) but tests `restricted_access` (two), so the fork value never reaches the check and `if $restricted_access;` expands to a bare `if ;` -- a bash syntax error. Since test-src and test-image list it in `needs:`, fork PRs may still skip those jobs until it is fixed or removed.

### DIFF
--- a/.github/workflows/test-chain-dpk_transform_chain.yml
+++ b/.github/workflows/test-chain-dpk_transform_chain.yml
@@ -20,7 +20,7 @@ on:
             - "!**/doc/**"
             - "!**/images/**"
             - "!**.gitignore"
-    pull_request_target:
+    pull_request:
         branches:
             - "dev"
             - "releases/**"
--- a/.github/workflows/test-chain-dpk_transform_chain.yml
+++ b/.github/workflows/test-chain-dpk_transform_chain.yml
@@ -39,45 +39,11 @@ concurrency:
     group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
     cancel-in-progress: true
 jobs:
-    check_if_allowed:
-        # check if user allowed to run restricted workflows
-        runs-on: ubuntu-22.04
-        outputs:
-            restricted_access: ${{ steps.permission.outputs.restricted_access }}
-        steps:
-            - id: permission
-              run: |
-                echo "Check if ${{ github.triggering_actor }} is allowed to run priviledge workflows"
-                restricted_acccess=${{ github.event.pull_request.head.repo.fork }}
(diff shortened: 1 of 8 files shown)


touma-I (changes requested):
the workflow exposes a git hub secret. With pull_request, anyone on the internet can fork the code, write a small script that prints out the secrets, submit a PR to trigger the test script, look at the log and see the values for the secrets. The ideas with pull_request_target is that the user can submit a PR but test will not trigger until the admin review the changes makes sure the code is save before starting the test and does not expose any of the secrets in the git hub.

touma-I (approved):
Looks good. **One caveat**: The changes were not tested (workflow was not triggered). while the code looks correct, you might consider triggering the workflow to test the new changes before merging.

